Issue #305, opened 2012-08-13
Table row hash: indifferent access is inconsistent


avit:
I couldn't find where the table row hashes are initialized to see what the problem is, but:

```
row[:name]  # => "Alice"
row["name"] # => "Alice"

row.delete(:name) # => nil
row[:name]  # => "Alice"

row.delete("name") # => Alice
row["name"] # => nil
```

(`.delete` with symbols and strings works consistently in ActiveSupport's HashWithIndifferentAccess, so I presume this uses a different beast?)

PR #313, opened 2012-09-13
Properly support symbol keys in table hashes

avit:
Re: #305

At a glance, the implementation in `ast/table.rb:383` just uses Hash's default value block to fetch missing symbol keys by string. A few more pieces are needed to make it truly blind... To avoid a dependency on activesupport gem, this pull request grabs a copy of HashWithIndifferentAccess and loads it in the Cucumber namespace when the original is not defined.

### DIFF
--- a/lib/cucumber/ast/table.rb
+++ b/lib/cucumber/ast/table.rb
@@ -1,6 +1,7 @@
 require 'gherkin/rubify'
 require 'gherkin/lexer/i18n_lexer'
 require 'gherkin/formatter/escaping'
+require 'cucumber/hash_with_indifferent_access' unless defined? HashWithIndifferentAccess
 
 module Cucumber
   module Ast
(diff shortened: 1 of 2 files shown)


dmorgan3405:
To solve the issue, I have been using a monkey patch for some time, and wrapped it up in a gem ([cucumber-symbolic-hashes](https://github.com/dmorgan3405/cucumber-symbolic-hashes)) 

```ruby 
class Cucumber::Ast::Table

  def symbolic_hashes
    @header_conversion_proc = lambda {|h| symbolize_key(h)}
    @symbolic_hashes ||= build_hashes
  end

  private

  def symbolize_key(key)
    key.downcase.gsub(' ','_').to_sym
  end

end
```

I know there was a solution discussed above around a hash factory, and that may be a better solution to move forward with. 

I would love to see this ability added to cucumber, as when working in the test data arrangement world it allows me to use factory girl more effectively without having to handling the conversion elsewhere. 

If anyone else is interested in seeing a solution added for this I would love discuss it.

Issue #425, opened 2013-04-08
Scenario has no reference to feature anymore


rbu:
Before 1.2.4, we used the `scenario.feature` attribute to access the Feature from a Scenario. Ever since 3404044fd3de557d0c83e7ec0b65b0fb933b2562, Scenario only has the feature's language and tags, but no (direct) reference to the other attributes.

PR #430, opened 2013-04-08
Expose names of AST elements in Before hook

mattwynne:
See https://github.com/vcr/vcr/issues/288 and #425.

We should make this part of the public API.


mattwynne:
I think that as of 7a1ec155dc509620c9cae13a5b529017c05dc7b3 this is now covered.

Issue #73, opened 2011-06-01
Wire protocol handling of exceptions in hooks


richardlawrence:
It would be nice if exceptions in a Before or After hook over the wire protocol were handled the same way as with the Ruby language implementation. Today, I get either a long, irrelevant backtrace (from the Ruby side, not the backtrace passed over the wire) or it fails silently after the Scenario line is output. I'm not sure why I get one or the other at this point.

I've added a scenario in my fork to express the behavior I'd like: https://github.com/richardlawrence/cucumber/commit/4effaf1b34b84497b3ab162d4fe63af916a96e5b.

PR #435, opened 2013-04-14
Wire protocol: display exceptions in hooks

mattwynne:
This replaces #73.

I've investigated, and the cause is that the before/after code in the 
central ProgrammingLanugage abstraction runs before the scenario has
started. This means any exceptions at this point are handled like
exceptions in the regular Cucumber source code, rather than 
exceptions in the scenario.

I'm deferring this until 2.0

*Update: 2015-03-16*

Exceptions in Wire protocol hooks no longer crash the Cucumber process, as of #807. However the error from the hook is still not displayed, it looks like.

### DIFF
--- a/examples/hooks/features/hooks.feature
+++ b/examples/hooks/features/hooks.feature
@@ -1,0 +1,14 @@
+Feature: Hooks
+  Scenario
+
+  Scenario: First scenario
+    Given the around hook has run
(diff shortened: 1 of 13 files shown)


mattwynne:
I'd like to start again from scratch on this one. Closing.

PR #663, opened 2014-04-05
Suggested regexes are bad and result in ambiguous matches

rwz:
Imagine I'm writing a new scenario that includes something similar to this:
```cucumber
When I press "Submit" button
Then I should see "can't be blank" validation error on "Email"
```

first time I run this new scenario, cucumber will suggest me code snippets to start implementing missing steps. The snippet for validation step would be
```ruby
Then(/^I should see "(.*?)" validation error on "(.*?)"$/) do |arg1, arg2|
  pending # express the regexp above with the code you wish you had
end
```

Later I add another scenario that includes this:
```cucumber
Then I should see "foobar"
```
and cucumber, again, helpfully suggest a snippet to start implementing the step:
```ruby
Then(/^I should see "(.*?)"$/) do |arg1|
```

The problem is, the suggested regexp also matches everything, a previously defined validation step does resulting with lot of previously green steps suddenly go red with `Cucumber::Ambiguous` error.

### DIFF
--- a/lib/cucumber/rb_support/snippet.rb
+++ b/lib/cucumber/rb_support/snippet.rb
@@ -2,7 +2,7 @@ module Cucumber
   module RbSupport
     module Snippet
 
-      ARGUMENT_PATTERNS = ['"(.*?)"', '(\d+)']
+      ARGUMENT_PATTERNS = ['"([^"]*)"', '(\d+)']
 
       class BaseSnippet
 
--- a/spec/cucumber/rb_support/snippet_spec.rb
+++ b/spec/cucumber/rb_support/snippet_spec.rb
@@ -28,7 +28,7 @@ module Cucumber
           @pattern = 'A "string" with 4 spaces'
 
           snippet_text.should == unindented(%{
-          Given(/^A "(.*?)" with (\\d+) spaces$/) do |arg1, arg2|
+          Given(/^A "([^"]*)" with (\\d+) spaces$/) do |arg1, arg2|
             pending # Write code here that turns the phrase above into concrete actions
           end
           })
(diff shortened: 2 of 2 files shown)


mattwynne:
I'm going to close this PR. I do appreciate it, thanks, but I'd prefer to keep the generated snippets non-intimidating for the 80/20 use-case. I hope we can use your more robust ones when #708 is implemented.

Any objections?

aslakhellesoy:
@mattwynne are you saying `([^"]*)` is more intimidating than `(.*?)`? (I would disagree).

Cucumber suggested `([^"]*)` in snippets for many years until we accepted a poor patch that switched it to `(.*?)` - a regression IMO.

I think we should switch it back - it should be easy to do.

aslakhellesoy:
Let me elaborate on why I disagree that `([^"]*)` is more intimidating than `(.*?)` and why I think this PR should be applied.

Let's assume we have two kinds of users - those who know enough regular expressions to understand both styles, and those who don't. There might also some people who understand one style, but not the other, but that is probably a very small percentage, so let's ignore them for the sake of this argument.

The users who do understand regular expressions are unlikely to find either style more intimidating than the other.

The users who don't understand regular expressions are likely to find both styles equally intimidating.

Of course, I'm only hypothesising here, but so are you @mattwynne. We don't have data to support any of these hypotheses.

Since we don't have data to support either hypothesis we should decide what style to use based on which one is most _correct_ and causes the least surprise during _use_. I would argue that even if we _did_ have data to support the claim that people find `([^"]*)` more intimidating than `(.*?)`, we should still use `([^"]*)` because it is more correct.

@rwz has already made a good case that the `(.*?)` can cause surprising behaviour while the original `([^"]*)` will not.

rwz:
I'd say that `(.*?)` is arguably more confusing than straightforward `[^"]*`. In my experience a lot of people do not understand the concept of non-greedy `?` modifier for `*`.

I find myself explaining this regexp to others a lot more often than `[^"]`.

mattwynne:
Oh the only intimidation I'm thinking about is in the aesthetic. I'm convinced by both your points of view, and we'll merge this in for release with 2.0

mattwynne:
Thanks @rwz!

PR #665, opened 2014-04-08
[i18n] Improve the Esperanto translation

tepreece:
The existing Esperanto translations are literal translations of the original English text, and not especially idiomatic. They also contain a few errors. I have re-written them with better translations.

Please note that this also requires my improved Esperanto translations for Gherkin.


mattwynne:
Thanks @tepreece!

PR #708, opened 2014-07-16
Add turnip style step definitions

tooky:
First stab at an implementation of turnip style StepDefs prompted by the [discussion](https://groups.google.com/forum/#!searchin/cukes/turnip/cukes/0j_5u9zkZng/1rbOOVDzHMcJ) on the mailing list.

A couple of questions

* Should we deprecate the old `$placeholder` style in cucumber ruby?
* Do we want to add turnip style custom placeholders - or leave people to rely on Transforms?
  * If we add custom placeholders, they should only affect turnip style stepdefs and Transforms only operate on RegexpStyle stepdefs - (I think this is the way to go)


mattwynne:
@tooky I'm tidying up. I suggest we close this and start from scratch. WDYT?

tooky:
Yes!

aslakhellesoy:
Related to #1002

PR #750, opened 2014-09-27
Remove unnecessary `multi_test` require

tomstuart:
This is a question in the form of a pull request: is the `require 'multi_test'` in `rb_language.rb` necessary? It already happens in `cucumber.rb` (via [`cucumber/runtime.rb`](https://github.com/cucumber/cucumber/blob/386c7b73523404139a4f35972f78a1261aaaac5c/lib/cucumber/runtime.rb#L4)), so it’s not obvious to me why it should be required here as well.

In an ideal world there’d only be one place where the assertion library is loaded, so I’m just double-checking that this duplication has intentional value.

### DIFF
--- a/lib/cucumber/rb_support/rb_language.rb
+++ b/lib/cucumber/rb_support/rb_language.rb
@@ -5,7 +5,6 @@ require 'cucumber/rb_support/rb_step_definition'
 require 'cucumber/rb_support/rb_hook'
 require 'cucumber/rb_support/rb_transform'
 require 'cucumber/rb_support/snippet'
-require 'multi_test'
 
 module Cucumber
   module RbSupport


mattwynne:
@tooky I'm concerned that the need to install an extra gem will add friction for newbies. Can you see a way we can avoid that?

tooky:
@mattwynne we could default to requiring rspec assertions if they are available, but allow users to require cucumber/minitest explicitly.

If cucumber/minitest is required we could prevent trying rspec?

That all seems quite complicated to maintain though?

Perhaps we could just output a warning at the start of a test run if no assertion library has been installed?

tooky:
@mattwynne what if the instructions we give newbies if they want the status quo is that they use `gem 'cucumber/rspec'` instead of `gem 'cucumber'`.

Minitest uses would just use `gem 'cucumber/minitest'`.

Those projects will depend on cucumber?

mattwynne:
It seems weird to me that you'd use a specific flavour of cucumber as your top-level dependency just to choose the assertion library. It's upside-down.

I think having `require 'cucumber/rspec'` or `require 'cucumber/minitest` is the right thing to do, and I think that if we want to test how those things behave in different scenarios, they need to be in one or more separate gems.

What is we compromise on having those `cucumber/rspec` and `cucumber/minitest` files in Cucumber, have them delegate to MultiTest, but get rid of the implicit inclusion of the helpers?

mattwynne:
This one has gone off-topic. 91fa6a49d0e7b87fafca8114373a4945721fb5fd means the require statement is only in one file, so I think it will make future hunts easier for you @tomstuart.

@tooky I agree we should get rid of the implicit loading, but let's deal with that in another ticket.

Issue #453, opened 2013-05-13
Same mutable string object is repeatedly passed to steps


jamespharaoh:
Strings for the arguments passed to steps are mutable string objects. So if I change one then that change persists across all future calls to the step.

For example:

``` ruby
Given "A file \"(.+)\"" do |file_name, file_contents|
    file_contents.gsub! "${placeholder}", @some_value
    File.open file_name, "w" { |file_io| file_io.write file_contents }
end
```

The ${placeholder} will be replaced by the value of @some_value the first time this step is run, and it keeps the same value forever after, even if @some_value is different for future test.

The simple workaround for this is not to modify the string, for example:

``` ruby
file_contents = file_contents.gsub "${placeholder}", @some_value
```

...but I keep making this mistake, and cucumber could certainly be more clever about this, and arguably should be.

The simple solution is just for cucumber to freeze the strings it uses. This will make my call to gsub! throw an error.

A slightly nicer solution might be to clone the strings when passing them into the steps.

I have had a look at the cucumber code to try and write a test and fix for this but I can't really see where this ought to go, so am submitting an issue instead. If someone can give me some pointers into the code I'll happily create a pull request.

PR #760, opened 2014-10-16
Each time a step_match is invoked it uses the same object as the step argument

tooky:
As reported by [Daniel Esponda on the ML](https://groups.google.com/forum/#!topic/cukes/66-hQThZpMU)

I've created an [example of this for cucumber 2.0](https://github.com/tooky/cucumber-reuse-params).

Cucumber 2.0 is different because the step arguments are frozen, but I do think its unexpected that we get the same `===` object each time a particular step match is invoked. Is it also slightly problematic if users can't modify step arguments because they are frozen?

### DIFF
--- a/lib/cucumber/step_match.rb
+++ b/lib/cucumber/step_match.rb
@@ -12,7 +12,7 @@ module Cucumber
     end
 
     def args
-      @step_arguments.map{|g| g.val.freeze }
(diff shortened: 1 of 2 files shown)


tooky:
@aslakhellesoy here's some history regarding deep clone of step arguments #453 and #1275 are interesting related issues

PR #765, opened 2014-10-27
fix snippet suggestions

richarda:
add more details to snippet scenario to create a failing step

### DIFF
--- a/features/docs/defining_steps/snippets.feature
+++ b/features/docs/defining_steps/snippets.feature
@@ -13,11 +13,21 @@ Feature: Snippets
         \"\"\"
           example with <html> entities
         \"\"\"
+        When a simple when step
+        Then a simple then step
       """
     When I run `cucumber features/undefined_steps.feature -s`
     Then the output should contain:
       """
       Given(/^a pystring$/) do |string|
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+      
+      When(/^a simple when step$/) do
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+
+      Then(/^a simple then step$/) do
         pending # Write code here that turns the phrase above into concrete actions
       end
       """
--- a/features/docs/defining_steps/snippets.feature
+++ b/features/docs/defining_steps/snippets.feature
@@ -14,6 +14,7 @@ Feature: Snippets
           example with <html> entities
         \"\"\"
         When a simple when step
+        And another when step
         Then a simple then step
       """
     When I run `cucumber features/undefined_steps.feature -s`
@@ -22,8 +23,12 @@ Feature: Snippets
       Given(/^a pystring$/) do |string|
         pending # Write code here that turns the phrase above into concrete actions
       end
-      
+
       When(/^a simple when step$/) do
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+
+      When(/^another when step$/) do
         pending # Write code here that turns the phrase above into concrete actions
       end
 
--- a/lib/cucumber/formatter/console.rb
+++ b/lib/cucumber/formatter/console.rb
@@ -138,11 +138,14 @@ module Cucumber
         return if undefined.empty?
 
         unknown_programming_language = runtime.unknown_programming_language?
+        previous_step_keyword = nil
         snippets = undefined.map do |step|
           # step_name = Undefined === step.exception ? step.exception.step_name : step.name
           # TODO: This probably won't work for nested steps :( See above for old code. 
           step_name = step.name
-          @runtime.snippet_text(step.actual_keyword, step_name, step.multiline_arg)
+          snippet = @runtime.snippet_text(step.actual_keyword(previous_step_keyword), step_name, step.multiline_arg)
+          previous_step_keyword = step.actual_keyword
+          snippet
         end.compact.uniq
 
         text = "\nYou can implement step definitions for undefined steps with these snippets:\n\n"
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -33,6 +33,7 @@ module Cucumber
         @img_id = 0
         @text_id = 0
         @inside_outline = false
+        @previous_step_name = 'Given'
       end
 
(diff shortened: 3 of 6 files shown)


richarda:
Still looking at this.  Making some progress adding some specs for the formatters (console.rb and html.rb)

richarda:
@mattwynne I'm a little confused about the debug_spec test that is failing.  When I run it individually, it passes, but it fails when all the formatter specs are run together or in `rake spec`
I think this may be getting closer to a fix here now if I've understood the model of the formatters correctly.
The tests that I added are passing as well as the initial failing scenario in features/docs/defining_steps/snippets.feature
Thanks!

richarda:
The current debug_spec test failure may be unrelated to this branch.
If I run:
`bundle exec rspec -f d --order rand:2 spec/cucumber/formatter/debug_spec.rb spec/cucumber/formatter/junit_spec.rb`
on the master branch, I get the same test failure.
running those specs with `--order rand:1` will pass
I'm not sure where or if there is a dependency between these two specs, but when debug_spec runs after junit_spec, debug_spec fails.

mattwynne:
Ugh, yeah. I can reproduce this in the master branch too, so it's nothing to do with your change.

It looks like any of the other formatter specs, if run first, will break this one. Some kind of state is bleeding over between them, but from a quick half-hour look over it, it isn't obvious where the problem is. :confused: 

I have been doing some work to simplify the runtime / mappings code in [this branch](https://github.com/cucumber/cucumber/tree/activate-steps-in-front-end) which will touch these unit tests. Hopefully that will resolve it.

Sorry about this @richarda - can you work around it for now?

richarda:
@mattwynne, would an acceptable work around be to figure out how to keep this debug_spec test running first?

mattwynne:
:clap:

It's dirty, but it works!

richarda:
@mattwynne do you think this PR needs anything else?

mattwynne:
I was waiting until we'd made some other changes in master, which are now done. There's a chance they've fixed the ordering thing. Can you try rebasing?

richarda:
Sure.  Just to clarify, do you all prefer rebasing or merging from the master into the feature/fix branch?

mattwynne:
A rebase is preferable - even if you force push. It's nice to have the PR looking neat and tidy with only commits that are relevant to this change before we merge that in.

So right now you could probably do a `rebase -i` and then a `git push --force` to your repo to tidy things up.

Make sense?

richarda:
Sure.  I'll do it.  I wasn't sure how github's pull request interface would
react to a force push.

mattwynne:
Brilliant! And the `debug_spec` hack went away too?

richarda:
That's right, the debug_spec hack wasn't needed anymore so I just dropped those 2 commits (adding and then removing the hack) during the rebase.
Thanks @mattwynne.  That was fun.  How else can I help?

mattwynne:
Thanks for the offer @richarda!

Any of these would be of help: https://github.com/cucumber/cucumber/issues?q=is%3Aopen+is%3Aissue+milestone%3A2.0

#768 should be quite easy.

brasmusson:
I looked through the changes in this PR the other day, and it occurred to me that given a scenario like:
```
Scenario:
  Given this step passes
  Then this step passes
  And this step is undefined
```
then the suggested snippet would use the `Given` keyword since it is the first undefined step, even though it actually is a `Then` step. I fixed that in https://github.com/cucumber/cucumber/commit/794bce14e86697562cd0e087e37199052cde95fe, so that the `previous_step_keyword` is updated also for defined steps. Ideally I would have reset the `previous_step_keyword` at the start of each test case, but I do not see how to determine when a new test case starts from `Console#print_snippets`. In case of the HTML formatter the `previous_step_keyword` is reset in `before_background` and in `before_feature_element`, a close enough approximation of the start of each test case, until the HTML formatter uses the new API.

While at it, I also fixed so the suggested snippets handle i18n languages (https://github.com/cucumber/cucumber/commit/cf0fff2f4fb0b8eeca0e73677eaa723bea6c8ddd)

mattwynne:
@brasmusson thanks for taking such careful notice!

Maybe you're aware already but, as of c6f662c1c0ad5eeff4ab449bdb6ed898189270ce you can use both APIs in a formatter, so the old formatters could listen for `before_test_case` or whatever in order to reset this state.

Does that help?

brasmusson:
@mattwynne Yes, I think I have seen something about the use of both APIs in a formatter, but No, I did not recall that at the time I did those changes. Using `before_test_case` in the HTML formatter to reset the `previous_step_keyword` at the start of each test case is straight forward (https://github.com/cucumber/cucumber/commit/b16a07eb982b65dca1cdd9095bfcd95223bed574).

`Console#print_snippets` gets the steps from the runtime, and I do not see how to determine from a set of `Cucumber::Core::Ast::Step` whether they belong to the same test case of not. The reset of `previous_step_keyword` in the formatter using the `Console`, seems to require that those formatter (Pretty and Progress) take control of the collection of undefined step to generate snippets for.

When considering using `after_test_step` for that, the location of `actual_keyword` in `LegacyApi::Ast::StepInvocation` seems a bit misplaced, since that logic then only is available to formatters using the legacy API. I moved it to `Cucumber::Core::Ast::Step`, where it can be accessed both from formatters using the legacy API (through `Legacy::Api::StepInvocation`) and formatter using the new API (through `Cucumber::Core::Test::Step`) https://github.com/cucumber/cucumber-ruby-core/commit/c33dd30ca9a97609bdb856ea25faaade99b38918.

With that in place I updated the Pretty and Progress formatter to collect the keywords and steps for the snippets, resetting the `previous_step_keyword` at the start of each test case. When the `Legacy::Api::StepInvocation` delegate to the `Cucumber::Core::Ast::Step` for the actual keyword logic, the passing language to the `Legacy::Api::StepInvocation` (just added) can also be removed, https://github.com/cucumber/cucumber/commit/d69fe32d9002011d6fd99a2e463a158f25c99879.

mattwynne:
Thanks for tidying this up @brasmusson, and for the thorough explanation.

Issue #781, opened 2014-12-11
Getting `undefined method 'unpack' for nil:NilClass` on ruby 2.2


myronmarston:
Ruby 2.2-preview2 was [released a couple weeks ago](https://www.ruby-lang.org/en/news/2014/11/28/ruby-2-2-0-preview2-released/) and I decided to add it to RSpec's travis build matrix to make sure we're all green on it (and, if not, to provide feedback to the ruby core team about any regressions we find).  However, I'm running into an exception from cucumber when travis runs our cukes on ruby-2.2-preview2.  (Actually, after looking at the full build logs, it looks like it's using 2.2-preview1...I guess `2.2` in `.travis.yml` doesn't yet run `preview2`).

Here's an example of the failure:

https://travis-ci.org/rspec/rspec-expectations/jobs/43763060#L155

Here's the full stack trace:

```
undefined method `unpack' for nil:NilClass (NoMethodError)
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/step.rb:81:in `text_length'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/step_collection.rb:62:in `block in max_line_length'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/step_collection.rb:62:in `map'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/step_collection.rb:62:in `max_line_length'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/has_steps.rb:53:in `max_line_length'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/has_steps.rb:49:in `source_indent'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/background.rb:40:in `accept'
/home/travis/build/rspec/bundle/ruby/2.2.0/gems/cucumber-1.3.17/lib/cucumber/ast/tree_walker.rb:64:in `block in visit_background'
[...]

PR #783, opened 2014-12-19
Use 2.2 on travis

tooky:
PR to have travis build against ruby 2.2.

If we have any issues we can fix them here and merge to master.

See #781


tooky:
closed by 3875ad2

Issue #768, opened 2014-11-05
cucumber2: scenario name with garbage


akostadinov:
In cucuber 1.x I see scenario.name == "scenario name"
In 2.x I see  scenario.name == "Scenario: scenario name"
In 2.x also scenario.title == "Scenario: scenario name"

IMHO one of these should be simple "scenario name"

Update: ditto Scenario outline - both name and title contain: "Scenario Outline: scenario name, Examples (row 1)"

Update 2: It is important for me to know which row of the outline is this and if it is an outline or just a scenario but would be nice to have raw name somewhere. Also a separate method to get row number would be nice. if this is not possible I can live with current implementation. I basically want to know all details. Just parsing strings is not very nice IMHO.

mattwynne:
The root of this is here: https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/case.rb#L93

brasmusson:
I realized that the keyword is still included in the `scenario.name` available to hooks (see [before_hooks.feature](https://github.com/cucumber/cucumber/blob/f5f8ae0635cbca28afc7b9841bec3019229c2a15/features/docs/writing_support_code/before_hook.feature)), In #814 there is a fix for this.

The checks in the Scenario Outline case is not very precise. The actual value of `scenario.scenario_outline.name` is `Scenario Outline: <scenario outline name>, <examples table name> (row <nr>)`, but since the check of newline is missing, the both scenarios in the feature file is passing.

Issue #82, opened 2011-06-08
JUnit reporter treating skipped steps as fails not skips


squeedee:
As the title says. In cucumber html output and on the command line, the tests are being skipped. In jUnit they're showing as failed.

Writing a cuke for it, will submit tommorow :D

PR #802, opened 2015-02-06
remove 'Scenario: ' prefix from expected output

richarda:
This is related to the [PR #82](https://github.com/cucumber/cucumber-ruby-core/pull/82) in `cucumber-ruby-core` that addresses issue #768.

Test::Case#name was changed to only output the name of the Scenario or Scenario Outline, not "Scenario: name"

This PR changes the expected output to reflect that change.  As noted in the other PR, I think this makes the output more compact while still being meaningful, but I could also prepare a change to keep the existing behavior in `cucumber`.

This PR will most likely fail the Travis build, as it's dependent on the other PR in `cucumber-ruby-core`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,6 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
   if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
     gem 'cucumber-core', :path => core_path
   else
-    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
+    gem 'cucumber-core', :git => "https://github.com/richarda/cucumber-ruby-core.git", :branch => '768-scenario-name'
   end
 end
--- a/lib/cucumber/formatter/legacy_api/adapter.rb
+++ b/lib/cucumber/formatter/legacy_api/adapter.rb
@@ -54,7 +54,7 @@ module Cucumber
         end
 
         def record_test_case_result(test_case, result)
-          scenario = LegacyResultBuilder.new(result).scenario(test_case.name, test_case.location)
+          scenario = LegacyResultBuilder.new(result).scenario("#{test_case.keyword}: #{test_case.name}", test_case.location)
           results.scenario_visited(scenario)
         end
 
--- a/lib/cucumber/ast/facade.rb
+++ b/lib/cucumber/ast/facade.rb
@@ -65,7 +65,7 @@ module Cucumber
         end
 
         def name
-          @test_case.name
+          "#{@test_case.keyword}: #{@test_case.name}"
         end
 
         def title


mattwynne:
Thanks! Shall we try getting the build to pass as we discussed?

I was thinking that we could add some extra logic [here](https://github.com/cucumber/cucumber/blob/master/Gemfile#L9) to use a branch for fetching the core if:

    ENV['CI'] == true

and

    `git rev-parse --abbrev-ref HEAD` != master

The name of the branch could, as you suggested, be conventionally the same in both PRs. Will that work, do you think?

brasmusson:
To remove the keyword form the listing of failing scenarios, will make the behavior of Cucumber v2.0 different from Cucumber v.1.3.x with respect to this ([feature from v1.3.x](https://github.com/cucumber/cucumber/blob/v1.3.x-bugfix/features/nested_steps.feature#L128)). As the general idea of Cucumber v2.0 is to be backward compatible, isn't it better to still include the keyword in this listings?

mattwynne:
Discussion here: https://twitter.com/mattwynne/status/563597525446569984

mattwynne:
Yes @brasmusson you're right.

mattwynne:
The change will need to be somewhere in the https://github.com/cucumber/cucumber/blob/master/lib/cucumber/formatter/legacy_api/adapter.rb to prepend the keyword to that string.

tooky:
Is the Ast Facade the correct place?

https://github.com/cucumber/cucumber/blob/master/lib/cucumber/ast/facade.rb#L67

I think we might need to expose keyword on the test case?

richarda:
@mattwynne I'll try some things [in here](https://github.com/cucumber/cucumber/blob/master/Gemfile#L9) and see if there is an extra check that will resolve the gem from git by branch name.  If it it works, I think it could be pretty nice.

richarda:
On the test case output, it seems like maintaining existing behavior is important.  I think I can work on that.

richarda:
@mattwynne - I was trying to add the branch dependency to get this passing as currently committed.
I think I was incorrect about the context Travis-CI uses to run the build.  I thought maybe I had a good way to get the PR owner/creator (in this case me), but I'm not sure if that's available once the build script is running `bundler`.  I don't think that's available from the git command at this point.
I wonder if travis-ci knows the name of who submitted the PR and happens to expose it in an environment variable.
As is, this doesn't work yet.

mattwynne:
@richarda for now, I'd say just change the git URL in the Gemfile to point to your fork / branch, and see if that makes it work. We can always revert that commit before merging this PR.

lib/cucumber/formatter/legacy_api/adapter.rb
```diff
@@ -54,7 +54,7 @@ def printer
         end
 
         def record_test_case_result(test_case, result)
-          scenario = LegacyResultBuilder.new(result).scenario(test_case.name, test_case.location)
+          scenario = LegacyResultBuilder.new(result).scenario("#{test_case.keyword}: #{test_case.name}", test_case.location)
```
richarda: Is there a better place to combine these?  It doesn't seem that nice to format that string in the argument to `#scenario`.  Would it be better to pass the whole `test_case`?  The formatting in the `#name` method in facade.rb seems ok.

richarda:
@mattwynne This got a little messy.  I can clean up the commits when this is finished, but for now, if you run these two branches locally, the tests should pass.  I'm having trouble getting the Travis build to pull from my fork / branch.
More to the point of all this, looking at the net changes, there a few simple changes and some new unit test expectations that seem to implement the original request and maintain backward compatibility.

Gemfile
```diff
@@ -6,6 +6,6 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
   if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
     gem 'cucumber-core', :path => core_path
   else
-    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
+    gem 'cucumber-core', :git => "https://github.com/richarda/cucumber-ruby-core.git", :branch => '768-scenario-name'
```
richarda: this is temporary while the corresponding pull request is pending.  I can run `bundle install`, and pull this dependency, but I can't seem to get it to work correctly on the Travis build.

richarda:
@mattwynne I reacted too quickly.  There aren't any build issues.  I needed to updated my `core` fork to meet the 1.1.0 version dependency in the `cucumber` Gemfile.  So the real discussion is probably about the changes in [this commit](https://github.com/richarda/cucumber/commit/ebb08f73ff2f83a061ccf214c0b52c08016b1655).
Do you think there is a better way or a better place to [format this string](https://github.com/cucumber/cucumber/pull/802/files#diff-7c15cef1ae22c03a391f089d20518b3cR57)?
There are no changes to any tests because we are maintaining existing behavior for backwards compatibility.
And you won't want the commit that changes the Gemfile.

lib/cucumber/formatter/legacy_api/adapter.rb
mattwynne: To be honest, I don't care too much about the code here in the legacy API adapter. We're planning to pull it out into a separate gem as soon as Cucumber 2.0 is released - I don't want to have to maintain this API anymore. So the code doesn't need to be perfect. There are lots worse problems in this file already! Just ask @brasmusson ;)

mattwynne:
Closed by ebb08f73ff2f83a061ccf214c0b52c08016b1655

PR #807, opened 2015-02-13
Around hooks not executing in correct world context

tooky:
This supersedes #640.

Based on @maxlinc's example this scenario demonstrates that the world an around hook is being instance_exec'd on is not quite right.

For the first scenario, self is nil, and the next scenario the world is remembered from the previous scenario.

### DIFF
--- a/features/docs/exception_in_around_hook.feature
+++ b/features/docs/exception_in_around_hook.feature
@@ -1,0 +1,80 @@
+Feature: Exceptions in Around Hooks
+
+  Around hooks are awkward beasts to handle internally.
+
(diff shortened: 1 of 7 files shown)


tooky:
@mattwynne using an around hook to prepare the world has changed the way some of the errors are reported. Is there anything that jumps out as something we need to do to change this?

The problem with preparing the world in a before hook is that they run after any around hooks, so the world that the rest of the test case is executed against has not been initialised yet.

mattwynne:
@tooky I think this is good to go. Can you tidy up the commits and merge (or use --squash) if you agree?

mattwynne:
thanks @tooky!

Issue #124, opened 2011-08-22
JUnit  formater doesn't report Scenario outline example with -x argument


floreal:
Hello,
I noticed a little annoying bug. consider my Rakefile:

```ruby
require 'rubygems'
require 'rake'
require 'cucumber/rake/task'

# some other tasks

namespace :cucumber do
  Cucumber::Rake::Task.new(:report) do |t|
    t.cucumber_opts = "-x -p #{browser} -f junit --out report -f pretty --no-color"
  end
  
  Cucumber::Rake::Task.new(:pretty) do |t|
    t.cucumber_opts = "-x -p #{browser}"
  end
end
```
The cucumber proffiles only define environment variable to set selenium url and desired browser capabilities

When i execure rake cucumber:report, only Scenarios are reported to the JUnit result report, Scenario Outlines examples aren't.
When I remove the -x flag they are reported.

mattwynne:
@brasmusson @richarda either of you fancy a crack at this one?

richarda:
Sure!

tooky:
Closed in favour of #808

PR #808, opened 2015-02-14
fix issue #124 - JUnit expand option

richarda:
@mattwynne Started looking at #124.  For now, this failing unit test highlights the problem.  I'm still looking at [formatter/junit.rb](https://github.com/cucumber/cucumber/blob/master/lib/cucumber/formatter/junit.rb) to see if that's the place to handle `@option[:expand]`

### DIFF
--- a/spec/cucumber/formatter/junit_spec.rb
+++ b/spec/cucumber/formatter/junit_spec.rb
@@ -3,193 +3,249 @@ require 'cucumber/formatter/spec_helper'
 
 require 'cucumber/formatter/junit'
 require 'nokogiri'
+require 'pry'
+module Cucumber
+  module Formatter
+    describe Junit do
+      extend SpecHelperDsl
+      include SpecHelper
 
-module Cucumber::Formatter
-  describe Junit do
-    extend SpecHelperDsl
-    include SpecHelper
+      class TestDoubleJunitFormatter < Junit
+        attr_reader :written_files
 
-    class TestDoubleJunitFormatter < Junit
-      attr_reader :written_files
-
-      def write_file(feature_filename, data)
-        @written_files ||= {}
-        @written_files[feature_filename] = data
+        def write_file(feature_filename, data)
+          @written_files ||= {}
+          @written_files[feature_filename] = data
+        end
       end
-    end
 
-    before(:each) do
-      allow(File).to receive(:directory?) { true }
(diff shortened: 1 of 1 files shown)


spec/cucumber/formatter/junit_spec.rb
```diff
@@ -3,193 +3,249 @@
...
+
+            FEATURE
+            # these type of tables shouldn't crash (or generate test cases)
+            it { expect(@doc.to_s).not_to match /milk/ }
+            it { expect(@doc.to_s).not_to match /cookies/ }
+          end
+        end
       end
+      
+      context "In --expand mode" do
```
richarda: This is where the new tests start.  Everything above was just moved into a context.

spec/cucumber/formatter/junit_spec.rb
```diff
@@ -4,192 +4,248 @@
...
-
-    class TestDoubleJunitFormatter < Junit
-      attr_reader :written_files
-
-      def write_file(feature_filename, data)
-        @written_files ||= {}
-        @written_files[feature_filename] = data
+module Cucumber
+  module Formatter
+    describe Junit do
```
richarda: changed `module Cucumber::Formatter` to
```
module Cucumber
  module Formatter
```
to get `runtime`

mattwynne:
I'm starting to feel guilty about pinging you about this one @richarda - it might well be too horrible to be worth it.

For version 3.0, I want to re-write the formatters to use the new, much simpler formatter API, precisely to get rid of bugs like this. So if it looks too tricky, we can just put it into the 3.0 milestone and expect it to get fixed then.

richarda:
Don't feel guilty @mattwynne ...
I have noticed the reports in our Jenkins builds have Scenario Outlines with missing details.  I'm glad to now know why :smile:

A silver lining:
I do think our team has a tendency to overuse the scenario outline anyway, so if this waits until 3.0, it will be a good reason to take a hard look at improving our features.  I'm sure there are good uses of the outline, but I often see a lot of [incidental details](https://cukes.info/school) in outlines, (redundant examples that don't help clarify behavior), which makes we wonder if the outline is a bit of a crutch for some people (including me).

brasmusson:
The actual fix of the issue is in https://github.com/cucumber/cucumber/commit/75b097507cefc6c5cfdd42ecb1c0392c5fa7c791.

PR #817, opened 2015-03-23
World reset in after hooks

tooky:
@mattwynne just noticed this problem as capybara session was getting
reset in after hook, so this no longer works.

I don't think we should pull this feature in - its more a test than
documentation :) but was a quick way for me to show the problem.

```ruby
After do |scenario|
  save_and_open_page if scenario.failed?
end
```

Possibly related to fix for #807.


tooky:
The step/scenario is also showing incorrectly as undefined with an error in an after hook.

tooky:
This isn't the bug I think I've found, will try and reproduce elsewhere.

PR #826, opened 2015-04-03
The check for multiple formatters using stdout is disabled when using profiles

brasmusson:
Somehow the check for multiple formatters using stdout is disabled when using profiles (even when the profile does not define any formatter). See failing scenario.

The root cause of this behaviour is the [logic for replacing the stdout formatter form the profile with the stdout formatter from the command line options](https://github.com/cucumber/cucumber/blob/13c621d2ea30733ddffdbfac42e2db0b39694f07/lib/cucumber/cli/options.rb#L366). Since the code remove all but the first stdout formatters, you can define several stdout formatters on the command line without getting an error when using profiles, but only the first one will actually be used. This behaviour cased some confusion [in this mailing list thread](https://groups.google.com/forum/#!topic/cukes/F2MM0pB9KTI).

To cover all cases, the check for multiple formatters needs to be performed multiple times:
* for the profile options and command line options individually, to catch multiple stdout formatters either in a profile or on the command line
* after merging the profile and command line options, to catch formatters from the profile and from the command line using the same file or directory for their output.

### DIFF
--- a/features/docs/cli/specifying_multiple_formatters.feature
+++ b/features/docs/cli/specifying_multiple_formatters.feature
@@ -3,8 +3,9 @@ Feature: Running multiple formatters
 
   When running cucumber, you are able to using multiple different
   formatters and redirect the output to text files.
+  Two formatters cannot both print to the same file (or to STDOUT)
 
-  Scenario: Multiple formatters and outputs
+  Background:
     Given a file named "features/test.feature" with:
     """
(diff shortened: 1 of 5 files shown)


brasmusson:
Now the check for multiple formatters using the same output stream/file/directory handle all cases.

mattwynne:
Looks good to me.

Issue #278, opened 2012-05-10
Cucumber::Ast::Table.diff! is broken when using no headers


troessner:
I have the following lines in one of my scenarios:

```
And I should see locations table                                             
  | Name    |      |      |
  | Berlin  | Show | Edit |
  | Hamburg | Show | Edit |
```

The step definition for this is (debugger added, see later):

``` Ruby
Then(/^I should see (.+) table$/) do |table_id, expected_table|
  html_table = table_at("##{table_id.parameterize.tableize}").to_a
  html_table.map! { |r| r.map! { |c| c.gsub(/<.+?>/, '').gsub(/[\n\t\r]/, '') } }
debugger
  expected_table.diff!(html_table)
end
```

Now, my table does exactly look like specified above (more no this later), but the step still fails:

```
Tables were not identical (Cucumber::Ast::Table::Different)
./features/step_definitions/common_steps.rb:18:in `/^I should see (.+) table$/'
features/manage_locations.feature:15:in `And I should see locations table'
```

The problem here is that some of the columns have an empty header and that's where the `diff!` method falls short.
Using the debugger I can see the following:

```
<(rdb:1) p expected_table.raw
[["Name", "", ""], ["Berlin", "Show", "Edit"], ["Hamburg", "Show", "Edit"]]
(rdb:1) p html_table
[["Name", "", ""], ["Berlin", "Show", "Edit"], ["Hamburg", "Show", "Edit"]]
(rdb:1) p expected_table.raw == html_table
true
```

So, as you can see, the "raw" representation of the expected_table is the very same as of the html_table but the step still fails.

**I think there are two things wrong here:**

1.) IMHO it shouldn't fail since the tables are identical regarding everything I can see.
2.) At least the error message should tell me what exactly is not identical. From the output above I have no chance to fix this.

I looked into the implementation of the diff!-method, but unfortunately this isn't really a beauty (no offense), so it's definitely no 5 minute job to improve this. (To be honest, I think the whole "diff!" method needs a huge refactoring - it's way to long, nothing is really abstract or loosely coupled and it's badly readable.)

**Possible cause:**

I think the problem is that the method calls `hashes` internally, which uses the table headers as keys, if more than one table header is not set -> boom, there goes one column.

A colleague of mine created a sample project to illustrate this: https://github.com/exviva/cucumber_table_diff

What do you think?
Am I missing something here?

PR #832, opened 2015-04-10
Bugfix: #278 Cucumber::Ast::Table.diff! is broken when using no headers

roschaefer:
Summary:

1. Bugfix: The former (intended) behaviour was to raise an error if more than one header column matches. 
Quickfix: ```raise ... if candidate_cols.size > 2``` ->```raise ... if candidate_cols.size > 1``` 

2. Change: From my point of view, duplicate header values are no problem, just match one after another.

### DIFF
--- a/lib/cucumber/multiline_argument/data_table.rb
+++ b/lib/cucumber/multiline_argument/data_table.rb
@@ -336,7 +336,7 @@ module Cucumber
         convert_columns!
 
         original_width = cell_matrix[0].length
-        other_table_cell_matrix = pad!(other_table.cell_matrix)
+        @cell_matrix, other_table_cell_matrix = pad_and_match(@cell_matrix, other_table.cell_matrix)
         padded_width = cell_matrix[0].length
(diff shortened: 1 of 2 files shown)


lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -563,32 +563,25 @@ def pad!(other_cell_matrix) #:nodoc:
...
         unmapped_cols = other_cell_matrix.transpose
 
-        mapped_cols = []
 
-        cols.each_with_index do |col, col_index|
-          header = col[0]
-          candidate_cols, unmapped_cols = unmapped_cols.partition do |other_col|
-            other_col[0] == header
-          end
-          raise "More than one column has the header #{header}" if candidate_cols.size > 2
```
roschaefer: what happens in case of ```candidate_cols.size == 2``` ?

mattwynne:
Thanks @roschaefer!

Issue #611, opened 2013-12-17
Cukes in subdirectories can't find steps


steveklabnik:
I searched for a previous issue and didn't find one, apologies if I'm bad at that.

I previous had a bunch of cukes in `features/whatever.feature`, and steps in `features/step_definitions/whatever.rb`. It's all good. And when I `bin/cucumber`, everything runs. Peachy! (the bin is from Bundler's binstubs feature. It also happens with `bundle exec` and just a plain `cucumber` in a fresh gemset.)

In an effort to organize my junk drawer of features, I moved a feature into a folder. `features/foo/whatever.feature`. `bin/cucumber` runs everything just fine. So I organized them all nice and neatly into sub-directories.

Another member of my team complained that `bin/cucumber features/whatever.feature` used to work, but now `bin/cucumber features/foo/whatever.feature` doesn't. It can't find the step definitions. I get a whole bunch of yellow in my output.

I couldn't find anything that said anything about it directly, but I remember reading _something_ that said that putting `.feature` files anywhere should work. So, this seems to be a bug.

I can make a sample repository if it's helpful, but I figured it was pretty straightforward.

mattwynne:
I want to push this back to 2.1, I think we can give some focus to the config generally in that release.

PR #842, opened 2015-04-24
Default require path to features unless specified explicitly.

pdswan:
In general there seem to be some weird dependencies (or lack there of) between
configuration options and information in `Cucumber::Cli::Configuration#feature_files`
that could probably be clarified with a little refactoring.

One example that this feature exposed directly is that
`Cucumber::Cli::Configuration#feature_files` does not
depend on `Cucumber::Cli::Configuration#feature_dirs` though their names
certainly imply dependence.

I had deleted `Cucumber::Cli::Configuration#feature_dirs` to eliminate
the confusion, but added it back in when I realized it was used in
`features/docs/post_configuration_hook.feature`.

I took a stab at re-architecting the configuration code awhile back, but got mired and gave up. Just realized I never submitted this pull request.

Looks like there's plans to give the configuration some attention for 2.1, but maybe this is still a useful step.

### DIFF
--- a/features/docs/cli/finding_steps.feature
+++ b/features/docs/cli/finding_steps.feature
@@ -1,0 +1,28 @@
+Feature: Loading the steps users expect
+  As a User
+  In order to run features in subdirectories without having to pass extra options
+  I want cucumber to load all step files
+
+  Scenario:
+    Given a file named "features/nesting/test.feature" with:
+      """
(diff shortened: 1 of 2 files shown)


pdswan:
#611

tooky:
Thanks for the contribution!

Issue #839, opened 2015-04-20
Re-write formatters against new API


mattwynne:
For 3.0 the https://github.com/cucumber/cucumber/blob/master/lib/cucumber/formatter/legacy_api/adapter.rb should be removed, and all formatters that ship with Cucumber should go through the new, test-case based API.

Altogether there are:
- ~~progress (#859)~~
- ~~pretty~~ (#1305)
- ~~html~~ (#1306)
- ~~json~~ (#851)
- ~~json_pretty~~ (is the same as json)
- ~~junit (#855)~~
- ~~stepdefs (#859)~~
- ~~usage (#859)~~
- ~~debug~~ can be removed - only for help writing old-API formatters
- ~~rerun~~ done in (83a0c75cbaa3ded4a5c2f596cbfdcd257fa73cb0)

I think the progress formatter is the place to start, as there's a lot of shared code in all the console formatters. If you want to tackle one, please create a ticket and reference this ticket.

mattwynne:
Pretty re-write investment may be best placed into https://github.com/cucumber/cucumber-pretty-formatter

mattwynne:
Charlie Rudolph has just dropped the pretty formatter from cucumber-js altogether, and just just pretty-printing failures.

As much fun as it is writing pretty formatters, I think we should give serious consideration to doing the same, and giving ourselves a clean slate for these.

mattwynne:
💖 🌈 hooray!

PR #851, opened 2015-05-16
Rewrite the Json Formatter to use the new formatter api 

brasmusson:
The main behavioural goal of the PR is to included results for test cases from Scenario Outlines also when the `--expand` option is not used (currently the `--expand` option is need to usable data when using Scenario Outlines).

A secondary behavioural goal is to align the produced Json file with what Cucumber-JVM produces, for the benefit of people that develop third party tools to be used together with the Cucumber version. The first tools that comes to mind are [the Masterthought tools](http://www.masterthought.net/section/cucumber-reporting) and [the Bamboo plugin](https://marketplace.atlassian.com/plugins/com.hindsighttesting.behave.cucumber-bamboo-plugin). This means that:
 * results for all executed steps are included (that means the Background steps are included each time they are executed), and
 * the executed hooks are reported (except for Around hooks, which does not seem to be included in the test case steps).

To get the location of the step definitions and hooks was somewhat challenging since the location of the Cucumber::Core::Test::Action is where the action block is create (in Cucumber::Runtime::SupportCode) and not the location hook/step definition itself). The Hooks are changed to hold the location of the hook method. In case of step definitions, the Json Formatter queries the Runtime for the location instead (same as the LegacyAdapter).

An additional benefit of this rewrite is that the Json Formatter now is independent of the Gherkin2 library.

Related to #839.

### DIFF
--- a/lib/cucumber/formatter/json.rb
+++ b/lib/cucumber/formatter/json.rb
@@ -1,19 +1,300 @@
-require 'cucumber/formatter/gherkin_formatter_adapter'
+require 'multi_json'
+require 'base64'
 require 'cucumber/formatter/io'
-require 'gherkin/formatter/argument'
-require 'gherkin/formatter/json_formatter'
 
 module Cucumber
   module Formatter
     # The formatter used for <tt>--format json</tt>
-    class Json < GherkinFormatterAdapter
+    class Json
       include Io
 
-      def initialize(runtime, io, options)
-        @io = ensure_io(io, "json")
-        super(Gherkin::Formatter::JSONFormatter.new(@io), false, options)
+      def initialize(runtime, io, _options)
+        @runtime = runtime
+        @io = ensure_io(io, 'json')
+        @feature_hashes = []
       end
+
+      def before_test_case(test_case)
+        builder = Builder.new(test_case)
+        unless same_feature_as_previous_test_case?(test_case.feature)
+          @feature_hash = builder.feature_hash
+          @feature_hashes << @feature_hash
+        end
+        @test_case_hash = builder.test_case_hash
+        if builder.background?
+          feature_elements << builder.background_hash
+          @element_hash = builder.background_hash
+        else
+          feature_elements << @test_case_hash
+          @element_hash = @test_case_hash
+        end
+      end
+
+      def before_test_step(test_step)
+        return if internal_hook?(test_step)
+        if hook?(test_step)
+          @step_or_hook_hash = {}
+          hooks_of_type(test_step) << @step_or_hook_hash
+          return
+        end
+        if first_step_after_background?(test_step)
+          feature_elements << @test_case_hash
+          @element_hash = @test_case_hash
+        end
+        @step_or_hook_hash = create_step_hash(test_step.source.last)
+        steps << @step_or_hook_hash
+        @step_hash = @step_or_hook_hash
+      end
+
+      def after_test_step(test_step, result)
+        return if internal_hook?(test_step)
+        add_match_and_result(test_step, result)
+      end
+
+      def done
+        @io.write(MultiJson.dump(@feature_hashes, pretty: true))
+      end
+
+      def puts(message)
+        test_step_output << message
+      end
+
+      def embed(src, mime_type, _label)
+        if File.file?(src)
+          content = File.open(src, 'rb') { |f| f.read }
+          data = encode64(content)
+        else
+          if mime_type =~ /;base64$/
+            mime_type = mime_type[0..-8]
+            data = src
+          else
+            data = encode64(src)
+          end
+        end
+        test_step_embeddings << { mime_type: mime_type, data: data }
+      end
+
+      private
+
+      def same_feature_as_previous_test_case?(feature)
+        current_feature[:uri] == feature.file && current_feature[:line] == feature.location.line
+      end
+
+      def first_step_after_background?(test_step)
+        test_step.source[1].name != @element_hash[:name]
+      end
+
+      def internal_hook?(test_step)
+        test_step.source.last.location.file.include?('lib/cucumber/')
+      end
+
+      def hook?(test_step)
+        ['Before hook', 'After hook', 'AfterStep hook'].include? test_step.source.last.name
+      end
+
+      def current_feature
+        @feature_hash ||= {}
+      end
+
+      def feature_elements
+        @feature_hash[:elements] ||= []
+      end
+
+      def steps
+        @element_hash[:steps] ||= []
+      end
+
+      def hooks_of_type(test_step)
+        name = test_step.source.last.name
+        if name == 'Before hook'
+          return before_hooks
+        elsif name == 'After hook'
+          return after_hooks
+        elsif name == 'AfterStep hook'
+          return after_step_hooks
+        else
+          fail 'Unkown hook type ' + name
+        end
+      end
+
+      def before_hooks
+        @element_hash[:before] ||= []
+      end
+
+      def after_hooks
+        @element_hash[:after] ||= []
+      end
+
+      def after_step_hooks
+        @step_hash[:after] ||= []
+      end
+
+      def test_step_output
+        @step_or_hook_hash[:output] ||= []
+      end
+
+      def test_step_embeddings
+        @step_or_hook_hash[:embeddings] ||= []
+      end
+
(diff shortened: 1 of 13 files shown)


aslakhellesoy:
With Gherkin3 coming soon I really want to completely redesign the formatter API to something simpler. I think the JSON format should only contain results and attachments, and no info from the source (gherkin, markdown, html).

Yep, markdown and/or html as an alternative to gherkin is coming!

We need to sync up on this. I'll post an invite to some future plans/design sessions we can run online (google hangouts).

So wrt this pr, let's hold off until we all have a clear vision of near-term direction and goals.

brasmusson:
This PR is actually a step towards Gherkin3, since it removes the dependency from Cucumber to the Gherkin2 formatters, so the argument can be turned around - let's merge this PR so we come closer to using the Gherkin3 parser in Cucumber (to among other things get away form the ["can't use Cucumber on x64 Windows"](https://github.com/cucumber/gherkin/issues/320) issues)

I'm all for a simpler, better Json result format, but do think that because it is lending itself for further tool processing and other tools already are using it, the current version needs to be deprecated and kept (for some time) in parallel with the new and improved Json result format.

mattwynne:
Thanks Bjorn. I'm right behind your strategy here, this is great work.

Is there anything specific you want me to review before we merge this in?

brasmusson:
Specific question 1, is about 73344a5, go from hook test_step to the hook location, as it is deeper in the code base, it is more important to get right.

lib/cucumber/formatter/json.rb
```diff
@@ -1,19 +1,280 @@
...
+      def add_match_and_result(step_source, result)
+        @step_or_hook_hash[:match] = create_match_hash(step_source, result)
+        @step_or_hook_hash[:result] = create_result_hash(result)
+      end
+
+      def create_match_hash(step_source, result)
+        if result.undefined? || hook_source?(step_source)
+          location = step_source.location
+        else
+          location = @runtime.step_match(step_source.name).file_colon_line
```
brasmusson: Specific question 2, this is the same as how the legacy adapter finds the location of step definition, but is it the way to go? Or should it be possible to go from test_step to action to step definition (similar as from test_step of a hook to the hook definition in 73344a5)?

lib/cucumber/formatter/json.rb
```diff
@@ -1,19 +1,280 @@
...
+      def prepare_world_hook?(test_step)
+        test_step.source.last.location.file.end_with?('cucumber/filters/prepare_world.rb')
+      end
+
+      def hook?(test_step)
+        hook_source?(test_step.source.last)
+      end
+
+      def hook_source?(step_source)
+        ['Before hook', 'After hook', 'AfterStep hook'].include? step_source.name
```
brasmusson: Specific question 3, is there or should there be a better way to differentiate hook-steps from step-definitions-steps? Like that whatever you get from `test_step.source.last` responds to `hook?` and `step?` ?

lib/cucumber/formatter/json.rb
```diff
@@ -1,19 +1,280 @@
...
+      def same_feature_as_previous_test_case?(feature)
+        current_feature[:uri] == feature.file && current_feature[:line] == feature.location.line
+      end
+
+      def first_step_after_background?(test_step)
+        test_step.source[1].name != @element_hash[:name]
+      end
+
+      def prepare_world_hook?(test_step)
+        test_step.source.last.location.file.end_with?('cucumber/filters/prepare_world.rb')
```
brasmusson: Specific question 4, this is a bit to tight coupling to the internal prepare-world-hook, but I do not think it should be included in the json-file.

spec/cucumber/formatter/json_spec.rb
```diff
@@ -0,0 +1,640 @@
...
+                      "result": {"status": "passed",
+                                 "duration": 1}}],
+                   "steps":
+                    [{"keyword": "Given ",
+                      "name": "there are bananas",
+                      "line": 4,
+                      "match": {"location": "spec/cucumber/formatter/json_spec.rb:552"},
+                      "result": {"status": "passed",
+                                 "duration": 1},
+                      "after":
```
brasmusson: Specific question 5, the json output is generally made consistent with what Cucumber-JVM produces, but here is uncharted territory, since Cucumber-JVM does not have AfterStep-hooks. I choose to model them as "after" hooks applied to "steps" (so that the three hook types are "before", "after" and "around", and that these types can be applied on different levels - well so far only one hook is applied to something other the a scenario/test case, the `AfterStep()` hook)

lib/cucumber/formatter/json.rb
mattwynne: Yes, I think we need a consistent way for you to find the location of the action of the step. I don't think you should need the `if` statement on like 168 - everything should quack the same.

lib/cucumber/formatter/json.rb
mattwynne: The trouble with having predicates like that is we'd really need to add them to every single AST object so that all source elements have the same protocol.

The intention is that you call `describe_source_to` passing some other object, and that object will receive `hook` or `step` messages depending on the source of the step. What is it that you need the `hook_source?` method for?

lib/cucumber/formatter/json.rb
mattwynne: It seems fair that any internal hooks should be ignored I think, so why not make it a bit more generic?

mattwynne:
Sorry it's taken me a while to give this attention @brasmusson. I think it's already a massive improvement as-is, so we can either merge it now and carry on discussing these minor refactoring points, or we can resolve them and then merge. You decide.

lib/cucumber/formatter/json.rb
brasmusson: For step_definition test_steps I first tried to get the location from the action of the test_step, but for step definition the location is always the line in `Cucumber::StepMatch` where the action is created from the step definitions, so I instead did what `LegacyApi::Adapter` did - asking the `Cucumber::Runtime::SupportCode` to get the StepMatch (with the correct location). When I came to hook, I had to "solve the problem the right way" - but then test_steps from step_definitions and hook are treated differently. With new support in Cucumber-Ruby-Core (https://github.com/cucumber/cucumber-ruby-core/commit/c595f23366ee1c2c4997bc734f8bf7cd53fed151) actions know the location of the step_definition of hook that they are based on, and test_steps can be queried for the location of their action, so no `if` statement is necessary any more (to pass the location of the step_definition when creating the action we easy when the support in Cucumber-Ruby-Core was in place 48c0f34).

lib/cucumber/formatter/json.rb
brasmusson: Ok, now all hooks which comes from files that include "lib/cucumber/" in their paths will be ignored.

lib/cucumber/formatter/json.rb
brasmusson: Sometimes it feel cumbersome having to write a visitor to let the subject describe it to, and then query the visitor when, "I just want to know X" (in this case if it is a test_step from a step of from a hook). I invented this method to find out whether it is a step or a hook here, and then also used it in the [JUnit Formatter](https://github.com/cucumber/cucumber-ruby/blob/9850d1db13551d35c93e84a6fb6badd658dde099/lib/cucumber/formatter/junit.rb#L99) and in the [Progress Formatter](https://github.com/cucumber/cucumber-ruby/blob/9850d1db13551d35c93e84a6fb6badd658dde099/lib/cucumber/formatter/progress.rb#L45). Yes, it should be changed, but I leave it for now.

mattwynne:
Don't forget to update History.md :)

(I wish we had a more automated way to do that!)

brasmusson:
I updated History.md in the merge commit (https://github.com/cucumber/cucumber-ruby/commit/a94e2ae597b3cad583a7b12b9a755066072cb974), personally I prefere "latest change first" but this time I put it last of the new features to be consistent with the rerun-entry.

lib/cucumber/formatter/json.rb
brasmusson: In https://github.com/cucumber/cucumber-ruby/commit/5462378b165c0ded654cec26ee6e47f83edadf9c a visitor is used to determine whether a test_step is based on a hook, and if so the type of the hook (no more using the name of the source step to detect hooks).

PR #855, opened 2015-05-24
Rewrite the JUnit Formatter to the new formatter Api

brasmusson:
The rewrite does also automatically fixed some problems regarding Scenario Outline test cases. Multiple failing test cases from the same Examples table, and pending and undefined test cases from Scenario Outlines in strict mode, have not been handled properly.

In addition the intercepted stdout and stderr data are moved from the testsuite elements (which representing Features) to the testcase element (fd5e38e). Stdout and stderr elements are already created (but empty) for testcase elements. It makes more sense to put the intercepted stdout and stderr in the testcase, and having empty stdout and stderr elements there (which stated "nothing was intercepted") seems to misinformation is anything.

The final change (4f3ebda) requires cucumber/cucumber-ruby-core#92, I think a Result#non_zero_exit_status? query can simplify logic not only in the JUnit Formatter, but also in other parts of the Cucumber code base.

Related to #839.

### DIFF
--- a/lib/cucumber/formatter/junit.rb
+++ b/lib/cucumber/formatter/junit.rb
@@ -27,6 +27,10 @@ module Cucumber
           start_feature(test_case.feature)
         end
         @failing_step_source = nil
+        # In order to fill out <system-err/> and <system-out/>, we need to
+        # intercept the $stderr and $stdout
+        @interceptedout = Interceptor::Pipe.wrap(:stdout)
+        @interceptederr = Interceptor::Pipe.wrap(:stderr)
       end
 
       def after_test_step(test_step, result)
@@ -41,6 +45,9 @@ module Cucumber
         scenario_designation = "#{scenario}#{test_case_name.name_suffix}"
         output = create_output_string(test_case, scenario, result, test_case_name.row_name)
         build_testcase(result, scenario_designation, output)
+
+        Interceptor::Pipe.unwrap! :stdout
+        Interceptor::Pipe.unwrap! :stderr
       end
 
       def done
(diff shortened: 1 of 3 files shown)


lib/cucumber/formatter/junit.rb
```diff
@@ -25,155 +21,134 @@ def initialize(runtime, io, options)
         @options = options
       end
 
-      def before_feature(feature)
-        @current_feature = feature
-        @failures = @errors = @tests = @skipped = 0
-        @builder = Builder::XmlMarkup.new( :indent => 2 )
-        @time = 0
+      def before_test_case(test_case)
+        if not same_feature_as_previous_test_case?(test_case.feature)
```
olleolleolle: [How to use Ruby's English and/or operators without going nuts](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/) explains why the `if not` here really should be `if !`

brasmusson: Shouldn't `unless` be as good as `if !`?

olleolleolle: Yes, you're right, in this case there is no `else`, and we are free to improve one step more.

(By preferring `unless` over `if !` and by not using `not` you now fixed two things.)

olleolleolle: In order to hear about code conventions, you can use [Rubocop](https://github.com/bbatsov/rubocop) on the files you've changed, and get a detailed report on any code-style deviations.

brasmusson: I reworked the commit and fixed all conditionals related Rubocop detected issues. I did not fix everything Rubocop remarked on. I usually think "evolution, not revolution", and when the code base generally doesn't seem to follow the "no lines longer than 80 characters", I do not go over the top to only write lines shorter than 80 characters either (for example).

PR #856, opened 2015-05-24
[cleanup] Nowadays, DocString compares to Strings

olleolleolle:
- In 2014, this was a trouble https://github.com/cucumber/cucumber/pull/669 but nowadays, String and DocStrings compare fine.


mattwynne:
Thanks Olle!

PR #858, opened 2015-05-27
AfterAll hooks

abezzub:
As suggested in https://github.com/cucumber/cucumber/issues/853 this PR first adds a failing `.feature` file. I will try to look at Cucumber code and make tests pass, but let me know if you have any suggestions about tests or next steps that I should take.

### DIFF
--- a/features/docs/writing_support_code/after_all_hooks.feature
+++ b/features/docs/writing_support_code/after_all_hooks.feature
@@ -1,0 +1,138 @@
+Feature: After All Hook
+
+  In order to extend Cucumber
+  As a developer
+  I want to run code after all Cucumber tests are finished
+
+  Background:
+    Given the standard step definitions
+
+  Scenario: AfterAll hook gets executed after all steps are executed
+    Given a file named "features/support/env.rb" with:
+      """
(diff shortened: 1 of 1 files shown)


mattwynne:
This spec looks great. Have you had a look at the code with a view to doing the implementation?

mattwynne:
Need any help @abezzub?

mattwynne:
Question for you @abezzub: What should happen if an exception occurs in the `AfterAll` hook?

abezzub:
Sorry, I got sidetracked on a different project. 

I would expect cucumber to return an error exit code in this case.

mattwynne:
Thanks @abezzub. Do you think we should present the exception in the output too? Would you like to sketch out a scenario for this?

wjlroe:
@mattwynne we have been bitten by this at work, I can pick this up and continue the work if that's the right thing to do?

mattwynne:
Thanks @wjlroe that would be very much appreciated. Remember we do need to consider the exit codes too, and add that to the acceptance test.

mattwynne:
Closing in favour of #1011

PR #860, opened 2015-05-31
Make the Rerun Formatter consistent with the exit code

brasmusson:
I think the invariant for the behaviour of the Rerun Formatter should be:
When the exit code is zero => the Rerun formatter output is empty
When the exit code is non-zero => the Rerun formatter output is not empty

This PR changes the behaviour the Rerun Formatter to fulfil this invariant.

The final change (fc356b1) requires cucumber/cucumber-ruby-core#92.


mattwynne:
This is going to need a little update to work since I tweaked the method name in https://github.com/cucumber/cucumber-ruby-core/pull/92

brasmusson:
So, it was easy enough to tweak the name used here to.

PR #864, opened 2015-06-05
Failing spec for html formatter scenario-level comment

smiller:
Between 1.3.19 and 2.0.0, if you have a comment in a feature file at the scenario level, the html formatter does not include it in the output.  This failing spec demonstrates that behaviour.


brasmusson:
Fixed by 95bcefa and https://github.com/cucumber/cucumber-ruby-core/commit/2c79be9e0a9f8aa481c9cd0a561a1c91bd707554

Issue #831, opened 2015-04-08
undefined method `map' for nil:NilClass (NoMethodError) in "print_snippets"


dg-ratiodata:
Hi @mattwynne,

I got another error. An instance variable seems to be nil where it should not be nil.

```bash
% bundle exec cucumber --format stepdefs | grep stdin
0.0000000 /^I close the stdin stream$/                                         # aruba-0.6.2/lib/aruba/cucumber.rb:121
undefined method `map' for nil:NilClass (NoMethodError)
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/console.rb:150:in `print_snippets'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/progress.rb:83:in `print_summary'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/usage.rb:69:in `print_summary'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/progress.rb:25:in `after_features'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/ignore_missing_messages.rb:10:in `method_missing'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/legacy_api/adapter.rb:110:in `after'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/legacy_api/adapter.rb:46:in `done'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/fanout.rb:16:in `block in method_missing'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/fanout.rb:15:in `each'
/home/user/.gem/ruby/2.2.1/gems/cucumber-2.0.0/lib/cucumber/formatter/fanout.rb:15:in `method_missing'
/home/user/.gem/ruby/2.2.1/gems/cucumber-core-1.1.3/lib/cucumber/core/test/runner.rb:35:in `done'
/home/user/.gem/ruby/2.2.1/gems/cucumber-core-1.1.3/lib/cucumber/core/filter.rb:61:in `done'
[...]

PR #868, opened 2015-06-14
Errors in formatters and LegacyApi can give weird error symptoms

brasmusson:
To illustrate this a added the explicit raising of an exception in Cucumber::Formatter::LegacyApi::Ast. The errors from one spec (junit_spec.rb) pinpoint this, but the errors from another (pretty_spec.rb) instead point to:
```
       undefined method `custom_profiles' for {}:Hash
     # ./lib/cucumber/formatter/console.rb:82:in `print_stats'
```

I guess that #831 is an expression of this too. The error that triggered the stacktrace of #831 is somewhere completely unrelated to the information in that stacktrace.

### DIFF
--- a/lib/cucumber/formatter/legacy_api/ast.rb
+++ b/lib/cucumber/formatter/legacy_api/ast.rb
@@ -216,6 +216,7 @@ module Cucumber
 
           def accept(formatter)
             formatter.before_table_row(self)
+            raise "Error raised in Cucumber::Formatter::LegacyApi::Ast"
             values.each do |value|
               formatter.before_table_cell(value)
               formatter.table_cell_value(value, status)


mattwynne:
I'm confused @brasmusson. What do we need to do with this PR?

brasmusson:
@mattwynne You should look at this as an Issue with an illustrating example, and not as an PR with a failing test. The problem is that you can't get from `undefined method `custom_profiles' for {}:Hash` to the `raise` in LegacyApi::Ast::DataTableRow::accept, can you?

I have seen this regularly when working on Formatters and LegacyApi. It is not that big problem when working with baby steps, when it occurs you just back down from the last baby step to get to a clean build and try again. But getting a `undefined method 'custom_profiles' for {}:Hash`, `undefined method 'map' for nil:NilClass` or `undefined method 'after_test_case' for nil:NilClass` isn't exactly helping you to realise "of course, I should use `step_source.location.line` not `step_source.line`" (or something).

We will never know what the root cause of #831 was, or if it is fixed. Even if we come across the exact stacktrace again, we will not know if it the same root cause of a different one.

Somewhere (one place high on my list of "the usual suspects", is somewhere close to the `fan_out` handling of formatters), exceptions are swallowed and the execution continued, instead of the failing fast that would have given an accurate error message and stacktrace.

mattwynne:
@brasmusson do you mind if we close this to tidy up?

brasmusson:
No, I do not mind closing this. This was meant as a reminder of that some issues may have backtraces that point to totally the wrong place compared to where the problem is located.

Issue #865, opened 2015-06-08
issues around treatment of rerun.txt - scenarios from same feature on multiple lines


mcrmfc:
Hi,

I am happy to add failing tests and/or pull request for this issue but just want to get a feel for whether this behavior would be considered a genuine bug or whether it's more 'we don't aim to support that'....

When using `parallel_tests` gem with `--group-by scenarios` you can end up with a rerun.txt file that looks something like this:

```
features/test.feature:3
features/test.feature:6
```

If you were not using `parallel_tests` it would look like this:

```
features/test.feature:3:6
```

There was a [change](https://github.com/cucumber/cucumber/commit/30483b9778b9339ff906872d982add386ec6f554) in [v1.3.19](https://github.com/cucumber/cucumber/compare/v1.3.18...v1.3.19) which caused the behaviour of cucumber when encountering a rerun file in the first format to change.

It now only runs the first scenario in this case.

Note: in versions prior to `1.3.19` although both scenarios were run there was another bug that meant when using junit reports that the first result was overwritten by the second scenario run.

The main thing I want to check is whether the core team think rerun files in the format above should be handled...an alternative would be to get `parallel_tests` to write out a rerun file in the second format (that's the current hack I am doing myself in the project I am working on).

I have a [repo](https://github.com/mcrmfc/cucumber-rerun-bug) with a minimal project to replicate this issue if that's of any use.

Thoughts appreciated, cheers.

mattwynne:
I think Cucumber should be able to understand both formats, so yes, a PR would be great.

Can you have a go at factoring out a class to model the Rerun file where we can isolate all this behaviour with specs?

mcrmfc:
Thanks Matt, no worries I will try and look at that.

That change in 1.3.19 actually appears to break cucumber for any rerun.txt with multiple lines.

mattwynne:
Thanks!

It's helpful if you can remember to close and issue if you replace it with a PR - one less bit of clutter in the list for us to deal with :smile:

brasmusson:
Or in the description of the PR use a [keyword for closing issues](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues) for instance `"closes #865"` (backticked here so it is not interpreted as GFM in this example), then the issue will be closed automatically when the PR is megerd.

mattwynne:
Yeah that’s the ninja move.

Vanuan:
[...]

PR #871, opened 2015-06-18
Model rerun file as a class and add tests - issue #865

mcrmfc:
Here is a starter for discussion around modelling the rerun file issue #865. See what you think and let me know. Cheers, Matt

### DIFF
--- a/lib/cucumber/cli/rerun_file.rb
+++ b/lib/cucumber/cli/rerun_file.rb
@@ -1,0 +1,29 @@
+module Cucumber
+  module Cli
+    class RerunFile
+      attr_reader :path
+
+      def self.rerun_file?(path)
+        path[0] == '@' && File.file?(real_path(path))
+      end
+
+      def self.real_path(path)
+        path[1..-1] # remove leading @
+      end
+
+      def initialize(path)
+        @path = self.class.real_path(path)
+      end
+
+      def features
+        lines.map { |l| l.scan(/(?:^| )(.*?\.feature(?:(?::\d+)*))/) }.flatten
+      end
+
+      private
+
+      def lines
+        IO.read(@path).split("\n")
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/cucumber/cli/rerun_file.rb
```diff
@@ -0,0 +1,29 @@
+module Cucumber
+  module Cli
+    class RerunFile
+      attr_reader :path
+
+      def self.rerun_file?(path)
```
mattwynne: if you change this to something more abstract like `#can_read?` then I think we could use this same pattern for the other feature-loading strategies, which would be nice.

mattwynne:
Looks great to me!

mcrmfc:
Thanks @mattwynne - changes made as suggested.

lib/cucumber/cli/rerun_file.rb
```diff
@@ -0,0 +1,29 @@
...
+      def self.real_path(path)
+        path[1..-1] # remove leading @
+      end
+
+      def initialize(path)
+        @path = self.class.real_path(path)
+      end
+
+      def features
+        lines.map { |l| l.scan(/(?:^| )(.*?\.feature(?:(?::\d+)*))/) }.flatten
```
pasku: lines.flat_map { |l| l.scan(/(?:^| )(.*?\.feature(?:(?::\d+)*))/) }

http://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-flat_map

mcrmfc: @pasku - agree that looks cleaner. The issue is this only flattens the arrays produced by `lines` and `scan`, but there are also captures which produce arrays and this doesn't deal with those (I think :confused:) ...can you think of anything else?

johnwake:
Good work @mcrmfc :bow:

tooky:
Thanks! @mcrmfc

PR #875, opened 2015-06-25
update badges

lucianosousa:
badges from codeclimate and travis are indicating to the old repo name. now, it's fixed!


tooky:
thanks!

Issue #869, opened 2015-06-16
cucumber --i18n <invalid_language> breaks


tejasbubane:
When I run `cucumber --i18n ja`, I get the proper japanese output.
But when I drop an invalid language it breaks:

```
tech-6@batman:~/calculator|⇒  cucumber --i18n ma
undefined method `[]' for nil:NilClass (NoMethodError)
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/gherkin-2.12.2/lib/gherkin/i18n.rb:93:in `initialize'
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/gherkin-2.12.2/lib/gherkin/i18n.rb:25:in `new'
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/gherkin-2.12.2/lib/gherkin/i18n.rb:25:in `get'
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/cucumber-2.0.0/lib/cucumber/cli/options.rb:374:in `list_keywords_and_exit'
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/cucumber-2.0.0/lib/cucumber/cli/options.rb:122:in `block (2 levels) in parse!'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1396:in `call'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1396:in `block in parse_in_order'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1383:in `catch'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1383:in `parse_in_order'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1377:in `order!'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1469:in `permute!'
/home/tech-6/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/optparse.rb:1491:in `parse!'
/home/tech-6/.rvm/gems/ruby-2.2.2/gems/cucumber-2.0.0/lib/cucumber/cli/options.rb:253:in `parse!'
[...]

PR #876, opened 2015-06-28
Fix issue 869

danascheider:
Here's a fix for issue #869, including passing Cucumber and RSpec tests. I ran all the tests and everything passes, but let me know if you see anything that needs fixing.

### DIFF
--- a/features/docs/writing_support_code/after_hooks.feature
+++ b/features/docs/writing_support_code/after_hooks.feature
@@ -11,7 +11,7 @@ Feature: After Hooks
   Background:
     Given the standard step definitions
 
-  Scenario Outline: Retreive the status of a scenario as a symbol
+  Scenario Outline: Retrieve the status of a scenario as a symbol
     Given a file named "features/support/debug_hook.rb" with:
       """
       After do |scenario|
--- a/features/lib/step_definitions/language_steps.rb
+++ b/features/lib/step_definitions/language_steps.rb
@@ -6,3 +6,7 @@ Then(/^cucumber lists all the supported languages$/) do
     expect(all_output.force_encoding('utf-8')).to include(language)
   end
 end
+
+Then(/^cucumber should print the message:$/) do |message|
+  expect(all_output).to include(message)
+end
--- a/features/lib/step_definitions/language_steps.rb
+++ b/features/lib/step_definitions/language_steps.rb
@@ -7,6 +7,6 @@ Then(/^cucumber lists all the supported languages$/) do
   end
 end
 
-Then(/^cucumber should print the message:$/) do |message|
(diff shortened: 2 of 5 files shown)


danascheider:
I'm looking into how to fix the JRuby failure my additions caused. I don't have experience working with JRuby, so if anybody knows off the top of their head what this null pointer error is about, please let me know so I can fix it faster.

tooky:
Hey @danascheider - thanks for taking a look at that.

I think the JRuby exception may be a different exception that is being thrown by gherkin when it tries unsuccessfully to initialize a new language.

I think the actual cause of the problem is this line here:

https://github.com/cucumber/gherkin/blob/master/lib/gherkin/i18n.rb#L92

It's checking for a `nil?` `@iso_code`, but I'm not sure that will ever be nil. I suspect it should raise "Language not supported" when `@keywords` is nil.

I don't know if it will be possible to release a fix for that in gherkin - I'm not 100% sure there is a plan to release any more.

cc @aslakhellesoy

danascheider:
Thanks, @tooky, I was able to make the builds pass by checking the validity of the language code before passing it along to Gherkin, and then give the table of allowed languages if the given code is invalid. That oughta fix it.

features/lib/step_definitions/language_steps.rb
```diff
@@ -6,3 +6,7 @@
     expect(all_output.force_encoding('utf-8')).to include(language)
   end
 end
+
+Then(/^the output includes the message "(.*)"$/) do |message|
```
mattwynne: This is a generic step that doesn't belong in 'language steps' - better to put it in `aruba_steps.rb` with the similar steps.

features/docs/writing_support_code/after_hooks.feature
```diff
@@ -11,7 +11,7 @@ Feature: After Hooks
   Background:
     Given the standard step definitions
 
-  Scenario Outline: Retreive the status of a scenario as a symbol
+  Scenario Outline: Retrieve the status of a scenario as a symbol
```
mattwynne: Please try and avoid doing this as part of an unrelated PR - better to put these kinds of changes into individual PRs, however small. :)

mattwynne:
Thanks @danascheider! I'll tidy up those little lose ends now it's merged into master.

features/docs/writing_support_code/after_hooks.feature
danascheider: Will do.

mattwynne: Thanks. I know how tempting it is to fix little things like this "while you're there" (and I very much appreciate you doing it!) but it makes it much easier for us to read / review a PR if it's focussed on just one change.

PR #878, opened 2015-06-29
Remove wire protocol to plugin

tooky:
Starting the work of moving the wire protocol into a plugin.

Right now the wire protocol is no longer checked, and ruby is the only supported 'language'. Wire protocol features are not run as part of the build anymore, but we should be able to use them to help drive development of a clean room version of the wire protocol using the new core api.

We can probably tidy up the Rb* namespace and remove some abstractions as we're no longer supporting multiple languages in cucumber itself. Maybe best to start inlining and look for a better organisation?


mattwynne:
I've started a new repo for the clean-room implementation: https://github.com/cucumber/cucumber-ruby-wire

mattwynne:
I think this is going to have to be 3.0 since people will need to update things in their code it make it continue working.

lib/cucumber/step_match_search.rb
```diff
@@ -0,0 +1,67 @@
...
+        )
+      )
+    end
+
+    class AssertUnambiguousMatch
+      def initialize(search, configuration)
+        @search, @configuration = search, configuration
+      end
+
+      def call(step_name)
```
mattwynne: @tooky I'm wondering whether it would be better to use the test_step rather than just step_name as the protocol here. This would allow us to pull out the event notification into another one of these decorators, and it also leaves this protocol a bit more flexible. Thoughts?

mattwynne:
OK there's so much good in this I'm merging it in.

Existing users of the wire protocol should not be affected, but I'm bumping the version to 2.2 as an indicator that we've changed things underneath.

PR #883, opened 2015-07-06
Update website and chat url

DaleGardner:
Update a couple gitter links and the main website URL :hatched_chick:


mattwynne:
:clap: thanks @DaleGardner, it's so nice having people fix things like this when we miss them.

PR #884, opened 2015-07-06
Use Gherkin3 instead of Gherkin2

brasmusson:
This PR make Cucumber-Ruby use Gherkin3 instead of Gherkin2. Some notes:
* Some minor files that were used from Gherkin2 (formatter/ansi_escapes.rb, formatter/escaping.rb, formatter/argument.rb, formatter/hashable.rb and one method from i18n.rb), are copied into [Cucumber::Gherkin](https://github.com/cucumber/cucumber-ruby/tree/c4c54d47b1dee5a51aed0c53517d46665cce2499/lib/cucumber/gherkin) to remove all dependencies to Gherkin2.
* To parse data tables and dynamic steps internal parts of Gherkin2 were used. Now [Cucumber::Gherkin::DataTableParser](https://github.com/cucumber/cucumber-ruby/blob/c4c54d47b1dee5a51aed0c53517d46665cce2499/lib/cucumber/gherkin/data_table_parser.rb) and [Cucumber::Gherkin::StepsParser](https://github.com/cucumber/cucumber-ruby/blob/c4c54d47b1dee5a51aed0c53517d46665cce2499/lib/cucumber/gherkin/steps_parser.rb) are used instead, but they also depend directly on internal parts of Gherkin3 - not the kind of dependencies you really want.
* ~~Escaping doc string separators (\"\"\") in doc strings does not work in Gherkin3, but Gherkin3 support two different doc string separator (""" and ```), so defining steps with doc strings in doc string does still work.~~
* The language defining "comment" at the start of a feature file is not represented as a comment in the AST in Gherkin3 (which is was in Gherkin2), so when the language is not 'en' LegacyApi::Adapter will add such a comment, so it is printed by the Pretty and HTML formatters. When the language is 'en' the language "comment" is never printed regardless whether the language was explicitly set to 'en' or not.
* Descriptions in Gherkin3 do not include any leading blank lines in descriptions, so regardless of the actual number of blank lines between the element and the comment text in the feature file, there will be no blank lines when printed by the Pretty formatter (changing that to another fixed number of blank lines is easy if preferred).

Depends on cucumber/cucumber-ruby-core#93

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
   if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
     gem 'cucumber-core', :path => core_path
   else
-    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
+    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git", branch: "integrate-gherkin3-parser"
   end
   gem 'gherkin3', github: "cucumber/gherkin3", branch: "integrate-ruby-core"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -8,5 +8,4 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
   else
     gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git", branch: "integrate-gherkin3-parser"
   end
-  gem 'gherkin3', github: "cucumber/gherkin3", branch: "integrate-ruby-core"
 end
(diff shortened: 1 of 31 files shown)


aslakhellesoy:
This sounds phenomenal! I'll try to take a deep look this week.

tooky:
This is great @brassmusson!

@aslak worth noting that cucumber/cucumber-ruby-core#93 does most of the integration with the gherkin3 parser.

cucumber-ruby-core is still where the compiler is, we haven't built pickles for ruby in gherkin3. We wanted to integrate the gherkin3 parser first and see about extracting the compiler from cucumber-ruby-core once we were sure it was working.

brasmusson:
Now this PR and cucumber/cucumber-ruby-core#93 are updated to work with the released Gherkin3 gem.

mattwynne:
Released in v2.1 today! Thanks @brasmusson you're a :star:

PR #890, opened 2015-07-13
Refactor constants to instance variables for backtrace filter ...

### LINKED ISSUE #886: Running cucumber with `Inprocess` from `aruba`
Hi, I merged cucumber/aruba#258 and released a pre-release 0.8.0.pre2. Unfortunately `cucumber` does not work well with this release. 

**The problem is this:**

* Cucumber uses `Inprocess` to make tests faster
* Cucumber reads in Feature-files somewhere in the code
* It seems like it uses the current working directory for this
* It outputs the file names of the steps as well

		Feature: test
		
		Scenario:               # features/test.feature:2
		-    Given this step fails # features/step_definitions/steps.rb:4
		+    Given this step fails # tmp/aruba/features/step_definitions/steps.rb:4
		
* The tmp/aruba directory is part of the output although Dir.chdir is used
* Using the "old" gem version the problem, the problem does not occure
* Re-Using parts of the old code does not solve the problem

**What does aruba do?**
```ruby
      def run!
        fail "You need to call Aruba.process.main_class = YourMainClass" unless self.class.main_class

        Dir.chdir @working_directory do
          before_run

		  # even change `PWD` now
          in_environment 'PWD' => @working_directory do
            self.class.main_class.new(@argv, @stdin, @stdout, @stderr, @kernel).execute!
          end

          after_run

          yield self if block_given?
        end
      end
```
I placed `Dir.chdir`-calls at different places, tried some code walking, but still I didn't have an idea why `cucumber` finds the feature-files, but also outputs the `tmp/aruba`-directory? Normally it complains about a missing `feature`-directory.

Any idea where to look for the code which uses `Dir.getwd`?

BTW:
Using `SpawnProcess` works like a charm.
Thanks a lot.

Cheers, Dennis

dg-ratiodata:
to make them evaluated on runime.

@brasmusson @mattwynne Otherwise `Dir.getwd` is not evaluated on runtime, but on load time. This is the missing piece of code to be changed to fix #886. @brasmusson Fixed the the other piece of code in refactoring #888.

### DIFF
--- a/lib/cucumber/formatter/backtrace_filter.rb
+++ b/lib/cucumber/formatter/backtrace_filter.rb
@@ -3,25 +3,25 @@ require 'cucumber/platform'
 
 module Cucumber
   module Formatter
-
     class BacktraceFilter
-      BACKTRACE_FILTER_PATTERNS = \
-      [/vendor\/rails|lib\/cucumber|bin\/cucumber:|lib\/rspec|gems\/|minitest|test\/unit|.gem\/ruby|lib\/ruby/]
-      if(::Cucumber::JRUBY)
-        BACKTRACE_FILTER_PATTERNS << /org\/jruby/
-      end
-      PWD_PATTERN = /#{::Regexp.escape(::Dir.pwd)}\//m
-
       def initialize(exception)
         @exception = exception
+        @backtrace_filter_patterns = \
+          [/vendor\/rails|lib\/cucumber|bin\/cucumber:|lib\/rspec|gems\/|minitest|test\/unit|.gem\/ruby|lib\/ruby/]
+
+        if(::Cucumber::JRUBY)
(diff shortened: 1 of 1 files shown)


mattwynne:
@dg-ratiodata it looks to me like `BACKTRACE_FILTER_PATTERNS` could remain a constant - no?

maxmeyer:
Yes. But why would you do so? It's not used anywhere else, but from an instance of the `BacktraceFilter`.

mattwynne:
I think it communicates intent better if only the things that need to be mutated, _can_ be mutated. It helps the `pwd_pattern` to stand out more so that people realise it's dynamically evaluated.

I wish this thing had some unit tests around it, but I don't have time to add them right now.

maxmeyer:
Thanks for merging. :-)

PR #891, opened 2015-07-14
Pass DocString as String

aslakhellesoy:
With this patch, `DocString` objects are now true `String` objects, and not a `Cucumber::MultilineArgument::DocString` object.

I understand that the intention with `Cucumber::MultilineArgument::DocString` is that it behaves just like a `String`, but it doesn't. For example, `expect(some_string).to eq(some_doc_string)` causes an unreadable error message with a `DocString`, but a nice one (with a diff) for a `String`.

Furthermore, `DocString` is an implementation detail of Cucumber that users shouldn't be concerned with. In DDD speak, these are two (or perhaps three) different bounded contexts: Parsing, (Execution) and user code.

A side-effect of this change is that the `content_type` attribute is no longer available. That's fine - I can't think of a reason why users would need this. The `content_type` is purely intended for tools that _render_ Gherkin, and I'm not aware of any tools using it yet.

When those tools exist, the Gherkin3 library should be used to parse the Gherkin document (without Cucumber), and the `content_type` will be available from the [Gherkin3 AST](https://github.com/cucumber/gherkin3#ast).

Finally, I removed the `doc_string` method from `World` - it's not needed, as passing a plain `String` is sufficient.

### DIFF
--- a/features/docs/defining_steps/nested_steps_with_second_arg.feature
+++ b/features/docs/defining_steps/nested_steps_with_second_arg.feature
@@ -34,7 +34,7 @@ Feature: Nested Steps with either table or doc string
 
       """
 
-  Scenario: Use #step with docstring
+  Scenario: Use #step with Doc String
     Given a step definition that looks like this:
       """ruby
       Given /two turtles/ do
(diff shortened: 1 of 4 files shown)


mattwynne:
Better to deprecate the method rather than remove it so we can release this soon.

I'm curious what motivated you to do this?

aslakhellesoy:
Ok, leave the `doc_string` method in if you think it'll bite people.

I was working on gherkin-lint, and wasn't happy with the feedback from Cucumber.

aslakhellesoy:
@mattwynne I've kept the `doc_string` method and deprecated it. Any objections if I merge this to master?

Issue #882, opened 2015-07-02
Rerun Failing Tests


mattwynne:
e.g. http://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

danascheider:
Anybody working on this?

richarda:
I started, but unfortunately I haven't been able to make time for it
recently. I'd be happy if someone else did, but if not, I'll check in next
week.

danascheider:
I think I can work on this. I'll take a look at your PR, @richarda, and go from there.

danascheider:
BTW, if anybody has any ideas on how to simulate a flaky test in the features, I'd be glad to hear them. It's trivial to make the feature under test fail or pass at random, but I'm still working out how to make it fail, then pass, as needed for the test. @richarda, I'm using pretty much your same feature file at this point. Any thoughts?

mattwynne:
You could use a global variable in the step def, a bit like we do here: https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/cli/randomize.feature

danascheider:
Thanks, @mattwynne, I'll go ahead and do it that way.

danascheider:
Any objections to my implementing this as a formatter?

brasmusson:
I think the event-API should be used (which @mattwynne change the fail-fast formatter to use https://github.com/cucumber/cucumber-ruby/commit/a2968211bfe13e29e771ac15d4ac454dcbbe5a61, #903). It is intended both for internal clients and external clients like user defined filters. Today filters generally live in [Cucumber::Filters](https://github.com/cucumber/cucumber-ruby/tree/master/lib/cucumber/filters), maybe we should create Cucumber::Listeners as a  home for event listeners?

mattwynne:
A simplistic solution would be to make this both a filter and an event listener - listen for failed test cases, them pump them back through the filter API for re-evaluation.

Where this falls down is that, presumably, we'll want to be able to report a test-case as passed if it's passed within the specified number of re-runs. To do that we'd need to reach deeper into things. I think this might need changes within the core to do it properly.

mattwynne:
@danascheider I think the place to start is with a failing feature anyway.

mattwynne:
> On 28 Sep 2015, at 07:47, Björn Rasmusson <notifications@github.com> wrote:
> maybe we should create Cucumber::Listeners as a home for event listeners?
> 
[...]

PR #895, opened 2015-07-17
add an initial feature for rerunning failing tests

richarda:
I'm taking a crack at implementing #882 with the outline/example from [Maven reruns](http://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html)

I thought that one particular example would be that you could get a pass out of this, but you should still see the error information.

I like the idea as a way to help get through a tough spot, but living with flaky tests should hurt a little bit :smile: 

I would love to get to adding some code if this is close to what people were looking for out of this enhancement.

What do you think @mattwynne ?

### DIFF
--- a/features/docs/cli/rerun_failing_tests.feature
+++ b/features/docs/cli/rerun_failing_tests.feature
@@ -1,0 +1,71 @@
+Feature:
+  Rerun gives you a way to get through flaky tests usually pass after a few runs.
+  This gives a development team a way forward other than disabling a valuable test.
+  
+  - Specify max rerun count in option
+  - Output information to the screen
+  - Output rerun information in test report
+  
+  Questions:
+  use a tag for flaky tests?  Global option to rerun any test that fails?
+  
+  Background:
+    Given a flaky step definition
+  
+  Scenario: run a flaky test
+    Given a file named "features/fail.feature" with:
+      """
+      Feature: Fails
+
+        Scenario: flaky test
+          Given a flaky step
+
+      """
+    When I run `cucumber features -q`
+    Then it should fail with:
+      """
+      Feature: Fails
+      
+        Scenario: flaky test
+          Given a flaky step
+             (RuntimeError)
+            ./features/step_definitions/steps.rb:4:in `flaky_pass'
+            ./features/step_definitions/steps.rb:1:in `/^a flaky step$/'
+            features/fail.feature:4:in `Given a flaky step'
+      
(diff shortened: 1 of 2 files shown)


features/docs/cli/rerun_failing_tests.feature
```diff
@@ -0,0 +1,71 @@
...
+        Scenario: flaky test
+          Given a flaky step
+
+      """
+    When I run `cucumber features --rerun_flakes=1 -q`
+    Then it should pass with:
+      """
+      Feature: Fails
+      
+        Scenario: flaky test
```
richarda: maybe a `@flaky` tag here?

features/docs/cli/rerun_failing_tests.feature
```diff
@@ -0,0 +1,71 @@
...
+
+        Scenario: flaky test
+          Given a flaky step
+
+      """
+    When I run `cucumber features --rerun_flakes=1 -q`
+    Then it should pass with:
+      """
+      Feature: Fails
+
```
richarda: maybe a `@flaky` tag here?

mattwynne:
Suggest we call it retry rather than rerun since rerun already has a meaning in our domain (the rerun formatter)

features/docs/cli/rerun_failing_tests.feature
```diff
@@ -0,0 +1,71 @@
+Feature:
+  Rerun gives you a way to get through flaky tests usually pass after a few runs.
+  This gives a development team a way forward other than disabling a valuable test.
+  
+  - Specify max rerun count in option
+  - Output information to the screen
+  - Output rerun information in test report
+  
+  Questions:
+  use a tag for flaky tests?  Global option to rerun any test that fails?
```
mattwynne: If we're going to reproduce the Maven behaviour (which is what I suggest we start with) then we should just have a global option.

mattwynne:
One concern I have for you with this is that there may be some yak-shaving to do in order to meddle around with how the summary results like `(1 passed, 1 flake)` are printed. Currently that code is pretty ugly and old, and has almost reached the top of my refactoring list. It might be sensible to do that first on master. But you could get on with figuring out how to do the retries, in any case.

richarda:
Thanks @mattwynne I'll take a look.  I won't spend much time on the output here, but the global option to run with retries for any failures makes sense.

mattwynne:
Superseded by #920

PR #896, opened 2015-07-24
Moved simplecov_setup into lib folder

tboerger:
As simplecov_setup is a runtime dependency of the cucumber bin script it
should not be placed into spec/ folder as there should be only test
parts be placed. The old code bites us at SUSE as all tests are split
into a testsuite package which results in a currently broken cucumber.


darix:
related to the discussion in https://github.com/cucumber/cucumber-ruby/commit/4b8343ab3d47c6f6a17faaf82a0b4fd30bb3f4ac

tboerger:
Build failed just because of coveralls API error.

mattwynne:
Thanks!

PR #900, opened 2015-08-05
Backport a version of Ruby > 2 Hash method

tooky:
This gives us way to use `Hash(object)` in cucumber including ruby 1.9.

I'm not sure if its a good idea to add this directly to Kernel though? There's a risk someone could use it in their app unknowingly and have passing tests but fail in production when cucumber is not required?

We could instead add an explicit `Cucumber::Hash` method to isolate this change against that?

### DIFF
--- a/lib/cucumber.rb
+++ b/lib/cucumber.rb
@@ -26,5 +26,18 @@ module Cucumber
       called_by = caller[1]
       warn("Deprecated: #{class_name}##{method} #{message}. Caller: #{called_by}")
     end
+
+    if Cucumber::RUBY_1_9
+      # Backported from Ruby 2.0 to 1.9
+      def Hash(other)
+        return {} if other.nil? || other == []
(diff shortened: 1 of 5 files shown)


mattwynne:
I agree with your concern @tooky - I think we should have a "no monkey patches" policy for any new code. Let's namespace it as you suggest.

mattwynne:
Do we still need this?

brasmusson:
To change back from `Hash.try_convert()` to `Hash()` as discussed in https://github.com/cucumber/cucumber-ruby/commit/4e51bafa7a405421acb0145e5223f770b36505c6, I think we need this.

tooky:
I'll do the namespace tonight and remove Hash.try_convert()

mattwynne:
How did that go?

brasmusson:
Hard to say, to get a proper Travis build https://github.com/cucumber/cucumber-ruby/commit/e567305ea75095fc5ec230257bc7cd33c79f58c2 need to be included - for instance by rebasing the pull request branch.

Issue #879, opened 2015-06-29
cucumber --fail-fast switch


danascheider:
Any interest in adding a `--fail-fast` switch like RSpec has, that would cause Cucumber to exit after the current scenario if the scenario fails? I've thought on a number of occasions that this functionality would be useful, so unless somebody has deliberately decided not to include it, I'd be glad to make a PR.

mattwynne:
Hi @danascheider - thanks for the offer that sounds like a great idea. Please make sure you add an acceptance test in `features/docs` that explains how it works as well as unit tests for the underlying code.

If you need any help with the implementation just give us a shout.

danascheider:
Will do, thanks @mattwynne.

danascheider:
@mattwynne - Am I correct in concluding that the CLI doesn't have access to the report object while the tests are running? I've been trying to have the runner check the report for failures while the suite is running, without much success. At first blush, it looks like the cucumber-ruby-core would have to be modified to allow that to happen, but intuitively that doesn't seem like a good idea, and I feel like the core and the CLI are better separated than that. Off the top of your head, is there something I'm missing, or would changing core be a valid implementation?

mattwynne:
Good question! The report is constructed [here](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L185) and we use the `Fanout` class to proxy the method calls to the report and forward them on to a collection of other reports. I think you could use that mechanism to add in a `FailFastReport` that would watch for any failing test steps and maybe throw an exception that we can trap in the CLI [here](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/main.rb#L28).

Does that make any sense? This code is still in a bit of a transitionary state as we tidy up post 2.0, so sorry it's not easier to navigate around.

danascheider:
Thanks! I've actually found the code pretty easy to work with for the most part - it was just that one thing I was unclear on.

danascheider:
@mattwynne - Would you suggest giving the aforementioned ++FailFastReport++ its own output stream? I was trying to avoid that because I wanted to make sure it played nice with the other formatters, but currently the way I'm handling the error it raises is forcing Cucumber to exit before they have a chance to print their output. I'll give it its own output stream if you think that would be a good idea, but I just wanted a sanity check. Sorry to draw this out, I've been really busy!

mattwynne:
[...]

PR #906, opened 2015-08-13
Add fail fast switch

danascheider:
Here are the changes I've made for the fail-fast switch in #879, @mattwynne. I ended up just setting `Cucumber.wants_to_quit` in the `FailFast` report. 

The problem I haven't figured out how to solve is, when I run the fail_fast features, the instance of Cucumber running the test also gets its `@wants_to_quit` set to `true`, with the result that running the passing `fail_fast.feature` invariably causes the entire thing to exit (with status 2, no less). I'll keep hammering away at this, so just let me know if any solutions spring readily to mind.

Thanks!


features/docs/cli/fail_fast.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Fail fast
+
+  The --fail-fast flag causes Cucumber to exit immediately after the first 
+  scenario fails.
+
+  Scenario: When a scenario fails
```
brasmusson: The annotation `@spawn` needs to be added to this scenario so that the cucumber instance started by the `When` step is not executed in the same process as the cucumber instance that rake has started. This problem reminds me of "singletons considered harmful" and a quote from one of the pages return from a search of that topic: "everything that can be done with Singletons, can be done better with dependency injection". @mattwynne `Cucumber.wants_to_quit` really?

danascheider:
Thank you, @brasmusson. I went ahead and added that tag.

danascheider:
Just an update - I'm working on fixing this but am unable to reproduce the failures locally (and it seems they are not consistent in Travis either). It looks like something I changed in runtime is interfering with some of the `WireSupport` functionality. This hasn't fallen off my radar, don't worry!

mattwynne:
The wire support tests can be a bit flakey, so it might be nothing to do with you. I've been very busy lately but one of us should be able to take a good look at this soon. Thanks @danascheider!

features/docs/cli/fail_fast.feature
mattwynne: Yeah totally agreed about the use of that global state @brasmusson - it's a legacy we're living with for now but I'd love to see a PR that gets rid of it. Should be much easier now that the behaviour is all centred in the `QuitFilter`.

mattwynne:
Merged, this should be released in 2.1, hopefully later today.

Thanks again @danascheider.

lib/cucumber/formatter/fail_fast.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Cucumber
+  module Formatter
+    class FailFast
+      def initialize(configuration)
+        @configuration = configuration
+      end
+
+      def after_test_case(test_case, result)
+        Cucumber.wants_to_quit = true unless result.ok? @configuration.strict?
```
mattwynne: I don't understand this line. Is it a typo? If I remove the `@configuration.strict?` bit all the tests pass.

mattwynne:
Having given this some focus today as I merged it in, I thought of a couple of new ways to implement this feature. @tooky @brasmusson @danascheider I'm interested in your thoughts here.

1) We could now use the new [Events](http://www.rubydoc.info/github/cucumber/cucumber-ruby/master/Cucumber/Events) and listen for the AfterTestStep event. But that would still mean we'd need to use the global `Cucumber.wants_to_quit` to tell the runner to stop. It's just a bit neater than using a formatter.

2) We could use a Filter to add (only in fail-fast mode) what are effectively AfterStep hooks that check for failure of the preceding step and raise a new type of exception. We can update the runner to respond to that type of exception by going into a new state that skips everything from then on.

I like (2) better. WDYT?

mattwynne:
I've gone ahead and implemented (1) #couldnthelpmyself.

mattwynne:
But I'd still like to consider doing (2)

lib/cucumber/formatter/fail_fast.rb
brasmusson: That because there are no test with a test case with an undefined step, then the test case result is "not ok" only in strict mode, and only then shall the execution be stopped.

danascheider:
@mattwynne I actually considered implementing something like (2) to begin with. I think it's a good idea because that way the exit status could be `1` instead of `2`, which has been bugging me a little about the way I wrote it. I did it the way I did mainly because I didn't want to block on doing it the most elegant way possible - figured it's better to get something that works and then tweak.

lib/cucumber/formatter/fail_fast.rb
mattwynne: Oh silly me! I'm old-school and I couldn't read the Ruby without parentheses.

danascheider: That's OK, I just remembered recently that you didn't have to use them, so I stopped using them just for the sake of novelty, really. I meant to use them in this code.

brasmusson:
@mattwynne You cannot use actual AfterStep hooks to do (2), since also hooks can fail. After hooks does not work either, since Around hook can still fail after the After hooks have executed. I think the idea of using exceptions to tell the runner that a test case has failed, is to complex - I mean the runner (I am talking about Cucumber::Core::Test::Runner) knows perfectly well if a test case has failed, it is actually passing the `after_test_case` message that the formatters will receive, so if the runner should be used to stop running test case, then it could be configured to a fail-fast mode before starting the execution of test cases.

I do not see the problem with the quit filter (except that is uses the global `Cucumber.wants_to_quit`), but why not use the event bus not only to enable formatters and plugins to receive event, but also to enable them to send events like `wants_to_quit`?

mattwynne:
Question for @danascheider - do you think it makes more sense to stop as soon as one test case fails, or should we continue but skip all the remaining ones (so we still get the same total number of scenarios)?

danascheider:
@mattwynne - I actually like the idea of skipping all the remaining ones. I often wish RSpec did that with its `--fail-fast` option.

PR #910, opened 2015-09-01
Fixed the backtrace_line method to prevent duplicate backtraces in the html report

pittgoose:
Fixed the backtrace_line method to prevent duplicate backtraces in the html report

### DIFF
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -590,7 +590,7 @@ module Cucumber
       end
 
       def backtrace_line(line)
-        line.gsub(/^([^:]*\.(?:rb|feature|haml)):(\d*).*$/) do
+        line.gsub(/\A([^:]*\.(?:rb|feature|haml)):(\d*).*\z/) do
           if ENV['TM_PROJECT_DIRECTORY']
             "<a href=\"txmt://open?url=file://#{File.expand_path($1)}&line=#{$2}\">#{$1}:#{$2}</a> "
           else


pittgoose:
Just to explain this a little better:
I'm experiencing an issue where the stacktrace on a failed test is showing multiple times in the html report. In 2011 there was a commit to cucumber which fixed this issue: https://github.com/cucumber/cucumber-ruby/commit/4e4839c143617f41cd17ad932870c6717ec258f1
This was then reverted back to the original version in September 2013 with this commit: https://github.com/cucumber/cucumber-ruby/commit/d827094c1fe5babacc15077762e757f8124efb9d

brasmusson:
The reverting that re-introduces this problem was done in #548, and it was done to make the backtrace lines clickable in TextMate. So this is not the change we want to do, but rather something that solves both the problem of duplicate lines in the backtraces, and still makes the lines clickable in TextMate.

pittgoose:
@brasmusson do you have any suggestions on what to do then? I don't have a Mac to test this on TextMate, we're working on Windows.
Also, am I the only one experiencing the backtrace duplication? I haven't seen any talk about this issue.

brasmusson:
@pittgoose I don't know what to do instead, but my point is that since this line has been changed back and forth - each time with with the intent to fix an issue - what stops that this line is changes back again in 6 months if this PR is merged?
Can you supply a screen shot or something to illustrate the problem you are experiencing? (or alternatively supply a failing test for the issue)

pittgoose:
Here's a screenshot of the issue we are seeing. As you can see, the stacktrace is repeated on the page 6 times.

pittgoose:
For the time being we have created a customer formatter, redefined this method, and inherited the Html class. Not the greatest solution but it's working for now.

mattwynne:
I think the root cause of this back and forth is people changing things without tests to pin down the behaviour they want.

@pittgoose can you please add some tests to https://github.com/cucumber/cucumber-ruby/blob/master/spec/cucumber/formatter/html_spec.rb to go with this PR? If you do that I'll happily break the TextMate behaviour until a TextMate user comes along and gives us a PR with some tests in it too.

josephks:
The problem is that if the if condition isn't met each time the regex matches it is replaced with the whole original line. This fixes the problem for me:

```
      def backtrace_line(line)
        if ENV['TM_PROJECT_DIRECTORY']
          line.gsub(/^([^:]*\.(?:rb|feature|haml)):(\d*).*$/) do
            "<a href=\"txmt://open?url=file://#{File.expand_path($1)}&line=#{$2}\">#{$1}:#{$2}</a> "
          end
        else
          line
        end
      end
```

mattwynne:
I do appreciate your contribution @pittgoose but without a test we can't merge this change, sorry!

Issue #821, opened 2015-03-27
Suites


mattwynne:
See https://github.com/Behat/Behat/blob/master/features/suite.feature

``` ruby
Cucumber.configure do |config|
  config.suite(:domain) do |suite_config|
    suite_config.world_modules << DriveDomain
    suite_config.filters << TagFilter.new("~@web")
    suite_config.filters << LocationFilter.new("features/core")
  end

  config.suite(:web) do |suite_config|
    suite_config.world_modules << DriveUI
    suite_config.filters << TagFilter.new("@web")
  end
end
```

This would run the features twice, applying the specified filters and mixing in the specified world modules. The overall result would aggregate results from both suites.

aslakhellesoy:
Want want!

richarda:
@mattwynne I'm pretty interested in this.  In an effort to minimize UI testing, I've been trying to carefully word steps so that they imply either a domain model, like a controller with no view attached or a fully connected app with a clickable view and controller, but I feel like I'm dancing around something more fundamental and leaving a confusing trail of nuanced wording.

For example, a step:
`When I choose to "Cancel"`
could either mean using an automation framework to click a button called `Cancel` or just calling `cancel()` on a controller directly.  At this point, to distinguish, I might put something like `When I select "Cancel"` vs `When I choose to "Cancel"` so that I could resolve to different step definitions.  That doesn't feel

Is that similar to your example?

mattwynne:
Yes similar @richarda. This is getting off-topic a bit, but what are you trying to cancel? I'd try and express the intent behind the step - the action the user is trying to perform, and try to stop thinking about the interaction. I find that helps me name the methods on my lower level APIs.

We're going to explore this some more in the last Cucumber School, so stay posted for that!

richarda:
Thanks @mattwynne .  You raise a very good point.  Taking a closer look, there were 2 slightly different intents.
 - One intent is to make the application go away or hide from view when you didn't want to use it any more.  This scenario was implemented with UI steps to make sure that the app got hidden away.
 - Another intent is to clear a list of selected items when you don't want to follow through with the series of tasks.  This scenario uses steps that just create a controller (view model), fill a list then cancel, but no UI elements are needed to verify this behavior.

Maybe there is a better scenario that covers all of that together, but it seemed valuable to try and minimize the UI interactions and be a little more granular with the behaviors.

[...]

PR #917, opened 2015-09-24
Suites

richarda:
Start with a simple example for a suites feature discussed in issue #821.

### DIFF
--- a/features/docs/suites.feature
+++ b/features/docs/suites.feature
@@ -1,0 +1,54 @@
+Feature: Suites
+  In order to use different worlds for executing scenarios,
+  suites can be configured to run certain tests in a specifc world.
+  Rules:
+    - multiple suites can be configured with aggrated the results
+  
+  Background:
+    Given Cucmber is configured with an ios suite and android suite
+    
+  Scenario: different automation for iOS and Android  
+    Given a file named "features/image_sharing.feature" with:
+      """
+      Feature: Sharing an image
+
+        Scenario: regular sharing
+          Given I have taken a picture
+          When I share it with my friend
+          Then my friend can see the picture    
+
+        @ios-only
+        Scenario: automactially add to iCloud
+          Given I have taken a picture
+          When I share it with my friend
+          Then my iCloud album contains the picture
+      """
+      When I run `cucumber -q`
+      Then it should pass with exactly:
+        """
+        Suite: ios (~@android-only)
+        Feature: Sharing an image
+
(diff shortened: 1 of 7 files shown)


mattwynne:
Still got any energy for this @richarda?

richarda:
Sure. I've been out for a while, but I should jump back in. Let me take a look.

richarda:
I am looking at this.  It's taking me a while to catch up.  Would the execution behavior of a suite be implemented similarly to a rerun scenario?

mattwynne:
I think the changes required here are a bit more fundamental @richarda. The code is going to offer some resistance to this change but once it's done we'll have tackled some of the remaining bits of coupling, so you'll need to be brave!

First off, I think we need to model the configuration differently, so that there's configuration for the overall run that won't change (formatters, for example) and then configuration per suite (e.g. tags to filter by).

Then it's a matter of breaking out some of the functionality in `Runtime` into a new `Suite` class that models the execution of a single suite. The `Runtime` would have a list of suites (by default, just one) that it iterates over and executes. The body of `Runtime#run!` and its friends would move into the `Suite` class.

I would start by picking off a single configuration option (like the tags to filter by) and trying to get that to work, then we can iterate and pick off other configuration options that make sense.

Does that help?

richarda:
Yes @mattwynne. That is helpful. That seems like a very good way to get started.

richarda:
Sorry if this is slow progress, but I think I'm starting to get the idea here.
I'm just putting the configuration step inline in `runtime.rb` temporarily but it seems like there should be a method `Cucumber.configure` coming out of this soon.

In what context would this be written?
```
Cucumber.configure do |config|
  config.suite(:domain) do |suite_config|
    suite_config.world_modules << DriveDomain
    suite_config.filters << TagFilter.new("~@web")
    suite_config.filters << LocationFilter.new("features/core")
  end

  config.suite(:web) do |suite_config|
    suite_config.world_modules << DriveUI
    suite_config.filters << TagFilter.new("@web")
  end
end
```
Does that get written somewhere in the support directory like `env.rb`?

mattwynne:
You could actually use an `AfterConfiguration` block for now, and yes, call it from `env.rb`

richarda:
Still working ... Update soon.

mattwynne:
Closing due to inactivity. This was a tough one to tackle @richarda. Feel free to ping us if you'd like to get involved again with something easier!

PR #918, opened 2015-09-25
Output bug fix.

### LINKED ISSUE #915: Cucumber overwrites variable values in output
I have long known that 'puts' statements inside of a step definition are delayed until the step is complete. That is, you will go from seeing no output from a step to all of it at once when the step has finished running. I assume that this is the intended behavior of the pretty formatter.

Apparently, however, the formatter does not determine the value of the variable that it is printing out until the end. This results in incorrect output if, for example, I am in the habit of modifying a variable several times in a step definition and put-ing its value to the console so that I can debug something. All displayed values will be the same value instead of displaying the several different values that the variable went through.

See this gist to recreate the problem:
https://gist.github.com/enkessler/18c4837c865c59a287dc

(Note: I used cucumber 2.0.2)

enkessler:
Fixed a bug that was causing the delayed output of #puts calls to be
affected by modifications to its arguments that occurred between the
time #puts was called and the time that its output was actually
displayed.

Fixes #915.

### DIFF
--- a/lib/cucumber/rb_support/rb_world.rb
+++ b/lib/cucumber/rb_support/rb_world.rb
@@ -94,9 +94,10 @@ module Cucumber
       #
       #   If you'd prefer to see the message immediately, call {Kernel.puts} instead.
       def puts(*messages)
-        # Copying the objects because they may be modified after #puts is called and their
-        # original value needs to be retained for later output.
-        messages.collect!{|message| Marshal.load(Marshal.dump(message))}
+        # Even though they won't be output until later, converting the messages to
+        # strings right away will protect them from modifications to their original
+        # objects in the mean time
+        messages.collect! { |message| message.to_s }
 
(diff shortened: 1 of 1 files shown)


enkessler:
On second thought, duplication via marshaling may have more problems than the test failure caused by #dup/#clone because it can't apply to all objects.

enkessler:
This should do the trick.

brasmusson:
Merged. I added a spec in https://github.com/cucumber/cucumber-ruby/commit/d5e0cfade9dab3fca21e7c135b284a2b96250576.

@enkessler I tip for PR:s that fixes issues it to include `"fixes #<nr>"` (of any of the other [keywords for closing issue](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues)) in the PR description (as I added in the description above), then GitHub will close the issue when the PR is merged.

Issue #251, opened 2012-03-20
Output table padding formatter goes wrong with UTF8 characters


darkskiez:
The padding of the |'s goes wrong for the chinese input example.

``` gherkin
Feature: Multilingual

  Cucumber output pads columns example

  Scenario Outline: Language Example
    Given dummy step <language>

  Examples:
        | language  | labelled  |
        | English   | OK        |
        | 日本語    | 确定      |
        | Español   | Aceptar   |
```

PR #919, opened 2015-10-02
Format tables including full width characters properly

soutaro:
This PR is to format tables including fullwidth characters properly based on Unicode East_Asian_Width informative property.

Pretty formatter does not print tables including fullwidth characters properly as discusses in https://github.com/cucumber/cucumber-ruby/issues/251. It prints as

```
  Scenario: Test   # features/table.feature:2
    Given A table: # features/step_definitions/table_steps.rb:3
      | Cucumber |
      | キュウリ     |
```

where we want 

```
  Scenario: Test   # features/table.feature:2
    Given A table: # features/step_definitions/table_steps.rb:3
      | Cucumber |
      | キュウリ  |
```

This is because the text width is calculated by number of Unicode code points and does not consider character width at all. This patch uses `unicode` gem to calculate text width if the library is loaded. (I am not very sure this improvement matters to have another dependency.)

The `Cucumber.treats_ambiguous_as_fullwidth` options is to control the width of _ambiguous_ characters like `ø`. I don't think it is the best place to have that option.

### DIFF
--- a/cucumber.gemspec
+++ b/cucumber.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'coveralls', '~> 0.7'
   s.add_development_dependency 'syntax', '>= 1.0.0'
   s.add_development_dependency 'pry'
+  s.add_development_dependency 'unicode'
 
   # For Documentation:
   s.add_development_dependency 'bcat', '~> 0.6.2'
--- a/lib/cucumber.rb
+++ b/lib/cucumber.rb
@@ -9,6 +9,7 @@ require 'cucumber/term/ansicolor'
 module Cucumber
   class << self
     attr_accessor :wants_to_quit
+    attr_accessor :treats_ambiguous_as_fullwidth
 
     def logger
       return @log if @log
--- a/lib/cucumber/formatter/pretty.rb
+++ b/lib/cucumber/formatter/pretty.rb
@@ -203,7 +203,8 @@ module Cucumber
         status ||= @status || :passed
         width = @table.col_width(@col_index)
         cell_text = escape_cell(value.to_s || '')
-        padded = cell_text + (' ' * (width - cell_text.unpack('U*').length))
+        text_width = defined?(Unicode) ? Unicode.width(cell_text, Cucumber.treats_ambiguous_as_fullwidth) : cell_text.unpack('U*').length
+        padded = cell_text + (' ' * (width - text_width))
         prefix = cell_prefix(status)
         @io.print(' ' + format_string("#{prefix}#{padded}", status) + ::Cucumber::Term::ANSIColor.reset(" |"))
         @io.flush
--- a/lib/cucumber/multiline_argument/data_table.rb
+++ b/lib/cucumber/multiline_argument/data_table.rb
@@ -692,7 +692,18 @@ module Cucumber
         end
 
         def width
-          map{|cell| cell.value ? escape_cell(cell.value.to_s).unpack('U*').length : 0}.max
+          map {|cell|
+            if cell.value
+              escaped = escape_cell(cell.value.to_s)
+              if defined?(Unicode)
+                Unicode.width(escaped, Cucumber.treats_ambiguous_as_fullwidth)
+              else
+                escaped.unpack('U*').length
+              end
+            else
+              0
+            end
+          }.max
         end
       end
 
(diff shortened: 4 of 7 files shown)


spec/cucumber/formatter/unicode_pretty_spec.rb
```diff
@@ -0,0 +1,85 @@
+# coding: utf-8
+
+require 'spec_helper'
+require 'cucumber/formatter/spec_helper'
+require 'cucumber/formatter/pretty'
+require 'cucumber/cli/options'
+
+unless Cucumber::JRUBY
```
mattwynne: Why this logic? Does the unicode gem not work for JRuby?

mattwynne:
@soutaro can you explain why we need the `treats_ambiguous_as_fullwidth` option?

spec/cucumber/formatter/unicode_pretty_spec.rb
soutaro: Yes, `Unicode.width` raises `NotImplementedError` with JRuby.

https://github.com/blackwinter/unicode/blob/24137f1565377c42e423144fbb8d9fa94264dc75/lib/unicode-java.rb#L113-L115

soutaro:
The reason is mainly for compatibility with terminal apps. At least, Terminal.app in Mac OS X have an option to control that. (I'm using with the default setting, disabled.)

(I'm not sure how many people are using the option. It may be okay to remove that...)

cucumber.gemspec
```diff
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'coveralls', '~> 0.7'
   s.add_development_dependency 'syntax', '>= 1.0.0'
   s.add_development_dependency 'pry'
+  s.add_development_dependency 'unicode'
```
mattwynne: I'm wondering why this is a development dependency. Won't we need it at runtime too? Or is the idea that we'll expect users to have to have installed it themselves if they want to benefit from it?

lib/cucumber.rb
```diff
@@ -9,6 +9,7 @@
 module Cucumber
   class << self
     attr_accessor :wants_to_quit
+    attr_accessor :treats_ambiguous_as_fullwidth
```
mattwynne: I think having this option is fine, but we need to add it to https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/configuration.rb instead.

This has implications for how we use it in the `DataTable`, but we can figure that out.

lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -692,7 +692,18 @@ def index
         end
 
         def width
-          map{|cell| cell.value ? escape_cell(cell.value.to_s).unpack('U*').length : 0}.max
+          map {|cell|
+            if cell.value
```
mattwynne: let's extract a `Cell#width` method here and call it.

lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -692,7 +692,18 @@ def index
...
+              escaped = escape_cell(cell.value.to_s)
+              if defined?(Unicode)
+                Unicode.width(escaped, Cucumber.treats_ambiguous_as_fullwidth)
```
mattwynne: Instead of using the global option here, let's accept this specific option as a constructor parameter (using an options hash or keyword argument would be best IMO) into the `DataTable` and pass it down to the `Cell`.

Wherever we construct instances of `DataTable` we'll almost certainly have access to an instance of the `configuration` where we can read that setting from.

lib/cucumber/formatter/pretty.rb
```diff
@@ -203,7 +203,8 @@ def table_cell_value(value, status)
         status ||= @status || :passed
         width = @table.col_width(@col_index)
         cell_text = escape_cell(value.to_s || '')
-        padded = cell_text + (' ' * (width - cell_text.unpack('U*').length))
+        text_width = defined?(Unicode) ? Unicode.width(cell_text, Cucumber.treats_ambiguous_as_fullwidth) : cell_text.unpack('U*').length
```
mattwynne: This code looks quite similar to what's in `DataTable::Cells#width` below. If we move that logic into a `Cell#width` method, could we re-use that logic here by constructing an instance of a `Cell`?

lib/cucumber.rb
mattwynne: I also suggest we rename it to `treat_ambiguous_width_characters_as_double_width` to be crystal clear.

PR #920, opened 2015-10-03
Add --retry option to retry failed tests as part of the same run

danascheider:
Here are the features I wrote up for #882. Let me know what you think, @mattwynne and @brasmusson. Travis isn't passing but it's those wire protocol features again - shouldn't have been affected by what I wrote.

### DIFF
--- a/features/lib/step_definitions/cucumber_steps.rb
+++ b/features/lib/step_definitions/cucumber_steps.rb
@@ -32,6 +32,70 @@ Given /^a step definition that looks like this:$/ do |string|
   create_step_definition { string }
 end
 
+Given /^a scenario "([^\"]*)" that fails once, then passes$/ do |name|
+  write_file "features/#{name}.feature",
+  <<-FEATURE 
+  Feature: #{name}
+    Scenario: #{name}
+      Given it fails once, then passes
+  FEATURE
+
+  write_file "features/step_defnitions/#{name}_steps.rb",
+  <<-STEPS
+  Given(/^it fails once, then passes$/) do 
+    $#{name.downcase} ||= 0
+    $#{name.downcase} += 1
+    expect($#{name.downcase}).to eql 2
+  end
+  STEPS
+end
+
+Given /^a scenario "([^\"]*)" that fails twice, then passes$/ do |name|
+  write_file "features/#{name}.feature",
+  <<-FEATURE
+  Feature: #{name}
+    Scenario: #{name}
+      Given it fails twice, then passes
+  FEATURE
+
+  write_file "features/step_definitions/#{name}_steps.rb",
+  <<-STEPS 
+  Given(/^it fails twice, then passes$/) do 
+    $#{name.downcase} ||= 0
+    $#{name.downcase} += 1
+    expect($#{name.downcase}).to eql 3
+  end
+  STEPS
+end
+
+Given /^a scenario "([^\"]*)" that passes$/ do |name|
+  write_file "features/#{name}.feature",
+  <<-FEATURE
+  Feature: #{name}
+    Scenario: #{name}
+      Given it passes
+  FEATURE
+
+  write_file "features/step_definitions/#{name}_steps.rb",
+  <<-STEPS
+  Given(/^it passes$/) { expect(true).to be true }
+  STEPS
+end
+
+Given /^a scenario "([^\"]*)" that fails$/ do |name|
+  write_file "features/#{name}.feature",
+  <<-FEATURE
+  Feature: #{name}
+    Scenario: #{name}
+      Given it fails
+  FEATURE
+
+  write_file "features/step_definitions/#{name}_steps.rb",
+  <<-STEPS
+  Given(/^it fails$/) { expect(false).to be true }
+  STEPS
+end
+
 When /^I run the feature with the (\w+) formatter$/ do |formatter|
   expect(features.length).to eq 1
   run_feature features.first, formatter
--- a/features/lib/step_definitions/cucumber_steps.rb
+++ b/features/lib/step_definitions/cucumber_steps.rb
@@ -32,42 +32,6 @@ Given /^a step definition that looks like this:$/ do |string|
   create_step_definition { string }
 end
 
-Given /^a scenario "([^\"]*)" that fails once, then passes$/ do |name|
-  write_file "features/#{name}.feature",
-  <<-FEATURE 
-  Feature: #{name}
-    Scenario: #{name}
-      Given it fails once, then passes
-  FEATURE
-
-  write_file "features/step_defnitions/#{name}_steps.rb",
-  <<-STEPS
-  Given(/^it fails once, then passes$/) do 
-    $#{name.downcase} ||= 0
-    $#{name.downcase} += 1
-    expect($#{name.downcase}).to eql 2
-  end
-  STEPS
-end
-
-Given /^a scenario "([^\"]*)" that fails twice, then passes$/ do |name|
-  write_file "features/#{name}.feature",
-  <<-FEATURE
-  Feature: #{name}
-    Scenario: #{name}
-      Given it fails twice, then passes
-  FEATURE
-
-  write_file "features/step_definitions/#{name}_steps.rb",
-  <<-STEPS 
-  Given(/^it fails twice, then passes$/) do 
-    $#{name.downcase} ||= 0
-    $#{name.downcase} += 1
-    expect($#{name.downcase}).to eql 3
-  end
-  STEPS
-end
-
 Given /^a scenario "([^\"]*)" that passes$/ do |name|
   write_file "features/#{name}.feature",
   <<-FEATURE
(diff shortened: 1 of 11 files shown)


mattwynne:
Looks like a good start but there's no implementation right? So the tests should not be passing... or did I misunderstand something?

mattwynne:
You could usefully rebase your branch (you can use `git rebase -i`) then `push --force` to clean up the commit history on this PR - it makes it easier to read.

danascheider:
I think I was the one who misunderstood something. A while ago you said something about failing features being the first step, and I assumed that meant you wanted a PR with the tests. I’m giving a 45-minute talk tonight so things have been kind of crazy, but I’ll have a chance to work on this more later this week. And re: `git rebase -i` - I meant to do that before, sorry!

mattwynne:
Comprende! I was confused because the branch was green, but that's because you've changed the travis config is it?

So yeah I think those scenarios are a good place to start. We need to update the description a bit to align with what we've decided (i.e. deferring things like using tags).

mattwynne:
You happy to have a crack at an implementation Dana? Give me a shout if you need some support. Thanks for taking interest in this!

brasmusson:
@mattwynne `retry_failing_tests.feature` is tagged `@wip`, therefore build is green.

danascheider:
I'm totally working on it already, just didn't want to pollute the existing PR with all the new commits so I switched to a different branch & will merge later.

danascheider:
Happy to help!

mattwynne:
> I'm totally working on it already, just didn't want to pollute the existing PR with all the new commits so I switched to a different branch & will merge later.

OK. I've put the WIP tag on this PR, so it would work well for me if you just use this branch as you work on the solution, `rebase` and `push --force` as much as you like and just keep sharing the latest with us. That way I can stay involved and help. That's if you want any help of course!

danascheider:
That works for me. I should have some time to work on this tomorrow as I just finished a couple other projects that were taking up my time.

danascheider:
OK, I didn't end up getting to this today because my server got rooted and I had to fix everything. Keeping my fingers crossed for tomorrow!

mattwynne:
Friday is OSS day so that's good. Grab me in https://gitter.im/cucumber/cucumber-ruby if you want to chat.

lib/cucumber/filters/retry.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'cucumber/events/after_test_case'
+
+module Cucumber
+  module Filters
+    class Retry < Core::Filter.new(:configuration)
+
+      def test_case(test_case)
+        super 
+
+        configuration.on_event(:after_test_case) do |event|
```
mattwynne: I'd have thought you'd need to register the event listener earlier than this, like maybe in the filter's constructor. Does it work?

lib/cucumber/filters/retry.rb
```diff
@@ -0,0 +1,19 @@
...
+          test_case.describe_to(receiver) if event.result.failed?
```
mattwynne: This event listener, once registered, will keep firing even after this test case has been and gone. I think you need to register a listener earlier, and use it collect yourself a Hash of test case results in an instance variable, which you can then check here where you want to retry.

Does that make any sense?

danascheider: Off hand I'm not sure I understand, but I'll take a closer look when I get home. Thanks!

danascheider:
@mattwynne I got a little stuck trying to figure out why the one spec on this still fails. It seems to be saying that when there's a failing test (not a flaky one), it's only retrying once, even though the configuration options say it should be retried twice. Am I missing something obvious?

mattwynne:
I have a flight on Monday when I should be able to give this some focus. Can you push up whatever you have (use another branch if you want) and I'll get my head around it then?

Thanks for your patience @danascheider.

danascheider:
@mattwynne No problem, I'm pretty sure what's on this branch is all I have at this point, but I'll check my home laptop for more and push anything I find. I've been really busy too, so no worries.

mattwynne:
@danascheider FINALLY I got a chance to look at this 😀. I'm lucky enough to have @brasmusson in the same room as me today, and we figured out some subtle problems that were preventing the retry filter from working quite right. We pushed them here: 8f9445f9cd8d5b9e9e49bea639eebafff2235256 for you to take a look at. I wouldn't copy the changes line-for-line as we were messing around quite a bit and probably changed more than is necessary. I'll put some comments on that diff to show you the interesting bits.

FYI, the acceptance test now seems to be failing for another reason - even when a test has been retried, the summary still shows `1 scenario`. @brasmusson and I are looking into that now.

danascheider:
@mattwynne Thank you! I'll take a look at it too.

mattwynne:
Hi @danascheider, just checking in to see how you're feeling about this one. I'm deliberately leaving it so you can get the sense of satisfaction from finishing it. Is there anything blocking you right now that I can help with?

danascheider:
Oh, thanks, @mattwynne. It's on my list, I'm just working a lot and in the middle of moving so I had it on the back burner.

mattwynne:
👍🚚💼

Take your time @danascheider, I think you have most of the puzzle pieces now. The existing code for the summary numbers in the formatters is a bit of a muddle but that's a problem we can deal with when you've got the main behaviour working.

danascheider:
Got all the tests passing - any other thoughts, @mattwynne or @brasmusson? What was the problem with the formatter for the summary numbers?

brasmusson:
An reasonable output from the progress formatter in case of one scenario that passed, one scenario that passed after one retry, one scenario that passed after two retries, and one scenario that still failed after two retries could be:
```
..R.RR.RRF
<exception stacktraces>
4 Scenarios (3 Passed, 1 Failed) - 2 Scenarios needed retires to pass
```
A problem with the implementation is that the summary is provided by the [Console module](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/formatter/console.rb#L28), it is used both by the pretty formatter and the progress formatter (and peoples custom formatter), and the progress formatter is on the way to be fully based on events (#977) using the [Summary class](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/report/summary.rb) from the Core , whereas the pretty formatter is still using the legacy formatter API. To implement what is needed to produce a summary as the one above which works for all the APIs that formatters uses seems to be - how should I put it - not trivial.

mattwynne:
> To implement what is needed to produce a summary as the one above which works for all the APIs that formatters uses seems to be - how should I put it - not trivial.

Classic Swedish understatement @brasmusson!

Actually I don't think it will be that hard, but equally I think it's an enhancement. With what @danascheider's already done, the thing is usable and useful. We can improve on the reporting in another iteration IMO.

Thoughts?

features/lib/step_definitions/cucumber_steps.rb
```diff
@@ -32,6 +32,70 @@
   create_step_definition { string }
 end
 
+Given /^a scenario "([^\"]*)" that fails once, then passes$/ do |name|
```
mattwynne: Do you think maybe these should go into a `retry_steps.rb`? They're quite specific...

spec/cucumber/filters/retry_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      filter.test_case(test_case)
+    end
+  end
+
+  context "flaky test cases" do
+
+    context "a little flaky" do
+      it "describes the test case twice" do
+        results = [fail, pass]
+        expect(receiver).to receive(:test_case) do |test_case|
```
mattwynne: Minor style nit-pick (we really should try and get Rubocop going) but I like to see `{ }` used whenever a block returns a value that's then used. In memory of [Jim](https://github.com/bbatsov/ruby-style-guide/issues/162#issue-13134473).

mattwynne:
There's lot of commits here. Probably best to squash when this gets merged? Unless you want to rebase @danascheider?

features/lib/step_definitions/cucumber_steps.rb
danascheider: Done.

spec/cucumber/filters/retry_spec.rb
danascheider: Also done!

danascheider:
Sure, they can be squashed.

lib/cucumber/filters/retry.rb
```diff
@@ -0,0 +1,32 @@
...
+          event.test_case.describe_to(receiver)
+        end
+
+        super
+      end
+
+      private
+
+      def retry_required?(test_case, event)
+        event.test_case == test_case && event.result.failed? && test_case_counts[test_case] < configuration.retry_attempts
```
brasmusson: This does not work, since the test_case you pass on is never the same that are in the `after_test_case` event. Filters like [activate_steps](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/filters/activate_steps.rb#L22) call [Case#with_steps](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/case.rb#L41) which creates a new test case object.

mattwynne: I've resolved this with https://github.com/cucumber/cucumber-ruby-core/pull/111 which makes equality for `Test::Case` objects work as you would expect it to.

PR #922, opened 2015-10-12
Fix Configuration.with_options merge direction

limhoff-r7:
Without this change, `factory.new(@configuration.with_options(out_stream: out_stream))` in Cucumber::Runtime#create_formatter doesn't actually changed the `out_stream` from STDOUT in the default configuration because the merge order was `new_options.merge(@options)`, so any keys in both `new_options` and `@options`, like `:out_stream`, would have the value from `@options` win.

The specs weren't catching this problem because `with_options` was being called on `Configuration.new`, which had no keys set.  I changed in spec in 63d4c37 to reveal the bug and fixed it in 067215c.

This bug may have manifested by users unable to override the --out from STDOUT when using new formatters that have `initialize(configuration)` instead of the legacy `initialize/3`

### DIFF
--- a/lib/cucumber/configuration.rb
+++ b/lib/cucumber/configuration.rb
@@ -32,7 +32,7 @@ module Cucumber
     end
 
     def with_options(new_options)
-      self.class.new(new_options.merge(@options))
+      self.class.new(@options.merge(new_options))
     end
 
     # TODO: Actually Deprecate???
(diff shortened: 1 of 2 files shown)


mattwynne:
SUPERSLEUTH! :mag: 

Thanks Luke, great work.

Issue #927, opened 2015-10-27
More issues with rerun file...


mcrmfc:
Hi,

So in https://github.com/cucumber/cucumber-ruby/issues/865 we made our treatment of rerun more tolerant to cope with some different formats thrown up by paralllel_cucumber.

Now we have found another (https://github.com/grosser/parallel_tests/issues/454)  - scenarios with no space in between. I think although we are now possibly on the verge of going too far probably we should tolerate this given there is no 'agreed' format for re-run files.

I have a PR prepared (tests updated) so see what you think.

PR #928, opened 2015-10-27
Issue #927 - handle even more exotic rerun file formats seen in the wild

mcrmfc:
Relates to #927 - handle more weird rerun formats.

### DIFF
--- a/spec/cucumber/cli/rerun_spec.rb
+++ b/spec/cucumber/cli/rerun_spec.rb
@@ -71,16 +71,17 @@ module Cucumber
         end
       end
 
-      context "contains features and scenarios with and without spaces same line and across lines" do
+      context "contains features and scenarios with and without spaces on same line and across lines" do
         before(:each) do
           allow(IO).to receive(:read).and_return(
-            "cucumber.feature:1:3\ncucumber.feature:5 cucumber.feature:10\n"\
+            "cucumber.feature:1:3\ncucumber.feature:5baz.feature:4qux test.feature cucumber.feature:10\n"\
             "cucumber space.feature:134 domain folder/cuke.feature\n")
         end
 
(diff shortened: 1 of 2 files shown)


Shashikant86:
@mcrmfc @mattwynne This fixed re-run issue for the parallel_cucumber. I have checked this by updating my test repo with Matt's fork. https://github.com/Shashikant86/parallel-cucumber-failed-rerun
It works great. Hopefully we can merge this provided TravisCI build passes.

spec/cucumber/cli/rerun_spec.rb
```diff
@@ -71,16 +71,17 @@ module Cli
         end
       end
 
-      context "contains features and scenarios with and without spaces same line and across lines" do
+      context "contains features and scenarios with and without spaces on same line and across lines" do
```
mattwynne: there are a lot of rules in this example! I think it would be easier to read if you split it into several examples.

mattwynne:
Also, that is a very exotic format indeed! What is producing that output?!

spec/cucumber/cli/rerun_spec.rb
Shashikant86: @mattwynne Parallel_test's Cucumber logger now creates this output without any space. It used to work well but this commit broken it https://github.com/grosser/parallel_tests/pull/416  
related issue can be found here https://github.com/grosser/parallel_tests/pull/398

mcrmfc: @mattwynne - yep fair point, will break them out a bit - probably this should also be fixed in parallel_tests but since we don't constrain users via a known interchange format (json/yaml whatever) I guess there is no harm in defending against 🎃 👻 formats..  I'll break up the specs a bit.

mcrmfc:
@mattwynne I have simplified the specs a little now.

Issue #450, opened 2013-05-09
Add step to AfterStep


drekka:
We are using frank-cucumber to run tests against an iOS app. As this app is heavily animated we need to pause after each step to ensure the animations are finished. However there are a couple of steps where we do not want to pause. We have tried the AfterStep hook however we have a significant issue with it. 

    AfterStep do |scenario|
       ...
    end

The issue is that it does not appear to be passed the text of the step. This makes it impossible to tell what step was being executed. It seems strange to us that an AfterStep hook would be created without being passed the step that was executed. Could we please have a version like this:

    AfterStep do |scenario, step|
       ...
    end

mattwynne:
Sadly, @chewi's hack will not work on 2.0 as things stand. We will aim to support this properly through the API during the 2.0 series. I'll keep the ticket open.

It would be awesome if someone wanted to start a PR with a failing test for this.

Carlbc18:
I am also interested in this functionality.

djwgit:
like to have this back too,   as in 1.3.x
in 2.x, scenario reference in the hook is a different one...  hope you guys could fix it soon, thanks
and it does make sense to be able to access the steps within the hook...

mattwynne:
If you're interested in this, please start with a failing scenario that describes what you want.

jdougit:
@mattwynne, thanks for your reply...

basically, I run cucumber test on jenkins, by default, the output flushed out to jenkins console after a scenario is finished. but I'd like to have the output send to jenkins console immediately, 
- before a step executed, print out the step to console.
- after that, there are more console output during the test code of the step,  
that makes the console output more readable and easier to located an issue if fails)
to have it this way, I print out steps in the Before and AfterStep hooks...  (might not be a common usage for others...)

after reading the cucuber-ruby code, seems, 
- in 1.3.8   the scenario instance in the hook is the "real" scenario class instance, it has raw_steps attributes. 
- but after 2.x,  the scenario passed into hook is not that one anymore...  instead, it is
Cucumber::RunningTestCase::Scenario, and from this, not sure how to access its steps...

I have this below in my env.rb
================================
# env.rb
# don't buffer the console output. get real-time update on jenkins console page
$stdout.sync = true

# setup and tear-down, run before and after each Scenario
Before do |scenario|
  print("Scenario: #{scenario.title}\n")
  $stop_count = scenario.raw_steps.count
  $step_index = 0
  # print out the first step if exists
  if $step_index < $stop_count
[...]

PR #931, opened 2015-11-10
Adding Test Step in AfterStep feature file.

t-morgan:
Working on Add step to AfterStep #450 

Do you think the argument list should be consistent with how formatters work or will it potentially cause problems moving the current argument out of its position as first argument?

### DIFF
--- a/features/docs/test_step_available_in_after_step_hook.feature
+++ b/features/docs/test_step_available_in_after_step_hook.feature
@@ -3,20 +3,42 @@ Feature: Test Step is available in AfterStep Block
   As a developer
   I want Test Step objects available in AfterStep blocks
 
-  Scenario: Access Test Step object in AfterStep Block
+  Background:
     Given the standard step definitions
-    And a file named "features/support/env.rb" with:
-      """
-      AfterStep do |test_step, result|
-        expect(test_step).to be_a(Cucumber::Core::Test::Step)
-      end
-      """
     And a file named "features/sample.feature" with:
       """
       Feature: Sample
 
         Scenario: Success
           Given this step passes
+      """
+
+  Scenario: Access Test Step object in AfterStep Block
+    Given a file named "features/support/env.rb" with:
+      """
+      AfterStep do |result, test_step|
+        expect(test_step).to be_a(Cucumber::Core::Test::Step)
+      end
+      """
+    When I run `cucumber features`
+    Then it should pass with:
+      """
+      Feature: Sample
+
+        Scenario: Success        # features/sample.feature:3
+          Given this step passes # features/step_definitions/steps.rb:1
+
+      1 scenario (1 passed)
+      1 step (1 passed)
+
+      """
+
+  Scenario: An AfterStep with one named argument does not change behavior
+    Given a file named "features/support/env.rb" with:
+      """
+      AfterStep do |result|
+        expect(result).to be_a(Cucumber::Core::Test::Result::Passed)
+      end
       """
     When I run `cucumber features`
     Then it should pass with:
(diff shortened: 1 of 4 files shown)


brasmusson:
Isn't not having `scenario` as first argument to the AfterStep hooks, such a breaking change that only can be introduced in a major version upgrade?

If the intention is to add both `step` and `result` as arguments to AfterStep hooks, I see it as two separate changes, from:
```
Before do | scenario |
After do | scenario |
AfterStep do | scenario |
```
to:
```
Before do | scenario |
After do | scenario |
AfterStep do | scenario, step |
```
And another change to.
```
Before do | scenario |
After do | scenario, scenario_result |
AfterStep do | scenario, step, step_result |
```

t-morgan:
I could be wrong, but it seems to me that since upgrading to Cucumber 2 the only argument I an receiving from the AfterStep hook is a Cucumber::Core::Test::Result of whatever status the step invocation was.

brasmusson:
Hm, based on #450 I thought that the AfterStep hooks in Cucumber v2 received a Cucumber::RunningTestCase, as the Before and After hooks do.

t-morgan:
It was definitely a scenario object in Cucumber 1.x when the issue was first reported (and it was pretty simple to get the step through that scenario at the time).

In Cucumber 2.x, I am only finding a result in the AfterStep block (Before and After still receive the scenario as normal). 

I know I can add the test_step when the hook action is created in the after_step_hooks method of Cucumber::Runtime::StepHooks in lib/cucumber/runtime/step_hooks.rb and my test will pass... I'm just not sure if that is the right/best place to do it.

t-morgan:
I went ahead and implemented my change to fix the AfterStep hook functionality. If there are any changes that should be made, please let me know.

I had to flatten an array in lib/cucumber/rb_support/rb_hook.rb because the argument passed into the AfterStep hook was being wrapped in an Array for some reason.

mattwynne:
Thanks sharing your fix @t-morgan.

mattwynne:
Could you rename the feature (and reword the description) to just be a general feature for AfterStep hooks? It should live here: https://github.com/cucumber/cucumber-ruby/tree/master/features/docs/writing_support_code

t-morgan:
Hey @mattwynne, I cleaned up the feature file and moved it as you requested. I think it might be a good idea to also include another scenario showing AfterSteps do run for background steps (apparently this was not always the case).

maxmeyer:
@t-morgan What would be a "background step"? Some command running in the background?

t-morgan:
@maxmeyer, no it is just a step defined under 'Background:' for the feature.

maxmeyer:
Ah. Ok. :-)

gergelyborsothy:
Are you planning to merge this to master? We're planning to measure the time it takes a step to complete, and this modification would be extremely useful.

chewi:
Please merge it, AfterStep is quite useless in its current form.

chewi:
Is there a nicer way to get the index of the step other than this?

```ruby
step.source[-2].children.index(step.source.last)
```

mattwynne:
Thanks @t-morgan!

mattwynne:
As part of our new policy I've invited you to become a committer to this repo @t-morgan. I liked how you found your way around the code and tested your change. I hope we'll get more contributions from you in future!

djwgit:
@mattwynne @t-morgan 
with cucumber-ruby gem 2.3.3,  I got nil class for the test_step... is this expected? according to the after_step_hooks.feature, it should be the step, right?  not sure if I missed something... thanks

```
AfterStep do |result, test_step|
  print result.class
  print "\n-------\n"
  print test_step.class
end
```
I got this
```
Cucumber::Core::Test::Result::Passed
-------
NilClass
```

PR #932, opened 2015-11-12
Refactor event name resolving

dg-ratiodata:
@mattwynne As discussed some weeks ago I finished the refactoring of the `Event::Bus`:

* Added more tests for Bus + NameResolver
* Refactored the the test suite to make use of `context` to show the code behaves in different contexts/situations

Maybe we should really move that event bus thing to a separate library. This would make sharing the code much easier and we could use rubygems + version requirement(s) to make aruba and cucumber depend on the correct interface. WDT? What about `cucumber-ruby-common` or just `event-bus`.

### DIFF
--- a/lib/cucumber/events/bus.rb
+++ b/lib/cucumber/events/bus.rb
@@ -1,6 +1,8 @@
+require 'cucumber/events/name_resolver'
+require 'cucumber/errors'
+
 module Cucumber
   module Events
-
     # Event bus
     #
     # Implements and in-process pub-sub events broadcaster allowing multiple observers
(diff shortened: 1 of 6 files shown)


dg-ratiodata:
@mattwynne I prepared a ruby gem for this: https://github.com/dg-ratiodata/event-bus

dg-ratiodata:
I prepared a PR to integrate the gem instead of the changes made in this PR. Either #932 or #933 should be merged.

mattwynne:
Thanks @dg-ratiodata. I'm pretty snowed under for the next few weeks so I may not have a chance to grok this code for a while. Don't let that hold you up using the event bus pattern for aruba.

maxmeyer:
@mattwynne Ok. I am going to use the `gem` for `aruba`.

dg-ratiodata:
Pinging @tooky on this as well.

lib/cucumber/events/bus.rb
```diff
@@ -8,79 +10,31 @@ module Events
...
-      end
+        raise ArgumentError, "Please pass either an object#call or a handler block" if handler.nil? || !handler.respond_to?(:call)
 
-      private
+        event_class = @resolver.transform(event_id)
 
-      def handlers_for(event_class)
-        @handlers[event_class.to_s] ||= []
-      end
+        fail EventNameResolveError, %(Transforming "#{event_id}" into an event name failed for unknown reason.) if event_class.nil?
```
mattwynne: I don't see why we need this line - it seems overly defensive. The `#transform` method always returns an event type, or raises an exception.

lib/cucumber/events/bus.rb
```diff
@@ -8,79 +10,31 @@ module Events
...
-            # Go down the ancestors to check if it is owned directly. The check
-            # stops when we reach Object or the end of ancestors tree.
-            constant = constant.ancestors.inject do |const, ancestor|
-              break const    if ancestor == Object
-              break ancestor if ancestor.const_defined?(name, false)
-              const
-            end
+      # Broadcast an event
+      def notify(event)
+        fail NoEventError, 'Please pass an event object, not a class' if event.is_a?(Class)
```
mattwynne: I can see why we need this check. Garbage in, garbage out. Don't you think so? Or was this a mistake you made when you first used the API?

spec/cucumber/events/bus_spec.rb
```diff
@@ -1,94 +1,185 @@
...
+  let(:name_resolver) { instance_double('Cucumber::Events::NameResolver') }
+
+  let!(:test_event_klass) do
+    class_double('Events::TestEvent').tap do |double|
+      stub_const('Events::TestEvent', double)
+      allow(double).to receive(:new).and_return(test_event_instance)
+      allow(double).to receive(:to_s).and_return('Events::TestEvent')
+      allow(double).to receive(:is_a?).with(Class).and_return(true)
     end
+  end
```
mattwynne: Sorry, but why not just write the classes inline as in the original test? I find this harder to read and unnecessary.

mattwynne:
Having read this, I think that my preferred refactoring (if we're going to bother extracting a resolver at all) would be to extract three strategy objects so that we can do away with the whole case statement.

mattwynne:
Superseded by #933

lib/cucumber/events/bus.rb
dg-ratiodata: I think I did it wrong myself and added this check afterwards.

spec/cucumber/events/bus_spec.rb
dg-ratiodata: Mmm... Yes, but this doesn't pollute the namespace and is easier to extend. I try to use this kind of pattern in a few of my libs....

lib/cucumber/events/bus.rb
dg-ratiodata: No. Since we support user defined NameResolvers, this is a countermeasure against "NameResolvers".

mattwynne: We should assume that user-defined NameResolvers will follow the protocol and never return nil. Otherwise we have to clutter up our code.

spec/cucumber/events/bus_spec.rb
mattwynne: I'd at least use an inline call to `Class.new` rather than all this stubbing stuff. But honestly for me the trade-off with polluting the namespace is not worth it.

dg-ratiodata:
Let's move the discussion here https://github.com/fedux-org/event-bus/pull/4.

PR #933, opened 2015-11-13
Outsource event bus

dg-ratiodata:
@mattwynne This PR outsources the whole event bus thing in a separate library but brings the same functionality like #932: Either #932 or this PR should be merged

If this is merged: @aslakhellesoy @mattwynne can you send me your email-addresses so I can make you owner of the event-bus-gem too? I would also prefer to move the library's repository to the cucumber organization to give some more people admin access to it. WDYT


mattwynne:
Thanks again @dg-ratiodata I will review soonish.

I agree in principle that the event-bus is something that we could factor out, I'm just a little bit dubious about whether it's worth it for 100 LOC.

@tooky and I paired on this code so he should be involved as a maintainer of any future gem too.

dg-ratiodata:
> I agree in principle that the event-bus is something that we could factor out, I'm just a little bit dubious about whether it's worth it for 100 LOC.

I agree, but:

* No C&P of the same code between aruba and cucumber
* Better documentation, btw. Aruba + Cucumber -> Living Documentation
* Better tests
* Versioninig of API

dg-ratiodata:
> @tooky and I paired on this code so he should be involved as a maintainer of any future gem too.

Yeah. No problem. :-) @tooky Can you send me your e-mail address as well. If you like via PM on [Twitter](https://twitter.com/nerdyptah/).

dg-ratiodata:
@mattwynne The code in `event-bus` has now a slightly different, but backward compatible API. You can now register multiple events (`Array`) at once for the same handler. That's something I needed for aruba to DRY up code.

dg-ratiodata:
I forgot to add this one: https://github.com/dg-ratiodata/event-bus.

dg-ratiodata:
`aruba` now uses `event-bus` I finally merged the code into `master`. :-) cucumber/aruba#309

PR #939, opened 2015-12-17
Fix extra arg passed to ensure_io.

benshan:
This was possibly missed from c1768eb91f9399c6533cf2ff22608886514ee5f8.

### DIFF
--- a/features/docs/formatters/custom_formatter.feature
+++ b/features/docs/formatters/custom_formatter.feature
@@ -1,0 +1,25 @@
+Feature: Custom formatter
+  Add test coverage for utility functions which are not used by other flows.
+
+  Scenario: test Cucumber::Formatter::Io::ensure_file
+    Given a file named "features/f.feature" with:
+      """
+      Feature: Test
+        Scenario: Test
+      """
+    And a file named "features/support/custom_formatter.rb" with:
+      """
+      require 'cucumber/formatter/io'
+      module Cucumber
+        module Formatter
+          class Test
+            include Io
+            def initialize(runtime, path_or_io, options)
+              ensure_file("features/f.feature", "features file")
+            end
+          end
+        end
+      end
+      """
+    When I run `cucumber features/f.feature --format Cucumber::Formatter::Test`
+    Then it should pass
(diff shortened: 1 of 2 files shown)


mattwynne:
Thanks Ben, but I worry this will just slip loose again without a test. How
did you find out it was needed?

benshan:
@mattwynne I am updating our project from calabash 0.16 to 0.17 which has brought many dependency versions up. We have a formatter which was calling ensure_io with the previous signature which included the name. I saw the errors while running. Looking at the source, I was confused why the line was left in io.rb with the old invocation. Looking at the history, it seemed to be an oversight. Yes, I assumed as much that there is no test for ensure_file which allowed the change to go through. If no one has reported yet, there might not be too many consumers of that method :)

benshan:
@mattwynne I'm happy to code a test, but it wasn't obvious to me where to put this. ensure_dir is tested by virtue of junit.rb being used. As far as I can see, I would add a new step for something like "verify the file ... exists", then put a scenario in place with "given a file ..." and then "verify the file ... exists". Sound right? If so, any guidance on which files to house these or create new files? (I started creating new files and it didn't feel right, so I wanted to ask.)

mattwynne:
From what you've told me, it sounds like we need a new scenario in https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/extending_cucumber/custom_formatter.feature that pins down what your code expects to be able to do with `ensure_io`. Am I right?

benshan:
That sounds most reasonable since this helper is apparently not used by the internals and is a helper for something like a custom formatter. And then we just need a step to exercise the actual ruby function. Is there a good existing file for that step under step_definitions or would I create a new file custom_formatter.steps?

mattwynne:
Have a look at the other scenarios in that feature. You should be able to use the existing steps to write the ruby files for a simple custom formatter that uses the method.

benshan:
Updated with a first attempt. Based on your guidance and what I saw in formatter_step_file_colon_line.feature, this seemed like the simplest and most compact way to get the coverage. Io.rb is now at 100% coverage when I run locally, and I verified that the original issue would have been caught by this test when the args were refactored.

PR #946, opened 2016-02-01
"Fix undefined method `support_code'" in Transform method

tkm-kj:
Cucumber 1.3.20, the following code.

```ruby 
# feature file

Feature: Test feature
  Scenario: Test scenario
    Given The folowing users exist
      | name | email         |
      | foo     | foo@foo.com |
      | bbb    | bbb@bbb.com |

# step file

Given /^The folowing users exist$/ do |table|
  table.rows.each do |name, email|
    Transform(name)
  end
end

Transform /foo/ do |ignore|
  'Foo'
end
```

The result is this.

```
1 scenario (1 passed)
1 step (1 passed)
0m0.006s
```

But, cucumber 2.3.2, the result is...

```
Feature: Test feature

  Scenario: Test scenario          # features/test.feature:2
    Given The folowing users exist # features/step_definitions/test_step.rb:1
      | name | email                 #|
      | foo     | foo@foo.com |
      | bbb    | bbb@bbb.com |
      undefined method `support_code' for #<Cucumber::Runtime::ForProgrammingLanguages:0x007fa082ae8880> (NoMethodError)
      ./features/step_definitions/test_step.rb:3:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/test_step.rb:2:in `each'
      ./features/step_definitions/test_step.rb:2:in `/^The folowing users exist$/'
      features/test.feature:3:in `Given The folowing users exist'

Failing Scenarios:
cucumber features/test.feature:2 # Scenario: Test scenario

1 scenario (1 failed)
1 step (1 failed)
0m0.012s
```

This problem is due to the fact that can't be refered to `@support_code` of `Cucumber::Runtime::ForProgrammingLanguages` when call `Transform` method.
So, I fixed it.

### DIFF
--- a/lib/cucumber/runtime/for_programming_languages.rb
+++ b/lib/cucumber/runtime/for_programming_languages.rb
@@ -11,6 +11,8 @@ module Cucumber
     class ForProgrammingLanguages
       extend Forwardable
 
+      attr_reader :support_code
+
       def initialize(support_code, user_interface)
         @support_code, @user_interface = support_code, user_interface
       end


mattwynne:
Thanks for the fix @tkm-kj! I added a scenario too, so it can't regress.

PR #950, opened 2016-02-15
Fix images embedding in html formatter for scenario outlines

sofaking:
Previously embedded images were shown above examples table which
is not very convenient.

Now each embed will be placed right after the example it's attached to.

I'm not sure about Then steps, they're really ugly. But I wasn't able to come up with better
idea, so any suggestions are welcome.

And yes I deliberately skipped text embeds, because wanted to do PR ASAP to gather some feedback.

### DIFF
--- a/features/docs/formatters/html_formatter.feature
+++ b/features/docs/formatters/html_formatter.feature
@@ -89,3 +89,78 @@ Feature: HTML output formatter
     makeRed('background_0')
     """
 
+  Scenario: embedding a screenshot
+    Given a file named "features/embed.feature" with:
+      """
+      Feature: Embed
+
+        Scenario: a screenshot
+          Given a step that is embedding a screenshot
+      """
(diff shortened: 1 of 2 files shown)


mattwynne:
Hi @sofaking sorry for the slow feedback.

Thanks for thinking about how to test your change. Instead of writing scenarios, try looking at https://github.com/cucumber/cucumber-ruby/blob/master/spec/cucumber/formatter/html_spec.rb as the place to drive your changes from.

Make sense?

sofaking:
Yes, totally make sense. I was surprised not to find extensive test suite
for html formatter, since the features are quite shallow. Now you opened my
eyes.

mattwynne:
@sofaking do you have any intention to move those tests into unit tests?

sofaking:
Oh, to be honest completely forgot about this. Will try to do that in following week or so.

sofaking:
Create a new one #1061, RSpec driven, as suggested by Matt

PR #954, opened 2016-03-20
WIP: Custom formatter config

mattwynne:
Currently, formatter instances fall into three categories:

1) the one that's writing to STDOUT
2) opening a file and writing to that
3) other arbitrary config (e.g. JUnit formatter)

We have an `io` argument that's passed in the constructor to v1.0-style formatters that could be one of the following:
- the STDOUT stream to write to
- the path to a file to open
- the path to the directory that the JUnit formatter is going to write files into

This is messy and confusing. This PR aims to introduce a new, per-formatter configuration structure and un-pick the old `--out` option.

### DIFF
--- a/lib/cucumber/cli/main.rb
+++ b/lib/cucumber/cli/main.rb
@@ -15,8 +15,8 @@ module Cucumber
       def initialize(args, _=nil, out=STDOUT, err=STDERR, kernel=Kernel)
         @args   = args
         @kernel = kernel
-        $stdout = @out = out
-        $stderr = @err = err
+        @out = out
+        @err = err
       end
 
       def execute!(existing_runtime = nil)
--- a/lib/cucumber/cli/main.rb
+++ b/lib/cucumber/cli/main.rb
@@ -14,9 +14,9 @@ module Cucumber
 
       def initialize(args, _=nil, out=STDOUT, err=STDERR, kernel=Kernel)
         @args   = args
-        @out    = out
-        @err    = err
         @kernel = kernel
+        $stdout = @out = out
+        $stderr = @err = err
       end
 
       def execute!(existing_runtime = nil)
--- a/features/docs/extending_cucumber/custom_formatter.feature
+++ b/features/docs/extending_cucumber/custom_formatter.feature
@@ -60,6 +60,7 @@ Feature: Custom Formatter
       {"foo"=>"bar", "one"=>"two"}
       """
 
+  @spawn
   Scenario: Support legacy --out
     Given a file named "features/support/custom_formatter.rb" with:
       """
@@ -67,7 +68,7 @@ Feature: Custom Formatter
         class Formatter
           def initialize(config, options={})
             config.on_event Cucumber::Events::FinishedTesting do |event|
-              puts options["out"]
+              $stdout.puts options["out"]
             end
           end
         end
@@ -76,13 +77,14 @@ Feature: Custom Formatter
     When I run `cucumber features/f.feature --format MyCustom::Formatter --out foo.file`
     Then it should pass with exactly:
       """
-      foo.file
       Deprecated: Please don't use --out, but pass the formatter options like this instead:
 
         --format junit,out=path/to/output
+      foo.file
 
       """
 
+  @spawn
   Scenario: Setting output using the new style per-formatter options
     Given a file named "features/support/custom_formatter.rb" with:
       """
@@ -90,7 +92,7 @@ Feature: Custom Formatter
         class Formatter
           def initialize(config, options={})
             config.on_event Cucumber::Events::FinishedTesting do |event|
-              puts options["out"]
+              $stdout.puts options["out"]
             end
           end
         end
--- a/features/docs/extending_cucumber/custom_formatter.feature
+++ b/features/docs/extending_cucumber/custom_formatter.feature
@@ -41,6 +41,48 @@ Feature: Custom Formatter
 
       """
 
+  Scenario: Custom config
+    Given a file named "features/support/custom_formatter.rb" with:
+      """
+      module MyCustom
+        class Formatter
+          def initialize(config, options={})
+            config.on_event Cucumber::Events::FinishedRunningTests do |event|
+              puts options.inspect
+            end
+          end
+        end
+      end
+      """
+    When I run `cucumber features/f.feature --format MyCustom::Formatter,foo=bar,one=two`
+    Then it should pass with exactly:
+      """
+      { "foo": "bar", "one": "two" }
+      """
+
+  Scenario: Support legacy --out
+    Given a file named "features/support/custom_formatter.rb" with:
+      """
+      module MyCustom
+        class Formatter
+          def initialize(config, options={})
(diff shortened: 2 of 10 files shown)


mattwynne:
Ping @maxmeyer - I'll coach you to fix this :+1:

maxmeyer:
@mattwynne I've got two ideas to capture the output of cucumber:

1. Spawn cucumber as normal command - this would not use the InProcess-feature of aruba and just write the output to STDOUT via puts
2. Write the output to a file and check it with `Then the file should contain:`

Or do you have got another idea? What would you prefer?

lib/cucumber/cli/main.rb
```diff
@@ -14,9 +14,9 @@ def execute(args)
 
       def initialize(args, _=nil, out=STDOUT, err=STDERR, kernel=Kernel)
         @args   = args
-        @out    = out
-        @err    = err
         @kernel = kernel
+        $stdout = @out = out
+        $stderr = @err = err
```
maxmeyer: those lines will be removed, when we agree about how to fix the tests.

dg-ratiodata:
@mattwynne I'm a bit out of time. Hope to find a time slot soon to work on this again.

dg-ratiodata:
@danascheider Dana, I saw you working on the CLI code of cucumber? Would you mind to take over this PR?
@mattwynne I would like to spend the bit of my spare time which is left on `aruba`

1. I saw a lot of changes happening to the CLI code
2. I got in involved in a big offline project of my hiking and climbing club
3. Unfortunately I don't have a enough time to catch up with the new code. Sorry!

danascheider:
I might be able to take this on. I just finished one PR but as I recall, I had committed to two, so I want to figure out what was on my docket next before I take anything else on.

dg-ratiodata:
Yeah, I totally understand that. Thanks anyway.

mattwynne:
@nodo this would also be a great one to finish off if you can wrap your head around it.

charlierudolph:
@mattwynne this just caught my eye and wanted to let you know how `cucumber-js` does things / is planning on doing things. 

In preparing cucumber-js 2.0, I'm planning to add in a `--format-option <JSON>` to specify options that get passed to all formatters. Went with JSON for ultimate flexibility and I will be suggesting people only use in profiles, where they can build the options as an object and convert it to JSON.

Very interesting idea of allowing users to pass options when specifying the formatter. I do want to be able to specify options for just one formatter but like having a way to specify options used for  all formatters. Mainly because the pretty / progress / summary all have some shared options.

The output path is engrained in `cucumber-js` when specifying the [format](https://github.com/cucumber/cucumber-js/blob/v1.3.0/docs/cli.md#formatters), though our syntax is also dependent on not having a `:` in the format name.

mattwynne:
Thanks @charlierudolph - it would be great to converge our UIs as much as possible, though JSON on the CLI seems a bit un-idiomatic to me... maybe I'm old-school? Anyway, the hard work here is cleaning up the constructor interface on the formatters, which is internal stuff. Once that's done we'll be free to try out different ways of passing in the options.

charlierudolph:
I do hate JSON on the CLI but if you only do it in the profile and are thus writing it only once, it wouldn't be bad. My main reason for doing it was we already have a json option for passing parameters to a custom world constructor (for people to use instead of env variables) and part of that request included sending an array parameter.

nodo:
@mattwynne , I will be happy to pick it up. I will start going through it in the next days.

nodo:
FYI: I have rebase this branch onto master and push my branch https://github.com/cucumber/cucumber-ruby/tree/nodo/clean-up-formatter-config . I didn't want to force push because the diff is quite big and I am not sure how many people are working on this branch.

mattwynne:
@nodo AFAIK only @dg-ratiodata was working on it and he's stopped so I think you'd be fine to force push. If you'd prefer a clean slate you could create a new branch / PR in this repo too.

maxmeyer:
I was the only one working on it. Go ahead! Cheers, Dennis

Am 4. Oktober 2016 19:15:41 MESZ, schrieb Matt Wynne <notifications@github.com>:

nodo:
@mattwynne I think I would be able to progress on that in during next days. However, I have a question. Should the "out_stream" be part of `config` or `options`.

I am looking at built-in formatters and I am quite confused. `json` formatter has `config.out_stream`. While `html` has:
```
      def initialize(runtime, path_or_io, options)
        @io = ensure_io(path_or_io)
         [...]
      end
```

Should we refactor this as well?

brasmusson:
@nodo The `pretty` and `html` formatters are the two formatters still not [re-written against the new API](https://github.com/cucumber/cucumber-ruby/issues/839), therefore they are instantiated with the `legacy_api` [here](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L199-L204), instead of [here](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L196-L197). It is possible that that the `pretty` and `html` formatters will not be re-written against the new API, before replaced by binary distributed formatters common to all Cucumber implementations, see https://github.com/cucumber/cucumber-pretty-formatter for that initiative.

My advice is to leave the `pretty` and `html` formatters out of the scope of this PR, and focus on the rest of the formatters that are using the new API and are not instantiated with the `legacy_api`.

nodo:
Thanks for the pointers @brasmusson ! Just to be sure, the goal of this PR is allow to pass options for CLI like:
```
--format MyFormatter,key=value
```
One of these options could be `out=path/to/file`.

The custom formatters will get 2 parameters:
- config (with out_stream method)
- options (which is the options passed to the formatter)

What do you think?

mattwynne:
@nodo this is complex and needs some careful thought so we can more likely get the API right this time.

There are a few rules / variables at play here.

1) at least one of the formatters will have control of STDOUT to write to the user's console
2) formatters that write to an output stream could either be writing the main console formatter or be connected to a file. They should probably not know or care which.
3) Not all formatters write to an output stream. e.g. the JUnit formatter, which writes various files into a folder. I can imagine others that write to databases or message queues.

At the moment, the "new" API Bjorn refers to is to pass a copy of the global `config` to each formatter's constructor, customised with the out_stream set to either STDOUT (if we've decided it's the main console formatter) or to the output file, if one was specified in the CLI options using the `--out` flag.

What I suggest is that you limit this PR to adding custom formatter options, and leave the `--out` flag behaviour as it is for now. We can then change the JUnit formatter to use it's own `--path` option, so that we can stop doing this ugly `path_or_io` thing.

It will be easier to think about the next step once we've done that. WDYT?

nodo:
Thanks for the explanation @mattwynne. I agree, I think it is better to focus on adding custom formatter options. I will think about the output stream problem as well and I could tackle it as next steps.

nodo:
@mattwynne @brasmusson as discussed I have opened a PR to add per-formatter options (https://github.com/cucumber/cucumber-ruby/pull/1058). Any comments or remarks is welcome, hope it will be useful.

PR #957, opened 2016-03-22
Load gems from local Gemfile which is not part of the repository

maxmeyer:
## Description

I normally use some gems for debugging etc. which are quite "individual". This PR adds the possibility to have an local `Gemfile.local` which is not part of the cucumber repository. In this Gemfile a developer can have "his/her" private gems which are not shared with the community because each developer has his/her own workflow/tool chain.

Using a git ignored `Gemfile.local` gives developers a "convenient" way to have their own gems installed - e.g `pry` + `byebug` - while not littering the projects `Gemfile` with developer's specific gems. If you're interested I can also submit a PR for cucumber-ruby-core.

## Motivation and Context

Developer can have "his/her" private gems which are not shared with the community because each developer has his/her own workflow/tool chain.

## How Has This Been Tested?

Tested locally with own `Gemfile.local`

## Types of changes

## Checklist:

Any thoughts about this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ Gemfile.lock
 tags
 .ruby-version
 .project
+Gemfile.local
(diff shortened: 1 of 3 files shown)


maxmeyer:
The build error is not related to this PR. The `mime-types`-gem makes the build fail for rubies < 2. I send you another PR to fix this.

PR #958, opened 2016-03-23
Set $stderr/$stdout to mocked io in runner

maxmeyer:
Set $stderr/$stdout to mocked io in runner

## Description

Set $stderr/$stdout to mocked io in runner so that the output is captured correctly if someone uses $stdout.puts or $stderr.puts inside cucumber. This trival changes makes testing output on $stderr/$stdout a lot of easier. Otherwise some quirks are needed to capture and check them using steps from `aruba`.

I found that one while implementing the new `options`-API for formatters. Without that change testing is cumbersome.

## Motivation and Context

I would like to be able to capture `puts` output to make feature https://github.com/cucumber/cucumber-ruby/pull/954/files#diff-61df043d3516e6b2647a8c992f671037R70 pass.

## How Has This Been Tested?

* Tested via cucumber run
* I verified the output of cucumber

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/cli/main.rb
+++ b/lib/cucumber/cli/main.rb
@@ -14,9 +14,9 @@ module Cucumber
 
       def initialize(args, _=nil, out=STDOUT, err=STDERR, kernel=Kernel)
         @args   = args
-        @out    = out
-        @err    = err
         @kernel = kernel
+        $stdout = @out = out
+        $stderr = @err = err
       end
 
       def execute!(existing_runtime = nil)


mattwynne:
I don't understand the use-case for this. Can you show me the code that needs it?

maxmeyer:
Sorry, that I was not that explicit about the reason behind the change.

1. `cucumber` uses aruba's `InProcess`-feature
2. In `lib/cucumber/cli/main.rb` a user can pass STDOUT and STDERR. This is used for the tests
3. In https://github.com/cucumber/cucumber-ruby/pull/954/files#diff-61df043d3516e6b2647a8c992f671037R70 I need to output information. I see two options for this

  1. Use `config.out_stream.print`
  2. Use `puts`

  Unfortunately `#out_stream` is not really an option for this example. It refers to the `out`-file. `puts`'s output is not captured by the `Then it should pass with exactly:` step, since `puts` does not use the the StringIO-object passed into `Cli::Main.new`.

If I change the code of `Cli::Main.new` to also set `$stdout` and/or `$stderr` all output generated by `$stderr.puts`, `$stdout.puts` and `puts` is captured and can be validated by `Then it should pass with exactly:` step.

I also added an description to `aruba`'s documentation some time ago. Hope this also makes things clearer: https://github.com/cucumber/aruba/blob/0a3bffa5817911b4f72e9bb286582fb325d75780/features/steps/command/in_process.feature#L315.

maxmeyer:
The failing test is not related to this PR. It's `mime-types` which fails on ruby < 2.0.0 I send you another PR to fix that.

mattwynne:
Right I think I understand. You want a way to print to the process's STDOUT from the custom formatter created in the test scenario so that you can validate the behaviour of the code you're testing by reading STDOUT. You can't just use `puts` because way Cucumber captures calls to `puts` and sends them through the formatters (and thus into the output file).

Did you consider using `$stderr.puts` directly in the custom formatter? I think that's what I normally do when I want to circumvent the `puts` capture.

dg-ratiodata:
> You want a way to print to the process's STDOUT from the custom formatter created in the test scenario so that you can validate the behaviour of the code you're testing by reading STDOUT. 

Exactly. But I also tried STDERR via $stderr.

> You can't just use puts because way Cucumber captures calls to puts and sends them through the formatters (and thus into the output file).

When I remember correctly only `config.out_stream.print` sends output to the output file. Which is expected for this test case.

> Did you consider using $stderr.puts directly in the custom formatter? I think that's what I normally do when I want to circumvent the puts capture.

Yes. I tried all three and they all fail.

* `$stderr.puts`
* `$stdout.puts`
* `puts`

*Background:*

In `aruba` we pass in [`StringIO`](https://github.com/cucumber/aruba/blob/018b486a96c7279da0e83b06e8f05cd604dcc38b/lib/aruba/processes/in_process.rb#L67)'s to `cucumber`'s [runner](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/main.rb#L15). In the steps we use all output written to those `StringIO`-objects. Or with other words `aruba` mocks stdout/stderr via arguments in `Cucumber::Cli::Main.new`. Those passed objects are used for all latter tests in the steps.

*Problem*

Without the change proposed in this PR, the following is true - at least from my understanding. This seems to be proofed because without the change using `puts/$stderr.puts` the output of cucumber is cluttered.

~~~
$stderr != @stderr # from Aruba
$stdout != @stdout # from Aruba
$stderr == STDERR
$stdout == STDOUT
~~~

*One possible solution* 

Within the `Main`-class you've got access to `@err` e.g. [in this line](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/main.rb#L45), but not in other "contexts", because `@err`/`@out` not passed down to other classes. I think the best we can do is overwriting `$stderr` and `$stdout` since they are global and accessible from each and every context.

With this PR merged:

~~~
$stderr == @stderr # from Aruba
$stdout == @stdout # from Aruba
~~~

And even a simple `puts` works - at least with ruby 2.3.0.

Does this makes sense to you @mattwynne

mattwynne:
I'm still finding this one a bit difficult to parse, sorry @maxmeyer.

What it comes down to, is that I'm very uncomfortable setting global variables inside this class. If you really need them to be set, can't you do that yourself in the client code?

maxmeyer:
@mattwynne Will have a look at that.

mattwynne:
I understand the problem now, and I think this is the wrong solution.

We need to fix Aruba's InProcess adapter to do this (or something like it) itself, don't we?

dg-ratiodata:
> We need to fix Aruba's InProcess adapter to do this (or something like it) itself, don't we?

Yeah, that seems to be the best idea.

PR #959, opened 2016-03-23
Lock mime-types-data to older version

maxmeyer:
## Reason:

Build fails on ruby 1.9.3 due to ruby version conflict for a gem.

## Description

This `cucumber-ruby`-[build](https://travis-ci.org/cucumber/cucumber-ruby/jobs/118065154) fails on ruby 1.9.3. A newer version of `mime-types` requires ruby 2.0, so installing it on ruby 1.9.3 fails. This gem is required by 'capybara'. `capybara` is required by the devs deps in the gemspec.

~~~
Gem::InstallError: mime-types-data requires Ruby version >= 2.0.
~~~

## Motivation and Context

Make the build pass again.

## How Has This Been Tested?

This is tested via Travis.

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,3 +11,5 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
     gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
   end
 end
+
+gem 'mime-types', '~>2.99'


brasmusson:
Is it possible to lock the mime-types to an older version only when executing on Ruby < v2.0?
Should this be done in the cucumber.gemspec where 'capybara' is required instead?

aslakhellesoy:
I suggest we drop support for 1.9

dg-ratiodata:
@brasmusson The following snippet should do the trick.

~~~ruby
platform :ruby_19 do
  gem 'mime-types', '~>2.99'
end
~~~

mattwynne:
> I suggest we drop support for 1.9

I agree but that would need to go into v3.0

dg-ratiodata:
Right now all PRs are failing because of this. When do you think you can/want to release 3.0? Or does it make sense to add the Gemfile-snippet as workaround for now?

mattwynne:
Definitely do the workaround for now. We can deal with removing 1.9 support later, that's a whole other can of worms.

PR #960, opened 2016-03-24
Replace git through https as transport protocol

dg-ratiodata:
## Summary

Replace git through https as transport protocol  since the latter is p…roxyable

## Details

If a developer is required to access the WWW via a HTTP proxy, git cannot be used as transport protocol since it is not proxyable. `git` is set in https://github.com/cucumber/cucumber-ruby/blob/master/Gemfile#L11.

Old:
~~~
gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
~~~

New:
~~~
gem 'cucumber-core', :git => "https://github.com/cucumber/cucumber-ruby-core.git"
~~~

There might be other cucumber projects where such a change makes sense.

## Motivation and Context

In bigger companies the use of HTTP proxies is quite common. There's not direct access from a laptop/computer to the WWW. Using `git` as transport protocol is a burden for new developers from such an environment as a local change is required each time they want to send a pull request. Since I've never had any performance impact in using HTTPS instead of GIT, put together this PR.

## How Has This Been Tested?

Yes, many times locally.

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ unless ENV['CUCUMBER_USE_RELEASED_CORE']
     gem 'cucumber-core', :path => core_path
     gem 'cucumber-wire', :path => wire_path
   else
-    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
+    gem 'cucumber-core', :git => "https://github.com/cucumber/cucumber-ruby-core.git"
   end
 end
 


danascheider:
Looks good to me.

dg-ratiodata:
@danascheider Thank you! :-)

PR #962, opened 2016-03-29
WIP: Event stream formatter

mattwynne:
The goal of this ticket is to produce a stream of JSON events that conform to the [standardised cucumber event protocol](https://github.com/cucumber/cucumber/tree/master/event-protocol)

This will allow cucumber to interoperate with other tools, and ultimately perhaps enable us to use a cross-platform formatter library.

The event protocol is new and emerging, and this PR aims to keep up with that.


features/docs/formatters/event_stream_formatter.feature
```diff
@@ -0,0 +1,53 @@
...
+        Scenario:
+          Given this step passes
+      """
+    When I run `cucumber -q --format Cucumber::Formatter::EventStream`
+    Then it should pass
+    And the stdout should contain exactly:
+      """
+      {"event":"TestRunStarted","protocol_version":"0.0.1"}
+      {"event":"GherkinSourceRead","id":"features/simple.feature:1","source":"Feature:\n  Scenario:\n    Given this step passes"}
+      {"event":"StepDefinitionFound","source_id":"features/simple.feature:3","definition_id": "features/step_definitions/steps.rb:1"}
```
aslakhellesoy: Could this be named `StepDefinitionMatched` instead? I think *found* can possibly be mistaken with *loaded*.

features/docs/formatters/event_stream_formatter.feature
```diff
@@ -0,0 +1,53 @@
...
+      {"event":"TestCaseStarted","id":"features/simple.feature:2"}
+      {"event":"TestStepStarted","id":"features/simple.feature:3"}
+      {"event":"TestStepPassed","id":"features/simple.feature:3"}
+      {"event":"TestCasePassed","id":"features/simple.feature:2"}
```
aslakhellesoy: I think a more robust event would be:
`{"event":"TestCaseFinished","status": "passed", "id":"features/simple.feature:2"}`

l3pp4rd: Failed or Undefined events are different because they have more details:

``` json
{"event":"TestCaseUndefined", "summary": "TODO: implement step definition", "snippet": "some src snippet may be provided"}
```

Though, these can optional fields. I agree it could be a single event in this case, because we have few statuses:

- passed
- failed
- skipped
- undefined
- ambiguous
- pending

l3pp4rd:
One more thing to point out, for pretty formatter. We probably need argument list to make them bold. Which should most probably go with **StepDefinitionFound** event:

``` json
{
  "event":"StepDefinitionFound",
  "source_id":"features/simple.feature:3",
  "definition_id": "features/step_definitions/steps.rb:1",
  "arguments": [
    [5, 10],
    [16, 20]
  ]
}
```

Which could mark positions from - to in step.Text.

aslakhellesoy:
@l3pp4rd yes, agree regarding arguments. The offsets must be from the beginning of the line, since the formatter won't know at what column the step text starts (the text that was matched, following the step keyword).

aslakhellesoy:
@mattwynne do you want to keep this PR open? Should I try to move it forward or do you recommend starting over?

mattwynne:
@aslakhellesoy I think it's practically done. The only criticism I'd have is that it wasn't built test-first - I mostly manually tested it with the electron cucumber-gui spike, though it does (or did) work against the protocol described in https://github.com/cucumber/cucumber/pull/172#issuecomment-347219777 so that was part of testing it too.

I'd have a crack and see if you can stabilise it.

mattwynne:
I think we can close this one now! 🎉

Issue #951, opened 2016-02-19
Scenario reports misleading state/result data in hooks


borsothy:
First of all, our basic setup is that we are invoking scenarios from an `Around` block, but I doubt it actually matters in this particular case.

The problem is that if an exception is raised in one of the steps, the scenario's state is not updated up until the `After` hook. If you call `scenario.passed?` in either `Around` or `AfterStep`, it returns `true`.

What happens is that `passed?` just calls `failed?`, negates the result, and returns that value. Since `failed?` returns `false` because the state of the scenario is still `:unknown`, when you call `passed?`, it's going to return `true`, which is not actual case.

Please note this is a bit similar to issue #880, but not quite the same.

mattwynne:
Around and AfterStep hooks are very different.

Around hooks actually surround the whole clump of code that executes the scenario. They are complex and personally I find them confusing to reason about. It doesn't surprise me at all if they report misleading results.

AfterStep hooks are just compiled in as `Test::Step`s so it should be straightforward to report their status.

The status of the scenario should only be `:unknown` until the first test step (that includes any `Before` hooks) has been run. After that, it should reflect the status of the last step (`passing`, `failing`).

It would be very helpful if you could look through our features and add a couple of scenarios that describe the behaviour you want and isolate these bugs. That will help them get fixed more quickly. In any case, I appreciate your taking the time to give us this feedback.

gergelyborsothy:
Sorry for the delay, I created a PR (#963) with a new scenario which I believe covers the issue I raised above.

PR #963, opened 2016-03-30
Example scenario for misleading result data in Around hook

borsothy:
Companion PR for issue #951.

### DIFF
--- a/features/docs/writing_support_code/around_hooks.feature
+++ b/features/docs/writing_support_code/around_hooks.feature
@@ -38,6 +38,36 @@ Feature: Around hooks
 
       """
 
+  Scenario: Check the failed status of a scenario in a hook
+    Given a file named "features/support/debug_hook.rb" with:
+      """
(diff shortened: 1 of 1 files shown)


mattwynne:
Thanks, this will make it a lot easier for us to get this working.

danascheider:
@gergelyborsothy Can you tag this as `@wip`? That way we can merge this without breaking Travis while we investigate the issue further.

gergelyborsothy:
@danascheider I can't apply labels 😦

danascheider:
Oh, @gergelyborsothy, I meant tag the feature as @wip, not the issue:
```gherkin
@wip
Feature: Whatever the feature is
  Scenario: Whatever the scenario is
  ...
```

danascheider:
Actually, @gergelyborsothy, never mind - I just saw a new PR changing that so `@wip` features will break the builds after all. So what we'll do with this is leave it open until the right code is added to make the feature pass.

PR #965, opened 2016-04-04
fix HTML formatter backtrace duplicate line bug

josephks:
An alternative to #910.  #910 attempts to fix the problem by altering the regular expression, which seems strange to me.  It could be that the regular expression needs fixing but I am experiencing this bug and I don't have `ENV['TM_PROJECT_DIRECTORY']` set at all.  If `ENV['TM_PROJECT_DIRECTORY']` isn't set at all why even try to match the regular expression against the input?

It should be noted that `line` is not in fact a single line.  It is the backtrace array which has already been joined with newlines.

I also noticed that the code is running `delete_if` on the array it gets from  `exception`.  Mutating objects that do not belong to us is bad practice.  A better practice would be to use `reject` instead to get a new array.  I'm keeping this PR focused on the actual bug though.

### DIFF
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -589,13 +589,13 @@ module Cucumber
         (["#{exception.message}"] + exception.backtrace).join("\n")
       end
 
-      def backtrace_line(line)
-        line.gsub(/^([^:]*\.(?:rb|feature|haml)):(\d*).*$/) do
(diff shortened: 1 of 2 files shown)


danascheider:
Any thoughts on this @brasmusson?

mattwynne:
Thanks @josephks, especially for adding a spec as you went.

Thanks also for keeping it focussed, and if you'd like to attack either of those refactorings you suggested that would be much appreciated.

PR #972, opened 2016-04-20
Demonstrate suggested way to capture output

mattwynne:
@maxmeyer this is my suggestion for how to deal with the issue we've
been discussing in #958, at least for now.

I think I understand your issue now (it's s00oo much easier with real code
in my hands sometimes!) and what I may have missed suggesting is to use
the `@spawn` tag which uses the "tranditional" adapter for Aruba instead
of the in-process one.

This commit shows you how you'd do that.

### DIFF
--- a/features/docs/extending_cucumber/custom_formatter.feature
+++ b/features/docs/extending_cucumber/custom_formatter.feature
@@ -60,6 +60,7 @@ Feature: Custom Formatter
       {"foo"=>"bar", "one"=>"two"}
       """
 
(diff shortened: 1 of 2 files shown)


maxmeyer:
Thanks @mattwynne merged it. But yes, I think we should think about refactor the `InProcess`-runner.

dg-ratiodata:
Opened an issue to track this: https://github.com/cucumber/aruba/issues/381

PR #974, opened 2016-04-26
Make random order stable and platform-independent

### LINKED ISSUE #971: Seeded order is platform-dependent
When using `--order=random` with a seed, the order is determined by Ruby's `Array#shuffle` method.

Seeded `Random` produces consistent results across different versions of Ruby and JRuby, but `shuffle` is implemented differently in different interpreters:

Ruby 2.2.4:

```
[1,2,3].shuffle(random: Random.new(4))
=> [2, 1, 3]
[1,2,3].shuffle(random: Random.new(5))
=> [1, 2, 3]
```

Ruby 1.9.3 and JRuby 1.7.23:

```
[1,2,3].shuffle(random: Random.new(4))
=> [1, 2, 3]
[1,2,3].shuffle(random: Random.new(5))
=> [3, 2, 1]
```

As I discovered whilst working on #970, this makes it difficult to write specifications that will work across all supported platforms. The current [`randomize.feature`](https://github.com/cucumber/cucumber-ruby/blob/7f917e5795021557eaa7312091c961463ff5089a/features/docs/cli/randomize.feature) effectively works only by accident because it shuffles only two items.

This also means that it's not always possible to replicate a particular seeded order on a different interpreter.

RSpec has a similar random ordering feature that uses a platform-independent [sort on a hash of the item index and the seed](https://github.com/rspec/rspec-core/blob/243c08d83fe81887aabace8ea34a9a1ee445585e/lib/rspec/core/ordering.rb#L25-L59). This seems like a more robust thing for `cucumber-ruby` to do, although it would come at the cost of invalidating any existing use of seeds.

threedaymonk:
This makes two important changes to random ordering.

1. Using the `--order=random` option with a particular seed will now produce the **same order** on every Ruby interpreter.
2. The order is **stable**, so that if step A runs before step B with a given seed, it will still do so for that seed and suite even if intervening steps are skipped.

Previously, the random seed would produce consistent results on a given interpreter, but this could not always be reproduced on another version of Ruby. This occurred because `Array#shuffle` is implemented differently in different interpreters: Ruby 1.9.3 and JRuby 1.7.x share one implementation, whilst Ruby 2.x has a different one.

Stability of ordering is a necessary precondition for using bisection to resolve an ordering dependency (as RSpec can do), should we wish to add that in the future.

Ordering is now implemented by sorting on the 256-bit SHA2 hash of the string representation of the seed plus the index of the scenario. As randomisation takes place before non-matching steps are discarded, stability is maintained.

Fixes #971.

### DIFF
--- a/features/docs/cli/randomize.feature
+++ b/features/docs/cli/randomize.feature
@@ -44,7 +44,7 @@ Feature: Randomize
 
   @spawn
   Scenario: Run scenarios randomized
-    When I run `cucumber --order random:41529 -q`
+    When I run `cucumber --order random:41515 -q`
     Then it should fail
     And the stdout should contain exactly:
       """
@@ -76,7 +76,7 @@ Feature: Randomize
       3 scenarios (2 failed, 1 passed)
       3 steps (2 failed, 1 passed)
       
-      Randomized with seed 41529
+      Randomized with seed 41515
       
       """
 
--- a/features/docs/cli/randomize.feature
+++ b/features/docs/cli/randomize.feature
@@ -9,6 +9,10 @@ Feature: Randomize
   you can reproduce that run by using the seed that's printed at the end of
   the test run.
 
+  For a given seed, the order of scenarios is constant, i.e. if step A runs
+  before step B, it will always run before step B even if other steps are
+  skipped.
+
   Background:
     Given a file named "features/bad_practice_part_1.feature" with:
       """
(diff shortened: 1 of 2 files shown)


danascheider:
Looks good @threedaymonk, thanks!

brasmusson:
@danascheider Do not forget to update the [History.md](https://github.com/cucumber/cucumber-ruby/blob/master/History.md), when you merge a PR (this PR, and issue #971 that it closes, needs to be mentioned there now).

danascheider:
Sorry, will do!

mattwynne:
I often forget too. It's a hassle but I think it's well worth it. I just wish there was some automatic way of doing it...

PR #977, opened 2016-05-08
Change the Progress, Usage and Stepdef formatter use events

brasmusson:
Change the Progress, Usage and Stepdef formatter use events and also the new initialize method signature.

For the Usage formatter to be able to find out which step definitions that were not matched to any step, the StepDefinitionLoaded event is added. This also means that the step definitions are loaded after the AfterConfigurationHooks have been executed and the formatter have been created, whereas the support files are loaded before the AfterConfigurationHooks are executed.

The API:s of Cucumber::Formatter::Console and Cucumber::Formatter::Summary are extended, but are backward compatible as they are use both by the Pretty formatter and existing custom formatters.

To simplify the usage of Cucumber::Core::Test::Result::Summary its API has been extended in https://github.com/cucumber/cucumber-ruby-core/commit/e7fb8b77e99bc881e254a9c18a1b3c92e435af3c (already on master of Cucumber-Ruby-Core).

### DIFF
--- a/features/docs/defining_steps/printing_messages.feature
+++ b/features/docs/defining_steps/printing_messages.feature
@@ -135,7 +135,6 @@ Feature: Pretty formatter - Printing messages
         Multiple
         
         Announce
-        
         Me
         ..UUU
         Announce with fail
--- a/lib/cucumber/configuration.rb
+++ b/lib/cucumber/configuration.rb
@@ -89,6 +89,30 @@ module Cucumber
       @options[:expand]
     end
 
+    def source?
+      @options[:source]
+    end
+
+    def duration?
+      @options[:duration]
+    end
+
+    def snippets?
+      @options[:snippets]
+    end
+
+    def skip_profile_information?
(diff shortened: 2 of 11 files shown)


lib/cucumber/events/step_definition_loaded.rb
```diff
@@ -0,0 +1,20 @@
+module Cucumber
+  module Events
+
+    # Event fired after each step definition has been loaded
+    class StepDefinitionLoaded
```
mattwynne: I think I like "registered" better than "loaded". WDYT?

mattwynne:
Great work @brasmusson! I like the use of the new event - I'd been thinking along the same lines, and it's great to see the code for real 👍

lib/cucumber/events/step_definition_loaded.rb
brasmusson: Considering that the event is fired in the method `register_rb_step_definition`, you are absolutely right.

PR #980, opened 2016-05-19
Refactor cli options

danascheider:
## Summary

This PR refactors the `#parse!` method for the `Cucumber::Cli::Options` class.

## Details

I extracted descriptions and option handlers to new private methods and turned the `opts.on` blocks into single lines.

## Motivation and Context

I noticed that Code Climate had rated this class F, and when I looked at the code, I could see why. I think the new way is more readable.

## How Has This Been Tested?

I tested the changes using the existing test suite to check for regressions. Since I didn't add or remove functionality, new tests were not required.

### DIFF
--- a/lib/cucumber/cli/options.rb
+++ b/lib/cucumber/cli/options.rb
@@ -88,19 +88,10 @@ module Cucumber
 
         @args.options do |opts|
           opts.banner = banner
-          opts.on("-r LIBRARY|DIR", "--require LIBRARY|DIR", *require_files_msg) do |v|
-            @options[:require] << v
-            if(Cucumber::JRUBY && File.directory?(v))
-              require 'java'
-              $CLASSPATH << v
-            end
-          end
+          opts.on("-r LIBRARY|DIR", "--require LIBRARY|DIR", *require_files_msg) {|lib| require_files(lib) }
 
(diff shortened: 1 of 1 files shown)


mattwynne:
I love the extracted helper methods. I wonder if, to keep the abstraction level consistent in the `opts.on` blocks, it would be better always use a helper method, even when it's only a one-liner. WDYT?

mattwynne:
Maybe two methods: `set_option` and `unset_option` would do.

e.g. `set_option :strict`

danascheider:
I love that idea.

danascheider:
@mattwynne Was there anything else you thought should be changed here?

mattwynne:
No, looks bloody marvellous to me @danascheider! I'm really happy to see you cleaning the code up like this.

mattwynne:
I know it feels a bit weird merging your own PR so I did it 😄 

I don't personally think this needs to go into `History.md` as it doesn't change any externally-visible behaviour but I may be biased by the fact that I'm doing this via the GitHub UI, it's late and I'm feeling a bit lazy. @brasmusson you tend to have good opinions about these things - WDYT?

danascheider:
Thanks @mattwynne!

PR #981, opened 2016-05-20
Remove tolerance for @wip scenarios in build

mattwynne:
## Summary

Clean up `@wip` scenarios from the default profile config, and avoid them being allowed again.

## Details

When I merged in #920, I missed the fact that the scenarios in that PR still had a `@wip` tag and were in fact still failing. This was actually expected, as we'd already discussed in that PR that the summary output is currently unable to handle retried scenarios.

## Motivation and Context

We want our build to fail if there are any `@wip` tagged scenarios that are failing, so we don't build up technical debt.

## To do

- [x] change the two `@wip` scenarios to match the current output
- [x] remove the `@wip` tags
- [x] create a new issue to specify the right output for retry (and link it to #920 because it will be much easier to implement once that is done)


mattwynne:
Closed in c253b06e6c2fa77c245ae8787dc5812c693d8e00

PR #982, opened 2016-05-26
Remove wip tag

danascheider:
## Summary

This PR removes the `@wip` tag from `retry_failing_tests.feature`.

## Details

Since `retry_failing_tests.feature` is now fully implemented and passes, it should not be tagged as `@wip` as this will break the build after #981 is merged.

## Motivation and Context

Since we have decided not to allow `@wip` features on master, we need to remove the `@wip` tag from features already on master (and, of course, fix ones that don't pass). This will prevent the build from breaking due to the presence of features that should not have been merged. (Something tells me this whole paragraph doesn't make much sense.)

## How Has This Been Tested?

The feature was already passing, so removing the tag doesn't require additional tests.

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
 
 ### Bugfixes
 
+* Use HTTPS instead of Git as transport protocol ([#960](https://github.com/cucumber/cucumber-ruby/pull/960))
 * Make random order stable and platform independent ([#974](https://github.com/cucumber/cucumber-ruby/pull/974), closes [#971](https://github.com/cucumber/cucumber-ruby/issues/971))
 * Run scenarios in fully random order ([#970](https://github.com/cucumber/cucumber-ruby/pull/970) @threedaymonk)
 * Adding Test Step in AfterStep hook. ([#931](https://github.com/cucumber/cucumber-ruby/pull/931) @t-morgan)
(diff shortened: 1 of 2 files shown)


History.md
```diff
@@ -6,6 +6,7 @@
 
 ### Bugfixes
```
danascheider: This change was not supposed to be in this PR! I thought I had pushed this directly to master previously :(

danascheider:
Oh no, looks like this feature doesn't, in fact, pass, which means we probably shouldn't have merged #920. I'll have to look into what's going on.

brasmusson:
@danascheider The retry filter is never added to the filters in the [runtime](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L228-L246). Your step definitions for the retry feature does not work. One problem is that their state is not reset between scenarios. I also made [this comment](https://github.com/cucumber/cucumber-ruby/pull/920/files#r64802812) on the PR.

danascheider:
Oh sorry, I didn't see that! I'll check it out when I get to the office.

mattwynne:
See also #981 :)

mattwynne:
Closed in favour of #981

Issue #961, opened 2016-03-24
Local clones make and missing clone of extracted gem `build install` fail


dg-ratiodata:
## Summary

Extracted wire-protocol-gem makes `bundle install` fail.

## Expected Behavior

~~~
bundle install
~~~

should succeed.

## Current Behavior

~~~
bundle install
~~~

fails with 

~~~
The path `/home/user/data/projects/cucumber-ruby-wire` does not exist.
~~~

## Possible Solution

Since I opened another PR #960 a few minutes ago which affects the same lines
but because of a different thing, I just openened an issue.

~~~ruby
gem "cucumber-pro", "0.0.13", :group => :test
source "https://rubygems.org"
gemspec
unless ENV['CUCUMBER_USE_RELEASED_CORE']
  core_path = File.expand_path("../../cucumber-ruby-core", __FILE__)
  wire_path = File.expand_path("../../cucumber-ruby-wire", __FILE__)

  if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
    gem 'cucumber-core', :path => core_path
  else
    gem 'cucumber-core', :git => "git://github.com/cucumber/cucumber-ruby-core.git"
  end

  if File.exist?(wire_path) && !ENV['CUCUMBER_USE_GIT_CORE']
    gem 'cucumber-wire', :path => wire_path
  else
    gem 'cucumber-wire', :git => "git://github.com/cucumber/cucumber-ruby-wire.git"
  end
end
~~~

Additionally we could add a bootstrap-script. Something really simple:

script/bootstrap
~~~
#!/usr/bin/env bash

git clone https://github.com/cucumber/cucumber-ruby
git clone https://github.com/cucumber/cucumber-ruby-core
git clone https://github.com/cucumber/cucumber-ruby-wire
~~~

This might be even useful for other cucumber projects. Maybe something similar like [Github](http://githubengineering.com/scripts-to-rule-them-all/) does.

## Steps to Reproduce (for bugs)

1. `git clone https://github.com/cucumber/cucumber-ruby`
2. `git clone https://github.com/cucumber/cucumber-ruby-core`
3. `cd cucumber-ruby`
4. `bundle install`

## Context

I had `cucumber-ruby-*` locally before the `wire`-protocol was extracted to a separate gem.

## Your Environment

* Version used: master
* Operating System and version: Not relevant in this case
* Link to your project: Not relevant in this case

danascheider:
[...]

PR #983, opened 2016-05-28
Resolve issue 961

danascheider:
## Summary

This PR resolves issue #961, where `bundle install` fails because of `cucumber-ruby-wire`. It makes changes to the Gemfile so that if the user does not have `cucumber-ruby-wire` locally, or doesn't have it in the expected location, it will be fetched from git instead of erroring out.

## Details

Added a conditional to the Gemfile to check whether the `cucumber-ruby-wire` code exists locally in the expected location and, if not, gets it from git instead.

## Motivation and Context

This issue was causing errors in `bundle install`, making it harder for developers to work on Cucumber.

## How Has This Been Tested?

Gemfile change probably doesn't need tests.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,11 +4,16 @@ gemspec
 unless ENV['CUCUMBER_USE_RELEASED_CORE']
   core_path = File.expand_path("../../cucumber-ruby-core", __FILE__)
   wire_path = File.expand_path("../../cucumber-ruby-wire", __FILE__)
-  if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
-    gem 'cucumber-core', :path => core_path
-    gem 'cucumber-wire', :path => wire_path
+  if File.exist?(core_path) && !ENV["cucumber_use_git_core"]
+    gem "cucumber-core", :path => core_path
   else
-    gem 'cucumber-core', :git => "https://github.com/cucumber/cucumber-ruby-core.git"
+    gem "cucumber-core", :git => "git://github.com/cucumber/cucumber-ruby-core.git"
+  end
+
+  if File.exist?(wire_path) && !ENV["CUCUMBER_USE_GIT_CORE"]
+    gem "cucumber-wire", :path => wire_path
+  else
+    gem "cucumber-wire", :git => "git://github.com/cucumber/cucumber-ruby-wire.git"
   end
 end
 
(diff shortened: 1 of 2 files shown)


History.md
```diff
@@ -6,6 +6,7 @@
 
 ### Bugfixes
 
+* Use HTTPS instead of Git as transport protocol ([#960](https://github.com/cucumber/cucumber-ruby/pull/960))
```
danascheider: This commit is haunting me, it wants to be in all my PRs for some reason!!

brasmusson: Of course, it is not on `cucumber:master` but it is on the PR branch. This time (but not in the case of #982), its parent commit is the tip of `cucumber:master` so I pushed it to `cucumber:master`, it should disappear from this PR now.

danascheider: Thank you!

danascheider:
Any objections to my merging this @mattwynne or @brasmusson? Speak now or forever hold your peace!

mattwynne:
👍 but bear in mind we should make the wire plugin opt-in for 3.0 anyway so was can remove it as a dependency now I think 😀 Right @tooky?

Gemfile
```diff
@@ -4,11 +4,16 @@ gemspec
...
   core_path = File.expand_path("../../cucumber-ruby-core", __FILE__)
   wire_path = File.expand_path("../../cucumber-ruby-wire", __FILE__)
-  if File.exist?(core_path) && !ENV['CUCUMBER_USE_GIT_CORE']
-    gem 'cucumber-core', :path => core_path
-    gem 'cucumber-wire', :path => wire_path
+  if File.exist?(core_path) && !ENV["cucumber_use_git_core"]
+    gem "cucumber-core", :path => core_path
   else
-    gem 'cucumber-core', :git => "https://github.com/cucumber/cucumber-ruby-core.git"
+    gem "cucumber-core", :git => "git://github.com/cucumber/cucumber-ruby-core.git"
```
dg-ratiodata: Please don't use git:// this cannot be forwarded via HTTP proxies.

danascheider: Oh jeez, can't believe I didn't notice that. Thanks for pointing that out - will change.

dg-ratiodata: Thanks a lot. :-)

Issue #984, opened 2016-06-03
Scenario outline with examples in title


rKrums:
Regarding this: https://github.com/cucumber/cucumber-js/issues/267

Greetings!
Do we have anything similar in ruby? 

> marcfallows commented on Oct 23, 2014
> Hello,
> 
> I would expect to be able to have my scenario outline title have more information by using the examples >within the title itself:
> 
> ```
> Scenario Outline: A <some> step is <result>
>    When a <some> step
>    Then I get <result>
> Examples:
>    | some    | result  |
>    | passing | passed  |
>    | failing | skipped |
> ```
> 
> Then my scenario titles end up very useful:
> `Scenario: A passing step is passed
> Scenario: A failing step is skipped`
> 
> This doesn't seem to be working for me. Is this a supported feature? If not, is it something that I could >assist in having included? Or, is it not expected to work like this?

If not, it would be very useful to have something like this. It would increase the readability of reports with alot of example data.

danascheider:
I really like this idea too, and I also think it's good to try to have some consistency to where features available in one implementation of Cucumber should also be available in others. Would you be able to put together a PR for this, @rKrums? I'd be happy to help out or pair on this.

tooky:
I'm sure we could do this in the compiler in cucumber-core, but I wonder if it will need to go into gherkin's pickles compiler in the long run.

cc @brasmusson @aslakhellesoy

brasmusson:
The gherkin compilers do implement this already, see https://github.com/cucumber/gherkin/blob/master/testdata/good/example_tokens_everywhere.feature and https://github.com/cucumber/gherkin/blob/master/testdata/good/example_tokens_everywhere.feature.pickles.json

rKrums:
@danascheider I'm on a really tight schedule right now, and would not be able to participate. Thank you for offer. I understand, that it's possible to do that already, but have not checked.

sharmarahul:
Any plans of having this feature in the Ruby version?

danascheider:
@sharmarahul This is already possible.

brasmusson:
@danascheider No, this is not possible in Cucumber-Ruby, as it only uses the Gherkin parser and implements its own (AST) compiler (instead of using Gherkins compiler - which did not exist when the compiler in Cucumber-Ruby-Core was implemented).

danascheider:
[...]

PR #985, opened 2016-06-05
Json formatter scenario outline table data in

ransom4real:
## Summary

This PR is meant to provide clarity in JSON Formatter id for Scenario outlines. At the moment this returns just the example row id which is not very descriptive for anyone using the JSON report on Jenkins or any other reporting tool. A comma separated table header and row value seems more appropriate.

## Details

Hence I have written a private helper method to extrapolate the header values, loop through each value using its index and fetch the associated row value using the index of the header as table columns have fixed indexes which means the row values have fixed indexes correlating to the header index. This way i can build a comma separated column=value data to append to the id instead of just an row id which doesn't give any descriptive information

## Motivation and Context

JSON reporting has become the go to formatter for test reporting especially when you have to rerun tests and merge in results from your rerun into the original test report. However this becomes more of an issue with scenario outlines when you try to find out what exact outline example is failing as the current JSON formatter does not give an example data context. This is my motivation for this change. It does not fix any open issues at the moment and it is more of a feature which i reckon will make the JSON formatter used even more.

## How Has This Been Tested?

I have updated all JSON related scenario outline based specs and features. I have also added new tests relating to tables with more than one column to ensure they are mapped correctly and comma separated as required. All tests were executed prior to this PR and they all pass.

## Types of changes

## Checklist:


brasmusson:
Changes to the Json formatter output should be coordinated with the other Cucumber implementations (maybe primarily [Cucumber-JVM](https://github.com/cucumber/cucumber-jvm)), so that tools that process the Json files can be used with more than one Cucumber implementation.

I think that the important trait of the id attribute in the Json file is uniqueness rather than descriptiveness. I would look for descriptiveness in that name attribute and other data attributes. Issue #984, points to a better solution in that respect, I think.

ransom4real:
@brasmusson i did give issue #984 a go but had no luck getting the scenario example data into the title. Also the reason i have fallen back on the id in the JSON is because the name and other attributes are not unique enough especially for Scenario Outlines. I never had issues with normal Scenarios just Outlines. Also it seems #984 is implemented in cucumber-JVM but not really in cucumber-ruby. Are you certain it is possible in cucumber-ruby? Do you reckon this solution will break tools that process JSON reports considering all the spec test and gherking features passed your build process?

danascheider:
@ransom4real Take a look at the conversation on #984 - this is entirely possible but would need to be implemented in [core](https://github.com/cucumber/cucumber-ruby-core).

ransom4real:
@danascheider do you mean #984 or my solution?

danascheider:
984.

ransom4real:
@danascheider i might not get the time to pair on this as this PR was meant to help my work. I will go ahead and close this PR in preference to the #984 solution whenever it is implemented. I will be happy to extend the JSON formatter in my framework and override the examples_table method for my needs. Thanks

PR #1007, opened 2016-08-19
Add world namespaces

nodo:
~~Explorative PR to discuss about adding namespaces to a `World`.~~

Add namespaces to a World. When adding modules to a World it is possible to specify a namespace using hash syntax (see feature file).


mattwynne:
I like the spec just the way you've done it - lovely!

nodo:
@mattwynne I have implemented the feature and fleshed out the spec file to describe different aspects of the namespaces.

Looking forward to see your comments or questions to improve the implementation.

mattwynne:
Thanks @nodo, I'm reviewing this now.

features/docs/writing_support_code/world_namespaces.feature
```diff
@@ -0,0 +1,186 @@
+Feature: World Namespaces
```
mattwynne: Looking at the existing features, it seems the existing `World` functionality was never properly documented in the acceptance tests.

Could you broaden the scope of this one, renaming it to just `world.feature` and adding two more scenarios:

- the one where you extend the world using a module
- the one where you define a custom class for the world (using a block)

features/docs/writing_support_code/world_namespaces.feature
```diff
@@ -0,0 +1,186 @@
...
+
+  In order to better organise the code supporting my tests
+  As a cucumber user
+  I want to access the method added to a World using an explicit namespace
```
mattwynne: It would be great to have some more explanatory text about this feature. I find the user story format a bit limiting for really good documentation.

Here's a good example: https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/formatters/rerun_formatter.feature

If you're stuck for inspiration I can help out with the copy.

features/docs/writing_support_code/world_namespaces.feature
```diff
@@ -0,0 +1,186 @@
...
+    And a file named "features/f.feature" with:
+      """
+      Feature: Calling a method
+        Scenario: I call a method from two namespaces
+          Then the helper method is called
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass
+
+  Scenario: Assigning different modules to the same namespace
```
mattwynne: While it's useful to know that this works, it does seem like an edge case. Is there a way to push this down to a unit test?

features/docs/writing_support_code/world_namespaces.feature
```diff
@@ -0,0 +1,186 @@
...
+      Feature: Calling a method
+        Scenario: I call a method
+          Then I call a method from module one
+          And I call a method from module two
+          And I call a method defined in both modules
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass
+
+  Scenario: Assigning the same module to the same namespace in different contexts
```
mattwynne: I wonder if this could also be made a unit test?

features/docs/writing_support_code/world_namespaces.feature
mattwynne: Oh wait I see you already have the first one below as `A single module is imported without namespace`

features/docs/writing_support_code/world_namespaces.feature
```diff
@@ -0,0 +1,186 @@
...
+      Feature: Calling a method
+        Scenario: I call a method
+          Then I call a method from module one
+          And I call a method from module two
+          And I call a method defined in both modules
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass
+
+  Scenario: A single module is imported without namespace
```
mattwynne: Let's put this one at the top, since it's the most common case.

lib/cucumber/rb_support/rb_world.rb
```diff
@@ -137,13 +143,47 @@ def inspect
...
+      end
+
+      # @private
+      def add_namespaced_modules!(modules)
+        @__namespaced_modules = modules
+        modules.each do |namespace, world_modules|
+          world_modules.each do |world_module|
+            variable_name = "@__#{namespace}_world"
+
+            miniworld = if self.class.respond_to?(namespace)
```
mattwynne: maybe `namespaced_world` would be a better name? Or `inner_world` :)?

mattwynne:
Lovely clean code @nodo! I've put some comments in, let me know your thoughts.

nodo:
@mattwynne thanks for your helpful comments. I have address them all with https://github.com/cucumber/cucumber-ruby/pull/1007/commits/329c6fd8870fafe5f5278f8212a5840904037193 , paying a special attention to testing. Please, let me know if you have further remarks.

features/docs/writing_support_code/world.feature
```diff
@@ -0,0 +1,129 @@
...
+  1. Passing a module. In this case its methods will be added directly to the
+     world and be usable in the step definitions.
+
+  2. Passing a block, where the return value is the world object.
+
+  3. Passing a hash, where the keys are namespaces and the values are
+     modules. In this case, the methods of each module will be accessible using
+     the key as prefix.
+
+  Scenario: A world is created using a module
```
mattwynne: I think 'extended' is more accurate than 'created' here

features/docs/writing_support_code/world.feature
```diff
@@ -0,0 +1,129 @@
...
+    And a file named "features/f.feature" with:
+      """
+      Feature: Calling a method
+        Scenario: I call a method from a namespace
+          Then the helper method is called
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass
+
+  Scenario: A world is created using a module with namespace
```
mattwynne: again, 'extended' is better than 'created' here, I think

features/docs/writing_support_code/world.feature
```diff
@@ -0,0 +1,129 @@
...
+    And a file named "features/f.feature" with:
+      """
+      Feature: Calling a method
+        Scenario: I call a method from a namespace
+          Then the helper method is called
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass
+
+  Scenario: A world is created using multiple modules with different namespaces
```
mattwynne: 'created' => 'extended'?

mattwynne:
Fantastic work @nodo! I've added a couple of nit-picks on the wording. I really like the docs you've written.

nodo:
Thanks @mattwynne . I agree with your word choice, I added your comments to the last commit.

PR #1011, opened 2016-08-22
AfterAll hook

wjlroe:
## Summary

Carrying on the work done in #858. This pull request aims to add a hook to replace using `Kernel#at_exit` for the end of a cucumber suite's cleanup tasks. 
## Motivation and Context

The problem with using `Kernel#at_exit` for test suite cleanup tasks is that it is very easy to accidentally clobber the exit status of the test suite. If you call any code that throws (and catches) exceptions, that will effectively overwrite the `$ERROR_INFO` value and turn an error status into a success status, which you likely don't want. 

Due to the tricky way in which developers have to treat `Kernel#at_exit`, it would be preferable for cucumber to provide a hook for the same use cases except the exit status of the test suite would be preserved - anything that happens in the hook, you can't "break" the expected behaviour of cucumber.
## Types of changes
## Checklist:
- [x] I've added acceptance tests for my code
- [ ] I've added specs for my code
- [ ] I've completed the implementation for this feature
- [ ] The name of the hook has been discussed and agreed upon.

### DIFF
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -67,6 +67,7 @@ module Cucumber
       receiver = Test::Runner.new(@configuration.event_bus)
       compile features, receiver, filters
       @configuration.notify :test_run_finished
+      fire_after_all_hook
     end
 
     def features_paths
@@ -108,6 +109,10 @@ module Cucumber
 
     def fire_after_configuration_hook #:nodoc
       @support_code.fire_hook(:after_configuration, @configuration)
+    end
+
+    def fire_after_all_hook #:nodoc
+      @support_code.fire_hook(:after_all)
     end
 
     require 'cucumber/core/gherkin/document'
--- a/features/docs/writing_support_code/after_all_hooks.feature
+++ b/features/docs/writing_support_code/after_all_hooks.feature
@@ -136,3 +136,22 @@ Feature: After All Hook
       """
       After All
       """
+
+  Scenario: Exit code is not zero if a test fails
+    Given a file named "features/support/hooks.rb" with:
+      """
+      AfterAll do
+        puts "After All"
+      end
+      """
+    And a file named "features/test.feature" with:
+      """
+      Feature:
+        Scenario:
+          Given this step fails
+      """
+    When I run `cucumber -f progress features`
+    Then it should fail with:
+    """
+    features/test.feature:3
+    """
--- a/features/docs/writing_support_code/after_all_hooks.feature
+++ b/features/docs/writing_support_code/after_all_hooks.feature
@@ -1,0 +1,138 @@
+Feature: After All Hook
+
+  In order to extend Cucumber
+  As a developer
+  I want to run code after all Cucumber tests are finished
+
+  Background:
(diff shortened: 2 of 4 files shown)


wjlroe:
My next action is to dive into the specs to drive these tests closer to the implementation, it's not going to be super quick because I'm working on this in fits and bursts here and there.

mattwynne:
👍 @wjlroe 

Feel free to ping me here on in gitter if you need some help finding your way around. There are some fairly gnarly bits of code you'll be getting into.

mattwynne:
How are you getting on @wjlroe? Need any help?

wjlroe:
@mattwynne I've just had a try at implementing this hook. It doesn't pass the tests, but you can see it is kinda working. I don't really understand what's going on though, so some pointers would be appreciated. The output from the AfterAll hooks doesn't get included in the output captured when the feature tests are run (but the puts code does execute), I don't know why that is. When I have a better understanding of all this, I'll add more specs, but right now this is exploratory code.

lib/cucumber/runtime.rb
```diff
@@ -67,6 +67,7 @@ def run!
       receiver = Test::Runner.new(@configuration.event_bus)
       compile features, receiver, filters
       @configuration.notify :test_run_finished
+      fire_after_all_hook
```
brasmusson: This should be moved up one line, so that the `test_run_finished` event means "all userland code that will be executed has been executed".

brasmusson:
I have argued [elsewhere](https://github.com/cucumber/cucumber-pretty-formatter/issues/2#issuecomment-209244351) that it should be possible by listening to the event stream exactly what userland code is execute (which would include `AfterAll` hooks). That means that we need new `xxStarted` and `xxFinished` events for these hooks that are not `TestSteps` (which `Before`, `After` and `AfterStep` hooks are), that are fired similar to the `TestStepStarted` and `TestStepFinished` events for [`TestSteps`](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/runner.rb#L26-L28)

When an `AfterAll` hook fails (an exception escapes it) I expect that the exit code in non-zero and the exceptions is captured by the formatters (by listening to the new `xxFinished` event).

mattwynne:
Good point Bjorn. I wonder if we could abstract the name 'test step' so that it could cover steps in scenarios, before / after scenario hooks, *and* these new after all hooks?

andrii-rymar:
@wjlroe do you have any plans to continue working on this?

wjlroe:
I don't know what to do, the discussion above doesn't mean anything to me, and I'm not likely to get any time to work on this, so no.

mattwynne:
Thanks for your efforts so far @wjlroe and sorry we lost you in the feedback discussion. I'm sure we can take it from here.

gwarjis:
Just wondering if there were any updates regarding an AfterAll hook?

alexttransisland:
is it still valid? I was able to use
```
AfterConfiguration do |config|
  config.on_event :test_run_finished do |event|
    config.out_stream.puts "test run finished"
  end
end
```

PR #1012, opened 2016-08-25
fix backtrace when using html formatter

denismartens:
## Summary

Fix an issue where the backtrace is repeated multiple times when using html formatter.

## Details

The method backtrace_line in html.rb was replacing each regex match with the entire backtrace, causing the backtrace to be repeated x times (x being the number of matches).

## Motivation and Context

Makes my cucumber output much shorter and cleaner for failing tests.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -590,11 +590,11 @@ module Cucumber
       end
 
       def backtrace_line(line)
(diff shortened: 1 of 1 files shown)


mattwynne:
Does this fix the same problem as #965?

denismartens:
Yes, it does. Sorry, I did not see that request when I made my change.

mattwynne:
Cool, thanks for making a contribution in any case @denismartens

PR #1018, opened 2016-09-13
Add first-timers paragraph

nodo:
Simply add a paragraph to give first-timers a few pointers.


mattwynne:
Nice! Do you think we should add an actual ticket (tagged Easy of course!) that points people to the rubocop violations?

nodo:
Done 👍

PR #1022, opened 2016-09-16
Fix Lint/AmbiguousOperator warnings #1021

### LINKED ISSUE #1021: Fix style violations
At the moment, the codebase has several style violations. It would be great to refactor the code, such that we are up to date with the community best practices https://github.com/bbatsov/ruby-style-guide.

The list of the violations is in two files:
- https://github.com/cucumber/cucumber-ruby/blob/master/.rubocop_todo.yml (auto generated)
- https://github.com/cucumber/cucumber-ruby/blob/master/.rubocop.yml

These files specify the ignored violations. It would be great to remove some of the violations from the files and fix them progressively in order to keep the PRs small.

hotovson:
partially fixes #1021


mattwynne:
So, this might we one we need to discuss @cucumber/cucumber-ruby 

@tooky and I, when pairing, have developed a convention that we use parentheses when we care about the return value, and not when we don't.

I'm struggling to find the reference that inspired that style, but [this article](https://heartmindcode.com/2013/09/05/parentheses-in-ruby/) is a reference.

I am fond of this style, but I'm more interested in consistency than pushing one particular style. If this parentheses-everywhere rules is easier to check with rubocop, I can live with it.

What are other people's thoughts?

nodo:
@mattwynne That is an interesting convention and I wouldn't mind to keep 👍 . 

If we decided to keep the convention, it would be great to ignore these violations in the `.rubocop.yml` file and add a paragraph in `CONTRIBUTING.md` that explains the code conventions of the project.

threedaymonk:
@mattwynne That's the convention I've long used myself, although, like you, I can't remember where I originally drew the inspiration. I suspect it flows naturally from Ruby's precedence rules:

```ruby
do_something transform(munge(a, b))
```

is easier to parse mentally than the equivalent but paren-less

```ruby
do_something transform munge a, b
```

tooky:
I think it's nice to use ruby's loose syntax for conventions - I remember Jim Weirich was a proponent of using `{}` blocks when you care about the return value and `do; end` when you don't. 

But if we are going to have conventions like that, I like @nodo's suggestion to document them. 

There are others we developed, like always returning self from commands. We should definitely write them down.

hotovson:
rebased to current master

nodo:
As suggested by @brasmusson  [in this comment](https://github.com/cucumber/cucumber-ruby/pull/1027#issuecomment-248895887) I changed the PR message to  "fix part of #1021" to inhibit Github to automatically close the related issue.

hotovson:
@nodo can you re-run Travis CI tests? it seems false negative

mattwynne:
I agree we should at least document this. I don't suppose there's a way to actually get the linter to check for that convention is there?

nodo:
@hotovson thanks for your work, I have restarted the CI tests and indeed it was a false negative.

@mattwynne unfortunately I don't think there is a easy way to get `rubocop` check some custom conventions. I will have a deeper look.

hotovson:
@nodo @mattwynne IMHO the only way is by [custom cops](http://rubocop.readthedocs.io/en/latest/extensions/#custom-cops)

mattwynne:
So what's the consensus with this one @cucumber/cucumber-ruby? I hate to waste @hotovson's good work, but it does seem as though a documented convention would work better in this instance, rather than a hard style check. Do we agree?

PR #1024, opened 2016-09-16
Fix rubocop violations

madundead:
## Summary

Partial fix for #1021 

## Details

* ```Hash#has_key?``` => ```Hash#key?``` due to [DEPRECATION](https://redmine.ruby-lang.org/issues/5555#note-10)
* Lint/BlockAlignment
* Lint/DefEndAlignment

### DIFF
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -85,7 +85,7 @@ module Cucumber
         )
 
         @builder << '<html xmlns ="http://www.w3.org/1999/xhtml">'
-          @builder.head do
+        @builder.head do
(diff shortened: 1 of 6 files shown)


lib/cucumber/formatter/html.rb
```diff
@@ -590,7 +590,7 @@ def format_exception(exception)
         (["#{exception.message}"] + exception.backtrace).join("\n")
       end
 
-     def backtrace_line(line)
+      def backtrace_line(line)
```
mattwynne: Is this bit of whitespace a mistake?

mattwynne: Oh no, I see - it was badly indented before!

mattwynne:
Awesome, thanks for your contribution @madundead!

PR #1036, opened 2016-09-24
fix Lint/UselessAccessModifier

hotovson:
## Summary

fix part of #1021

### DIFF
--- a/lib/cucumber/multiline_argument/data_table.rb
+++ b/lib/cucumber/multiline_argument/data_table.rb
@@ -494,11 +494,10 @@ module Cucumber
         raw.to_json(*args)
       end
 
-      private
-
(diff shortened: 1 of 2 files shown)


lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -494,11 +494,10 @@ def to_json(*args)
         raw.to_json(*args)
       end
 
-      private
-
       TO_S_PREFIXES = Hash.new('    ')
       TO_S_PREFIXES[:comment]   = '(+) '
       TO_S_PREFIXES[:undefined] = '(-) '
+      private_constant :TO_S_PREFIXES
```
nodo: Good stuff! :)

PR #1045, opened 2016-10-05
Fixes Style/IndentArray

pmatsinopoulos:
## Summary

Styling changes that have to do with array indentation. 
## Details

Rubocop complains if indentation of items of arrays are not as it expects them to be based on `Style/IndentArray` cop. So, I have fixed the indentation to be compatible with rubocop.
## Motivation and Context

This change is required if we want to comply with rubocop rules.

Refs #1021 
## How Has This Been Tested?

I have run all the tests with `bundle exec rake` and everything was green.
## Screenshots (if appropriate):

N/A
## Types of changes
- [x] Style Changes
## Checklist:
- [x] I have only changes the code style to be compatible with `Style/IndentArray` cop.


spec/cucumber/formatter/legacy_api/adapter_spec.rb
```diff
@@ -158,19 +158,19 @@ def after_hooks(source)
...
-                    :scenario_name,
-                  :after_feature_element,
-                :after_feature,
-              :after_features,
+                    :after_tags,
+                  :scenario_name,
+                :after_feature_element,
+              :after_feature,
+            :after_features,
           ]
```
nodo: Just wondering, is this indentation correct?

pmatsinopoulos: @nodo, the error that this PR fixes is related to this violation message:

```
spec/cucumber/formatter/legacy_api/adapter_spec.rb:162:15: C: Use 2 spaces for indentation in an array, relative to the start of the line where the left square bracket is.
              :before_features,
              ^^^^^^^^^^^^^^^^
```
So, I avoided doing any more changes other than the ones minimum needed to get rid of this violation 🙂  
Moving the indentation 2 spaces to the left (for the whole block) fixed that. Do you think that I should work on this with further enhancements? If yes, what enhancements would you suggest?

nodo: Hi @pmatsinopoulos, sorry for the late reply. Makes sense! It would be great to fix the indentation as well. I that all the elements of the array should be aligned with an indentation of 2 spaces. What do you think?

nodo: Absolutely :) thanks @pmatsinopoulos !

mattwynne: Yes, this indentation is there to make the tests easier to read / write, since you need matching pairs of those messages. I'd be happy to see it disappear though if it's a pain - I hope we'll be able to [remove this entire file soon](https://github.com/cucumber/cucumber-ruby/issues/839)

mattwynne: Certainly, moving the indentation two spaces to the left would be fine!

jaysonesmith:
Does this need help with the conflict to get it completed?

jaysonesmith:
Merge needs some work before it's all set. Sorry for the early push!

spec/cucumber/formatter/legacy_api/adapter_spec.rb
```diff
@@ -991,7 +992,7 @@ def after_hooks(source)
 
           # rubocop:disable AlignArray
           expect( formatter.legacy_messages ).to eq [
-              :before_features,
+            :before_features,
```
olleolleolle: Perhaps someone already noted this, but all the places where the rule AlignArray has been explicitly disabled, those places in the code are carefully arranged ASCII art of a sort. 

The code indentation there explains levels of nesting and ordering. 

Could you make those align in a harmonious way in this change? I mean: Avoid four-space indents.

jaysonesmith: I certainly can. Are you concerned with this even if all tests are passing? There are other instances of this are all on the same indent and they work as well. 

Matt's input on this is that the tabs are more of a guideline for visibility but aren't actually required. 

Let me know your thoughts. :)

olleolleolle: I care enough about the previous effort of “sketching the structure” using indents that I would like for it to be kept. 

This is a nonfunctional design detail but a quite good one. It communicated things to me, at least.

brasmusson: I would like to point out that the legacy_api module will be deleted as soon as possible (we just have to rewrite the pretty and html formatter first, see https://github.com/cucumber/cucumber-ruby/issues/839), so any work in the legacy_api or its spec it a waste of time. I would recommend to remove all changes in the legacy_api module and its spec from this PR (and any other similar PR).

jaysonesmith: Should I do a wholesale exclude on the legacy api file?

olleolleolle:
That’s a zen way of not having this problem. I like this resolution for this.

spec/cucumber/formatter/legacy_api/adapter_spec.rb
jaysonesmith: @olleolleolle fair enough! I'm afk for a bit but I'll get to this when I can. Thanks for the input guys! (You too @brasmusson !)

jaysonesmith:
Update made(pulled from commit message):
- overwrite file with master version
- remove inline rubocop disable/enable
- exclude adapter_spec file from rubocop.yml file to prevent any other style changes being performed on this file.

PR #1049, opened 2016-10-19
move --retry to latest relase

phoet:
i just installed 2.4.0 and the `--retry` option is available there. the history still lists it as "in git" while it's part of the commits 2.3.0..2.4.0


brasmusson:
Yes, there are changes related to `--retry` in v2.4.0, but those are not enough to make `--retry` actually work (nothing happens if you use `--retry`), it did not work until also #992 was done - therefore I deliberately moved `--retry` out of v2.4.0 in the history.

phoet:
Makes sense, thx for the clarification

Issue #1003, opened 2016-08-08
Add support for Tag Expressions


aslakhellesoy:
## Summary

[Tag Expressions](http://docs.cucumber.io/tag-expressions/) is a more intuitive way to specify tags in tagged hooks and on the command line's `--tags` argument.
## Expected Behavior

Tagged Hooks should accept a single string:

``` ruby
Before("@foo and not @bar") do
end
```

Likewise, the `--tags` command line argument should allow the same. (Specifying it more than once should cause an error)

```
--tags "@foo and not @bar"
```
## Context & Motivation

The current mechanism to specify `or` with comma and `and` with multiple `--tags` is incredibly awkward. Tag Expressions are intuitive!

nodo:
@aslakhellesoy I am quite new to the project, but I would be happy to give this feature a try. What do you think?

aslakhellesoy:
That would be fantastic! Here is how I would do it:

First, fork the cucumber/cucumber repo and add a ruby implementation for tag expressions, porting the JavaScript or Java code. When that's done, send a PR and I'll create a mirror of it at https://github.com/cucumber/tag-expressions-ruby

Then, it's time to update Cucumber-Ruby to use it:

* First make sure you can run all tests successfully
* Locate all the old tag expression code and delete it
* Update existing tests
* Make them pass by using the new library

How does that sound?

nodo:
Sounds good, thanks a lot @aslakhellesoy :) I am planning to dedicate some times next week and keep you updated here.

aslakhellesoy:
That's great @nodo - let us know if you have any questions. You can also try the [contributors](https://gitter.im/cucumber/contributors) chat room.

mattwynne:
If someone wants to take a crack at integrating @nodo's work in https://github.com/cucumber/cucumber/issues/67, I'd like to ask that they help us to stay backwards-compatible for one major release by making the choice of old / new tag expression syntax a configuration option. That will enable us to transition users over to the new syntax gradually. We can make it opt-in first, then make it the default for the next major release.

aslakhellesoy:
Great work so far @nodo! I have released [cucumber-tag_expressions 1.0.0](https://rubygems.org/gems/cucumber-tag_expressions)!

aslakhellesoy:
I second @mattwynne's comment about backwards compatibility.

In the next version, use the new `tag-expression` library only if:

[...]

PR #1054, opened 2016-11-13
Use tag expressions from the cucumber-tag_expressions gem.

brasmusson:
## Summary

Support [standard library tag expressions](https://github.com/cucumber/cucumber/tree/master/tag-expressions) from the cucumber-tag_expressions gem.

## Details

The tag limits from the `--tags` options are stripped of and handled separately, so that the stored tag expressions can be passed the tag expressions library directly.

Old style tag expressions are deprecated but still handled correctly.

Requires https://github.com/cucumber/cucumber-ruby-core/pull/116

Closes https://github.com/cucumber/cucumber-ruby/issues/1003

## How Has This Been Tested?

Features and spec are updated accordingly.

## Types of changes

## Checklist:

- [X] I've updated the help test (`-h` option)

### DIFF
--- a/cucumber.yml
+++ b/cucumber.yml
@@ -3,18 +3,19 @@ cucumber_pro_opts = ENV['ENABLE_CUCUMBER_PRO'] ? "--format Cucumber::Pro --out /
 rerun = File.file?('rerun.txt') ? IO.read('rerun.txt') : ""
 rerun_opts = rerun.to_s.strip.empty? ? "--format progress features" : "--format pretty #{rerun}"
 std_opts = "--format progress features -r features --strict #{cucumber_pro_opts}"
-std_opts << " --tags ~@wip-jruby" if defined?(JRUBY_VERSION)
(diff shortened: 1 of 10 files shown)


mattwynne:
@brasmusson you left "i have updated the documentation unticked" but you've changed https://github.com/cucumber/cucumber-ruby/pull/1054/files#diff-f622824b18af7ee8c0203bd93e1c7d7d - I think that counts!

mattwynne:
Thanks for making it backwards compatible but adding deprecation warnings. Perfect!

brasmusson:
Before removing the backward compatibility (in Cucumber-Ruby v4?) Aruba needs to be updated to not use [`~` as "not" in tagged hooks](https://github.com/cucumber/aruba/blob/master/lib/aruba/cucumber/hooks.rb#L24) (I suppressed the [warning for it](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/rb_support/rb_hook.rb#L24) - since the user seeing the warning cannot just change the Aruba code).

PR #1058, opened 2016-11-26
Add per-formatter options

nodo:
## Summary

Add per-formatter options passed via CLI.  More details regarding this feature can be found at this PR: https://github.com/cucumber/cucumber-ruby/pull/954.

E.g.
`bundle exec cucumber example.feature --format PersonalModule::MyFormatter,a=1,b=2`

Will pass the options `{ 'a' => 1, 'b' => 2 }` to the formatter.

## Types of changes

## Checklist:


aslakhellesoy:
I know @charlierudolph has an opinion about this and how it's done in cucumber-js.

Let's agree on a consistent mechanism for all impls.

nodo:
I am not sure if `cucumber-js` allow per-formatter options (according to https://github.com/cucumber/cucumber-ruby/pull/954#issuecomment-249248085). However, I agree, it would be better to be consistent across implementations.

IMO this syntax is quite nice for per-formatter options.

mattwynne:
I like this syntax too. I'd like to keep consistency too, but I don't want to drag down the progress of one implementation to wait for consensus. We could always change it in a future release.

I suggest we merge this, and write a PR on cucumber/cucumber where we can discuss formalising it for all other implementations. If something different comes out of that discussion we can change Ruby to match. WDYT @aslakhellesoy?

aslakhellesoy:
cucumber.js master (upcoming 2.0) has `--format-options <JSON>`: https://github.com/cucumber/cucumber-js/blob/735e42656031b1192c8c4444599a7fa23d5d9b58/src/cli/argv_parser.js#L38

I think cucumber-ruby should do the same.

The principle here is that whatever implementation is first to implement a feature decides how that feature is implemented. Other implementations should follow this by default, unless there are good reason to do it differently.

WDYT?

nodo:
@aslakhellesoy I believe that `--format-options` is about passing options to *all* formatters. Your point would be to avoid passing options to a single formatter? Or add both use cases?

aslakhellesoy:
I'm not sure, @nodo - the `--help` docs do say "repeatable", so maybe it's a positional option that applies to the previous `--format`? I haven't read the code yet.

mattwynne:
@aslakhellesoy I think a better protocol is to let individual implementations experiment with different stuff and learn from those experiments, and to build consensus in PRs at cucumber/cucumber. Charlie's already told me he doesn't particularly like the JSON syntax, so I don't see the point of reworking @nodo's implementation here, which might turn out to be a more popular solution in the end.

Trending *towards* consistency is the key here IMO - we shouldn't let things fly off in different directions for too long, but I think we'll just slow the project down and frustrate people if we try to have too many rules about this.

aslakhellesoy:
I think we're in agreement here. I said:

> Other implementations should follow this by default, *unless there are good reason to do it differently*.

You've just pointed out a good reason to do it differently, @matt: @charlierudolph has told you he's not crazy about the JSON syntax.

So @nodo go ahead with what you proposed!

nodo:
Thanks @aslakhellesoy @mattwynne. From my side, this PR is ready for review. Let me know if you have any suggestions. Adding `for-review` label.

features/docs/extending_cucumber/custom_formatter.feature
```diff
@@ -41,7 +41,27 @@ Feature: Custom Formatter
...
+            end
+          end
+        end
+      end
+      """
+    When I run `cucumber features/f.feature --format MyCustom::Formatter,foo=bar,one=two`
+    Then it should pass with exactly:
+    """
+    {"foo"=>"bar", "one"=>"two"}
+    """
```
mattwynne: Nice acceptance test!

mattwynne:
So the next thing we can do with this, is to make the JUnit formatter take its own option, rather than using the --out parameter. I think that's the only one that uses it as a path, so in that same PR we can then rename `path_or_io` everywhere or just get rid of it and use `config.out_stream` instead.

mattwynne:
@nodo sorry it took me so long to pay this full attention - I've been crazy busy. Thanks a million for your contributions - they're a pleasure to read.

nodo:
Thanks @mattwynne, appreciated!

Issue #1041, opened 2016-09-30
Step defs formatter should report all available step defs even if they haven't been matched


tooky:
## Summary

The `stepdefs` formatter should report all step definitions available in the library of step defs, including steps from included libraries like [aruba](https://github.com/cucumber/aruba). The current behaviour is to only print step defs in the current project, or matched from included step libraries

Consider a project set up like this:

`features/support/aruba.rb`

``` ruby
require 'aruba/cucumber'
```

`features/step_definitions/steps.rb`

``` ruby
Given(/^a step$/) do
end
```
## Expected Behavior

Running `cucumber --format stepdefs`

```


0.0000000 /^I run "(.*)"$/                                                                                     # aruba-0.14.2/lib/aruba/cucumber/command.rb:6
  NOT MATCHED BY ANY STEPS
0.0000000 /^I'm using a clean gemset "([^"]*)"$/                                                               # aruba-0.14.2/lib/aruba/cucumber/rvm.rb:1
  NOT MATCHED BY ANY STEPS
0.0000000 /^the test(?:s)? should( not)?(?: all)? pass with( regex)?:$/                                        # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:81
  NOT MATCHED BY ANY STEPS
0.0000000 /^the tests(?:s)? should( not)?(?: all)? pass(?: with (\d+) failures?)?$/                            # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:65
  NOT MATCHED BY ANY STEPS
0.0000000 /^the spec(?:s)? should( not)?(?: all)? pass with( regex)?:$/                                        # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:48
  NOT MATCHED BY ANY STEPS
0.0000000 /^the spec(?:s)? should( not)?(?: all)? pass(?: with (\d+) failures?)?$/                             # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:32
  NOT MATCHED BY ANY STEPS
0.0000000 /^the feature(?:s)? should( not)?(?: all)? pass with( regex)?:$/                                     # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:14
  NOT MATCHED BY ANY STEPS
0.0000000 /^the feature(?:s)? should( not)?(?: all)? pass$/                                                    # aruba-0.14.2/lib/aruba/cucumber/testing_frameworks.rb:2
[...]

PR #1059, opened 2016-11-27
Start the event bus after all listeners are registered

brasmusson:
## Summary

To make sure that all listeners will receive all events broadcasted through the event bus, start the event bus only after all listeners have had a change so register themselves with the event bus.

One case where this is important is so that the usage formatter get to know all step definitions that are registered, even though the step definitions are registered before the formatters have been created.

## Details

Revert changes in file loading from #977, so that the support and step definition files again are loaded at the same time (before the formatters has been created).

Call `EventBus#start` after all formatters and other event bus listeners have been created and had a change to register themselves with the event bus.

Requires https://github.com/cucumber/cucumber-ruby-core/pull/117.

## Motivation and Context

Fixes https://github.com/cucumber/cucumber-ruby/issues/1041.

## How Has This Been Tested?

The current `usage_formatter.feature` verifies that the step definitions loaded before the usage/stepdefs formatter has registered itself with the event bus are reported by the usage/stepdefs formatter.

## Types of changes

## Checklist:

- [X] I've added tests for my code (the current features verify the correctness of the change).

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -122,7 +122,7 @@ Metrics/MethodLength:
 # Offense count: 19
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 1976
+  Max: 1977
 
 # Offense count: 7
 # Configuration parameters: CountKeywordArgs.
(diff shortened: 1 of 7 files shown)


brasmusson:
Superseded by https://github.com/cucumber/cucumber-ruby-core/pull/136.

PR #1073, opened 2017-01-20
Aruba 0.14.x support

voxik:
## Summary

This fixes compatibility with latest Aruba.

## Details

Aruba changed its implementation a bit and therefore cucumber test suite needs to be modified. This does not include any modifications of Gemfile.

## Motivation and Context

Keep with upstream.

## How Has This Been Tested?

I am going to apply this patch to be able to properly test Cucumber package on Fedora, where there is already latest Aruba.

## Types of changes

## Checklist:

### DIFF
--- a/features/docs/work_in_progress.feature
+++ b/features/docs/work_in_progress.feature
@@ -124,7 +124,6 @@ Feature: Cucumber --work-in-progress switch
 
       features/wip.feature:15:in `Scenario: Passing'
 
-
       """
 
   Scenario: Fail with Passing Scenario Outline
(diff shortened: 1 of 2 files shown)


voxik:
Also please note that there are still some deprecation warnings etc, but they can be fixed later ...

voxik:
The #all_stdout could be removed ....

mattwynne:
Thanks @voxik! Please go ahead and send us more PRs if you can see other ways to clean up how we use Aruba. Aruba was originally extracted from Cucumber's features a very long time ago, and so we might be using quite a out of date API!

PR #1074, opened 2017-01-30
Run rubygems update before installing

nodo:
Travis build is failing due to this bug: https://github.com/sickill/rainbow/issues/49.

Running `gem update --system` before running `bundle install` fixes the problem (see https://github.com/sickill/rainbow/issues/49#issuecomment-274714272 and https://github.com/bundler/bundler/issues/5357#issuecomment-274742137)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,12 @@ matrix:
     - rvm: 2.2
     - rvm: 2.1
     - rvm: jruby-9.1.2.0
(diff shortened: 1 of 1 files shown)


mattwynne:
@nodo thanks for doing this in a PR so it's nicely visible, but do feel free to merge your own PRs for small but valuable things like this rather than get blocked.

Issue #1071, opened 2017-01-20
features/docs/getting_started.feature:7 fails on Ruby 2.4.


voxik:
## Summary

The test case features/docs/getting_started.feature:7 fails on Ruby 2.4:

## Expected Behavior

The test should pass.

## Current Behavior

~~~
expected "No such file or directory @ rb_sysopen - features. You can use `cucumber --init` to get started.\n" to include "No such file or directory - features. You can use `cucumber --init` to get started."
Diff:
@@ -1,2 +1,2 @@
-No such file or directory - features. You can use `cucumber --init` to get started.
+No such file or directory @ rb_sysopen - features. You can use `cucumber --init` to get started.
 (RSpec::Expectations::ExpectationNotMetError)
features/docs/getting_started.feature:10:in `Then it should fail with:'

Failing Scenarios:
cucumber features/docs/getting_started.feature:7 # Scenario: Run Cucumber in an empty directory
~~~

## Possible Solution

This obviously fails due to  https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L34 . Isn't it about the time to drop the https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L15 ? All currently supported Ruby prints this and I don't think it is going to change soon.

voxik:
Other option is to include it unconditionally ....

mattwynne:
Thanks for looking into this @voxik!

I'd definitely prefer if we can avoid showing the `@ rb_sysopen` noise to our users.

I think it would be nicer to stop subclassing the `FileException` and just raise our own error with our own message, and get rid of all of that cruft. Would you like to send a PR for that?

PR #1069, opened 2017-01-09
Fix an json compile error for "bundle install" on Ruby 2.4.

junaruga:
## Summary

Fix an json compile error for "bundle install" on Ruby 2.4.
Installed json-1.8.3 does not have a compatibility for Ruby 2.4.

## Details

```
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-linux]
```

```
$ bundle -v
Bundler version 1.13.7
```

On lastest master branch (commit hash: 8871da436987e2c46623a24909a925818be52314)

```
$ pushd ../cucumber-ruby-core
$ git checkout v2.0.0
$ popd
```

```
$ pushd ../cucumber-ruby-wire
$ git checkout v0.0.1
$ popd
```

```
$ bundle install --path vendor/bundle
...
current directory:
/home/jaruga/git/cucumber-ruby/vendor/bundle/ruby/2.4.0/gems/json-1.8.3/ext/json/ext/generator
/usr/local/ruby-2.4.0/bin/ruby -r ./siteconf20170109-6691-15ef87j.rb extconf.rb
creating Makefile

current directory:
/home/jaruga/git/cucumber-ruby/vendor/bundle/ruby/2.4.0/gems/json-1.8.3/ext/json/ext/generator 
make "DESTDIR=" clean

current directory:
/home/jaruga/git/cucumber-ruby/vendor/bundle/ruby/2.4.0/gems/json-1.8.3/ext/json/ext/generator
make "DESTDIR="
compiling generator.c
generator.c: In function ‘generate_json’:
generator.c:861:25: error: ‘rb_cFixnum’ undeclared (first use in this function)
     } else if (klass == rb_cFixnum) {
                         ^~~~~~~~~~
generator.c:861:25: note: each undeclared identifier is reported only once for each
function it appears in
generator.c:863:25: error: ‘rb_cBignum’ undeclared (first use in this function)
     } else if (klass == rb_cBignum) {
                         ^~~~~~~~~~
generator.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-self-assign’
cc1: warning: unrecognized command line option ‘-Wno-constant-logical-operand’
cc1: warning: unrecognized command line option ‘-Wno-parentheses-equality’
Makefile:241: recipe for target 'generator.o' failed
make: *** [generator.o] Error 1
...
```

## Motivation and Context

We may want to run cucumber on Ruby 2.4.

## Screenshots (if appropriate):

I ran below command, and done successfully.

```
[...]

PR #1077, opened 2017-02-08
Fix tests for Ruby 2.4.0

junaruga:
## Summary

Fixes tests for Ruby 2.4.0.
This fixes https://github.com/cucumber/cucumber-ruby/issues/1071

I referred https://github.com/cucumber/cucumber-ruby/issues/1071#issuecomment-276709766 about how to fix

## Motivation and Context

Want to pass tests on Ruby 2.4.0.

## How Has This Been Tested?

Same way with https://github.com/cucumber/cucumber-ruby/pull/1069#issuecomment-278122402

```
$ bundle exec rake spec
...
568 examples, 0 failures

$ bundle exec rake cucumber
...
207 scenarios (207 passed)
1217 steps (1217 passed)
0m25.223s
```

## Types of changes

- [] New feature (non-breaking change which adds functionality)

  I changed from `FeatureFolderNotFoundException(e, path)` to `FeatureFolderNotFoundException(path)`

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,10 @@ before_install:
   - gem update --system
   - gem update bundler
 
+before_script:
+  # Patch for Ruby 2.4.0 compatibility.
+  - ./scripts/patch_simplecov_0_12_0
+
 notifications:
   webhooks:
     urls: # gitter
--- a/scripts/patch_simplecov_0_12_0
+++ b/scripts/patch_simplecov_0_12_0
@@ -1,0 +1,9 @@
+#!/bin/bash
+# This script modifies simplecov-0.12.0 for Ruby 2.4.0 compatibility.
+# https://github.com/colszowka/simplecov/commit/050eb68
+
+set -ev
+
+SIMPLECOV_ROOT=$(bundle show simplecov)
+sed -i 's/Fixnum/Integer/g' "${SIMPLECOV_ROOT}/lib/simplecov/configuration.rb"
+sed -i 's/Fixnum/Integer/g' "${SIMPLECOV_ROOT}/lib/simplecov/source_file.rb"


junaruga:
Finally we passed the tests for Ruby 2.4.0.
About a script for Travis, I referred this page: https://docs.travis-ci.com/user/customizing-the-build/

.travis.yml
```diff
@@ -27,6 +28,10 @@ before_install:
   - gem update --system
   - gem update bundler
 
+before_script:
+  # Patch for Ruby 2.4.0 compatibility.
+  - ./scripts/patch_simplecov_0_12_0
+
```
nodo: I think it would be better to upgrade to `simplecov 0.13`. This issue should be fixed: https://github.com/colszowka/simplecov/blob/0.13.0/CHANGELOG.md

lib/cucumber/runtime.rb
```diff
@@ -30,8 +30,12 @@ def initialize(original_exception, path)
...
 
-  class FeatureFolderNotFoundException < FileException
-    include FixRuby21Bug9285 if Cucumber::RUBY_2_1 || Cucumber::RUBY_2_2 || Cucumber::RUBY_2_3
+  class FeatureFolderNotFoundException < Exception
+    attr :path
+    def initialize(path)
+      super('No such file or directory - features')
+      @path = path
+    end
   end
```
nodo: I am not sure why this change is needed. Please, can you tell which test is failing for this?

.travis.yml
junaruga: @nodo yes of course using `simplecov 0.13` is better if we can use it on Travis.
But right now we can not use it on Travis.
Because we are using latest version `coveralls 0.8.19` in our `cucumber-ruby`.
And it needs `simplecov` 0.12.x.

See `Gemfile.lock` file after `bundle install --path vendor/bundle` in your local environment.

Gemfile.lock
```
...
    coveralls (0.8.19)
      json (>= 1.8, < 3)
      simplecov (~> 0.12.0)
...
```

About this situation, I have reported to `coveralls` upstream github.
So, they may change their dependency to use `simplecov 0.13`.

So the script `./scripts/patch_simplecov_0_12_0` is temporary one to pass the test for Ruby 2.4.0.
After they may update `coveralls`, we can remove this script.

lib/cucumber/runtime.rb
junaruga: @nodo because of fixing this issue https://github.com/cucumber/cucumber-ruby/issues/1071#issuecomment-276709766 (Reason of the issue is https://bugs.ruby-lang.org/issues/9285)

There are 2 possible ways to to fix this issue. And @mattwynne prefer "1".
1. Adding `Cucumber::RUBY_2_3` in above logic.
2. Not extend `FileException`

.travis.yml
junaruga: As a reference, running `./scripts/patch_simplecov_0_12_0` is same with patching https://github.com/colszowka/simplecov/commit/050eb68

lib/cucumber/runtime.rb
junaruga: @nodo the failing test is `features/docs/getting_started.feature:7 # Scenario: Run Cucumber in an empty directory` (#1071)

.travis.yml
junaruga: See also https://github.com/lemurheavy/coveralls-ruby/blob/v0.8.19/coveralls-ruby.gemspec#L23
If `coveralls` new version with `simplecov` 0.13 will be released, we do not need this patch for `simplecov`.
It's also good to watch `coveralls` project for that.

junaruga:
I think my solution for failure tests by simplecov warnings, was not good.
I would close this PR.
Thanks for taking time for this PR.

PR #1079, opened 2017-02-14
Relax filter for "/usr/local/ruby-X.Y.Z/bin/bundle".

junaruga:
## Summary

This fixes tests with /usr/local/ruby-X.Y.Z/bin/bundle.
The solution is mentioned by https://github.com/cucumber/cucumber-ruby/pull/1069#issuecomment-278132114
Can you check this PR before #1077 ?

## Motivation and Context

Want to pass tests.

## How Has This Been Tested?

There is still an error. but it is another reason.

```
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-linux]

$ bundle -v
Bundler version 1.14.3

$ bundle exec rake spec
...
568 examples, 0 failures

$ bundle exec rake cucumber
...
(::) failed steps (::)

expected "No such file or directory @ rb_sysopen - features. You can use `cucumber --init` to get started.\n" to include "No such file or directory - features. You can use `cucumber --init` to get started."
Diff:
@@ -1,2 +1,2 @@
-No such file or directory - features. You can use `cucumber --init` to get started.
+No such file or directory @ rb_sysopen - features. You can use `cucumber --init` to get started.
 (RSpec::Expectations::ExpectationNotMetError)
features/docs/getting_started.feature:10:in `Then it should fail with:'

Failing Scenarios:
cucumber features/docs/getting_started.feature:7 # Scenario: Run Cucumber in an empty directory

207 scenarios (1 failed, 206 passed)
1217 steps (1 failed, 1216 passed)
```

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1080, opened 2017-02-15
Avoid showing "@ rb_sysopen" noise for Ruby 2.4.

### LINKED ISSUE #1071: features/docs/getting_started.feature:7 fails on Ruby 2.4.

junaruga:
## Summary

This fixes #1071.

## Details

This issue happens for only Ruby 2.4.0.

## Motivation and Context

Want to pass tests.

## How Has This Been Tested?

```
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-linux]

$ bundle exec rake spec
...
568 examples, 0 failures

$ bundle exec rake cucumber
...
207 scenarios (17 failed, 190 passed)
1217 steps (17 failed, 1200 passed)
```

There are still failed tests. But they will be fixed by https://github.com/cucumber/cucumber-ruby/pull/1079 .
The `rb_sysopen` message was disappeared.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -30,8 +30,14 @@ module Cucumber
   class FileNotFoundException < FileException
   end
 
-  class FeatureFolderNotFoundException < FileException
-    include FixRuby21Bug9285 if Cucumber::RUBY_2_1 || Cucumber::RUBY_2_2 || Cucumber::RUBY_2_3
+  class FeatureFolderNotFoundException < Exception
+    def initialize(path)
+      @path = path
+    end
+
+    def message
+      "No such file or directory - #{@path}"
+    end
   end
 
   require 'cucumber/core'
@@ -141,8 +147,8 @@ module Cucumber
           set_encoding
         rescue Errno::EACCES => e
           raise FileNotFoundException.new(e, File.expand_path(path))
-        rescue Errno::ENOENT => e
-          raise FeatureFolderNotFoundException.new(e, path)
+        rescue Errno::ENOENT
+          raise FeatureFolderNotFoundException.new(path)
         end
       end
 


lib/cucumber/runtime.rb
```diff
@@ -30,8 +30,12 @@ def initialize(original_exception, path)
   class FileNotFoundException < FileException
   end
 
-  class FeatureFolderNotFoundException < FileException
-    include FixRuby21Bug9285 if Cucumber::RUBY_2_1 || Cucumber::RUBY_2_2 || Cucumber::RUBY_2_3
+  class FeatureFolderNotFoundException < Exception
+    attr :path
```
nodo: Do we need a getter for the `path` variable here?

lib/cucumber/runtime.rb
```diff
@@ -30,8 +30,12 @@ def initialize(original_exception, path)
...
+    def initialize(path)
```
nodo: Do we need a getter for `path` variable here?

lib/cucumber/runtime.rb
```diff
@@ -30,8 +30,12 @@ def initialize(original_exception, path)
...
+      super('No such file or directory - features')
+      @path = path
```
nodo: Instead of calling `super`, I would suggest to override `message` method and output the path related to the error. E.g. `"Feature folder not found at #{@path}"` or similar. If we follow this path I would also use `File.expand_path(path)` for clarity.

lib/cucumber/platform.rb
```diff
@@ -14,6 +14,7 @@ module Cucumber
     RUBY_2_2      = RUBY_VERSION =~ /^2\.2/
     RUBY_2_1      = RUBY_VERSION =~ /^2\.1/
     RUBY_2_3      = RUBY_VERSION =~ /^2\.3/
+    RUBY_2_4      = RUBY_VERSION =~ /^2\.4/
```
nodo: Do we need these constants at all? /cc @mattwynne @brasmusson

nodo:
Thanks @junaruga - just added a couple of comments.

junaruga:
@nodo ok, let me check your coments later. Thanks.

lib/cucumber/platform.rb
mattwynne: I think this PR removes the only usage, so I'd say we can remove them.

nodo: Thanks @mattwynne 👍

junaruga:
@nodo wait a moment. I will check it today or tomorrow as right now I am busy.

nodo:
@junaruga no problem at all. Thanks for working on this!

lib/cucumber/runtime.rb
junaruga: @nodo I thought previously FeatureFolderNotFoundException had the getter for `path` variable, because it had extended `FileException`. But as we do not use the getter right now, removing it might be better. I am going to remove it.

lib/cucumber/runtime.rb
junaruga: @nodo ok I will follow your way.

lib/cucumber/platform.rb
junaruga: @nodo ok I will remove them 👍

junaruga: > this PR removes the only usage,

I am still not sure above point. actually this PR can remove these constants?
or this PR should not touch these constants?

The constants RUBY_2_* are used in `gem_tasks/cucumber.rake`.

```
$ grep -r RUBY_2_ *
gem_tasks/cucumber.rake:                   elsif Cucumber::RUBY_2_0
gem_tasks/cucumber.rake:                   elsif Cucumber::RUBY_2_1
lib/cucumber/runtime.rb:    include FixRuby21Bug9285 if Cucumber::RUBY_2_1 || Cucumber::RUBY_2_2 || Cucumber::RUBY_2_3
lib/cucumber/platform.rb:    RUBY_2_2      = RUBY_VERSION =~ /^2\.2/
lib/cucumber/platform.rb:    RUBY_2_1      = RUBY_VERSION =~ /^2\.1/
lib/cucumber/platform.rb:    RUBY_2_3      = RUBY_VERSION =~ /^2\.3/
```

junaruga:
@nodo I pushed one more commit for review. you can do "merge with squash" from Github GUI, when you will merge it. You can check the result on Travis too.

features/docs/getting_started.feature
```diff
@@ -9,7 +9,7 @@ Feature: Getting started
     When I run `cucumber`
     Then it should fail with:
       """
-      No such file or directory - features. You can use `cucumber --init` to get started.
+      Feature folder not found at
       """
```
nodo: I think this message should be more detailed. Some possible solutions:
1. Use `Aruba.config.root_directory` and somehow use it in the string
2. Change the exception to report a relative path and not a absolute path (so that here we can write `tmp/features`
3. Change the exception to not report the path

Sorry for this, I haven't notice this test before. However, I think it's important to keep it as informative as possible.

What do you think?

junaruga: @nodo I like kind of 2. "kind of 2" means when I removed "File.expand_path", I did not need to add `tmp/features' to features/docs/getting_started.feature. Check my additional commit, I am going to push it right now.

I prefer `File.expand_path`.
But I like you would add `File.expand_path` by yourself later, if you need it.
Because right now I want to pass all the tests on Ruby 2.4.0.

voxik:
The error message "Feature folder not found at features.", which is default output with this patch applied, is more confusing then helpful IMO. It is hardly obvious that the "features" refers to folder. If there was the dash as there used to be, that would be a bit better probably ....

junaruga:
@nodo I changed and did rebase of my code. I do not change the message by myself on this PR.
I did not make it worse from current state for this PR. Maybe.

I like you will merge this, and ask you to improve the message by yourself after this PR as you like.

junaruga:
@nodo I can tell you other options too.

- You can send your additional code to my repo https://github.com/junaruga/cucumber-ruby branch: `feature/fix-rb_sysopen-on-ruby-2.4`. I am going to accept your any code.
- You can close this pull-request. and you can fix it by yourself.

It does not matter for me. You can make a choice.

nodo:
@junaruga sorry for not being responsive. I will check this PR as soon as possible. Thanks for your patience and for your work. Your changes looks good to me, I will just check the message as suggested by @voxik.

junaruga:
@nodo Thank you!!

PR #1082, opened 2017-02-15
Add `test_run_starting` event to broadcast all test cases before a run

mattwynne:
Any listener that wants to know how many test cases there will be (e.g. to display a progress bar) will need this information.


mattwynne:
Part of #962

brasmusson:
Is it really a good idea that because one listener needs to know the number test cases, also when that listener is not used, the first test case cannot be executed before all feature files have been parsed and compiled? Isn't this a major change from the original design of "the new core". Shouldn't this be subject of a command line option?

lib/cucumber/events.rb
```diff
@@ -32,6 +32,7 @@ def self.registry
         StepActivated,
         TestRunFinished,
         GherkinSourceRead
+        TestRunStarting,
```
olleolleolle: The comma seems to be misplaced?

brasmusson:
Also I think that an event named "TestRunStarting" should be issued first before any feature files has been read, and that the time printed in the summary should be the time difference between the time stamp of the TestRunStaring event and the time stamp of the TestRunFinished event.

mattwynne:
*now* you tell me 😜

@brassmusson good point. maybe we could be clever enough to only add this filter if there is anyone listening to the event?

lib/cucumber/events.rb
olleolleolle: https://github.com/cucumber/cucumber-ruby/pull/1124

PR #1048, opened 2016-10-07
Fixes for rubocop violations

phoebeclarke:
## Summary

Multiple fixes for rubocop violations: 
- Style/WordArray
- Style/ParallelAssignment
- Style/MultilineOperationIndentation
- Style/MultilineMethodCallIndentation
- Style/LeadingCommentSpace
- Style/IndentArray
- Style/HashSyntax

References #1021

## Details

Fixes for multiple violations as detected with rubocop. The violations are fixed and the test cases removed.
## Motivation and Context

Changes required to make code compliant with the Ruby style guide.
## How Has This Been Tested?

Ran `bundle exec rake` and no issues were reported. Builds on Travis.
## Screenshots (if appropriate):

N/A
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Style change
## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1084, opened 2017-02-19
1048 second attempt

phoebeclarke:
## Summary

References #1048

A clean-up of the previously attempted pull request; commits have been cherry-picked and merge conflicts fixed accordingly. All commits are in regards to fixing style violations.

Let me know if there's anything else I need to do!

## How Has This Been Tested?

`bundle exec rake` was performed and all checks passed.

## Types of changes

- [x] Style fixes

## Checklist:


mattwynne:
Awesome, thanks @phoebeclarke. I'm just about to board a flight but I'll take a look ASAP, unless someone beats me to it :)

spec/cucumber/step_match_search_spec.rb
```diff
@@ -111,7 +111,7 @@ module Cucumber
         end
       end
 
-      # TODO: remove this - it's ... redundant
+      #  TODO: remove this - it's ... redundant
       # http://railsforum.com/viewtopic.php?pid=93881
```
nodo: Please, can you make sure that there is only one space before the `TODO` word? 

The misleading bit is that as first character there is a NON-BREAK SPACE utf-8 character (https://en.wikipedia.org/wiki/Non-breaking_space), which should be removed. Please can you fix also the lines above, there are many with this problem.

nodo:
Thanks for your work @phoebeclarke ! Really appreciate your effort.  Please can you check my comment? There are some problems with `Style/LeadingCommentSpace violations` fixes, unfortunately there are some utf-8 chars around that should be removed as part of the style fix.

@mattwynne feel free to add your review 😄

lib/cucumber/events/step_activated.rb
```diff
@@ -7,19 +7,20 @@ module Events
...
-      # Information about the matching definition.
+      # Information about the matching definition.
       #
       # @return [Cucumber::StepMatch]
       attr_reader :step_match
 
-      # @private
+      # @private
       def initialize(test_step, step_match)
-        @test_step, @step_match = test_step, step_match
```
mattwynne: Which rule made us change this? I like this idiom a lot :) Especially in constructors.

What do other people think?

lib/cucumber/events/step_definition_registered.rb
```diff
@@ -4,16 +4,16 @@
 module Cucumber
   module Events
 
-    # Event fired after each step definition has been registered
+    # Event fired after each step definition has been registered
     class StepDefinitionRegistered < Core::Event.new(:step_definition)
 
 
-      #_The step definition that was just registered.
+      # _The step definition that was just registered.
```
mattwynne: I think the original underscore there is a mistake and could be removed.

lib/cucumber/events/step_definition_registered.rb
```diff
@@ -4,16 +4,16 @@
...
       #
       # @return [RbSupport::RbStepDefinition]
       attr_reader :step_definition
 
-      #_@private
+      # _@private
```
mattwynne: this underscore here also looks wrong to me.

spec/cucumber/multiline_argument/data_table_spec.rb
```diff
@@ -8,9 +8,9 @@ module MultilineArgument
     describe DataTable do
       before do
         @table = DataTable.from([
-          %w{one four seven},
-          %w{4444 55555 666666}
-        ])
+                                  %w{one four seven},
```
mattwynne: I don't feel all that strongly about it, but I personally prefer the original indentation. What do others think?

mattwynne:
In general @phoebeclarke I think that in future having more smaller PRs would make it easier to review these changes, and merge the un-contentious ones :)

mattwynne:
Thanks again for all your effort @phoebeclarke 👍

lib/cucumber/events/step_activated.rb
nodo: The rule is here: https://github.com/bbatsov/ruby-style-guide#parallel-assignment. The motivation seems to be related to readability. However, I don't mind to keep it like this, feels more concise and equally readable 👍

spec/cucumber/multiline_argument/data_table_spec.rb
nodo: I would follow the ruby-style-guid in this case. If you feel it's better we could lower the `[` parenthesis and keep the same indentation.

spec/cucumber/formatter/legacy_api/adapter_spec.rb
```diff
@@ -1703,7 +1703,7 @@ def apply_before_hooks(test_case)
             end
 
             expect( formatter.legacy_messages ).to eq([
-                :before_features,
+                                                        :before_features,
                   :before_feature,
```
nodo: I wouldn't change the indentation of this file. I think the spaces gives an intuition of what is going on and following the rules here has less value. What do you think @phoebeclarke @mattwynne ?

lib/cucumber/events/step_activated.rb
mattwynne: Pft. I do see the point about readability. I guess it's probably clearer to go with the style guide, and it's just me being lazy when I write constructors :)

spec/cucumber/multiline_argument/data_table_spec.rb
mattwynne: Good idea!

spec/cucumber/formatter/legacy_api/adapter_spec.rb
mattwynne: I think we could safely exclude this file from linting, I hope we'll be able to remove it from the project altogether fairly soon.

lib/cucumber/events/step_activated.rb
nodo: 👍  😄

danascheider: Yeah, I'm a fan of parallel assignment tbh.

dkowis: If it's more than one or two things, it can get very confusing. That's probably why the style guide just recommends putting them each on their own line.

danascheider: True, I definitely wouldn't use it for more than 3 things.

spec/cucumber/formatter/legacy_api/adapter_spec.rb
danascheider: When I was making my recent PRs with Rubocop fixes I opted not to change the indentation in this file and to keep it excluded for the reason @nodo gave.

PR #1086, opened 2017-02-26
adding result in junit testcase report

yeshwant-singh:
## Summary
Adding status in junit xml report for individual test cases. 

## Details
In junit xml report generator, individual testcase/scenario status is not shown.

## Motivation and Context
- I wanted to send html report of cucumber tests , for that I was creating a simple table like html in which individual testcases's status, time are shown in rows. Time was mentioned in xml report for each scenario/testcase, but status was not mentioned.
- I tried to write an XSLT file reading the xml file and added conditions like if std error is mentioned then scenario is passed else failed. 
- But finally I added this change in junit file and made my work easy.

## How Has This Been Tested?
- Ran the tests for failed and passed scenario each.
- Test setup: ruby 2.3, cucumber 2.4.0 
- Code doesn't effect any other areas

## Types of changes

## Checklist:

### DIFF
--- a/features/docs/formatters/junit_formatter.feature
+++ b/features/docs/formatters/junit_formatter.feature
@@ -120,7 +120,7 @@ Feature: JUnit output formatter
           <![CDATA[]]>
         </system-err>
       </testcase>
-      <testcase classname="Subdirectory - One passing scenario, one failing scenario" name="Failing" time="0.05">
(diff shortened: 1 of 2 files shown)


yeshwant-singh:
fixed test cases

yeshwant-singh:
Fixed all the test cases , build passing

PR #1087, opened 2017-02-26
Add ruby-head as a allow failures in .travis.yml.

junaruga:
## Summary

Change Ruby 2.4.0 from allow failures to required.

## Details

I changed Ruby 2.4.0 from allow failures to required.
Because this looks better than current "allow_failures".

We can skip Ruby 2.4's coverage until `coveralls` supports `simplecov` (>= 0.13.0) as a dependency package for Ruby 2.4 compatibility.
https://rubygems.org/gems/coveralls

I added "ruby-head" as a "allow_failures".
I think this is useful, because we can prepare `cucumber` for next Ruby version 2.5 in advance before it will be released.

Adding "ruby-head" as a "allow_failures" is actually practiced by `rspec` and  `rails`.
And this is the secret that they could support Ruby 2.4.0 quickly.
So, I want to add it to `cucumber` too.

https://github.com/rspec/rspec-core/blob/master/.travis.yml
https://github.com/rails/rails/blob/master/.travis.yml

`fast_finish` option in `.travis.yml` is also useful.
https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

## Motivation and Context

It  looks better than current "allow_failures".

## How Has This Been Tested?

We can see if my code is correct on Travis CI.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 
 matrix:
   include:
+    - rvm: ruby-head
     - rvm: 2.4.0
     - rvm: 2.3.1
     - rvm: 2.2
@@ -18,7 +19,9 @@ matrix:
         - LANG=en_US.UTF-8
         - LANGUAGE=en_US.UTF-8
   allow_failures:
+    - rvm: ruby-head
     - rvm: 2.4.0
+  fast_finish: true
 
 # whitelist
 branches:


Rakefile
```diff
@@ -11,7 +11,10 @@ RuboCop::RakeTask.new
 
 task :default => [:spec, :rubocop, :cucumber]
 
-if ENV['TRAVIS']
+# Skip Ruby 2.4 until coveralls supports simplecov (>= 0.13.0)
+# for Ruby 2.4 compatibility.
+# https://rubygems.org/gems/coveralls
+if ENV['TRAVIS'] && RUBY_VERSION < '2.4'
   ENV['SIMPLECOV']  = 'ci'
```
nodo: I am not sure about this. What is the advantage in skipping 2.4 coverage? I think having ruby 2.4.0 in allowed failures is not great, but at least is more explicit.

Also, as far as I understand there is also PR open on `coveralls` to fix this: https://github.com/lemurheavy/coveralls-ruby/pull/117

junaruga: @nodo Yeah, actually it does not matter. This is my favorite. But if the cucumber reviewers does not feel it is problem, I do not change the code.

junaruga:
@nodo I did rebase of my code. and changed commit message, and this PR's title.
I removed my modification for `Rakefile` for your code review.

Rakefile
junaruga: @nodo yes there is the PR on `coveralls`. I am also watching it.

nodo:
Thanks @junaruga 👍

mattwynne:
Nice work @junaruga :)

junaruga:
@nodo @mattwynne :)

junaruga:
I am gong to send this kind of PR for `ruby-head` to all the `cucumber` Ruby related sub projects as much as possible little by little.

PR #1097, opened 2017-03-27
Fix diffing empty DataTables

### LINKED ISSUE #1096: Diffing empty DataTables crashes with a nil error
## Current Behavior

Opinions seem to vary on exactly how to construct an empty (0x0) DataTable, but in either case:

```ruby
2.2.6 :001 > require 'cucumber'
2.2.6 :002 > include Cucumber::MultilineArgument
2.2.6 :003 > DataTable.from([[]]).diff! [[]]
NoMethodError: undefined method `length' for nil:NilClass
	from /home/micah/.rvm/gems/ruby-2.2.6@chop/gems/cucumber-2.4.0/lib/cucumber/multiline_argument/data_table.rb:357:in `diff!'
2.2.6 :004 > DataTable.from([]).diff! []
NoMethodError: undefined method `length' for nil:NilClass
	from /home/micah/.rvm/gems/ruby-2.2.6@chop/gems/cucumber-2.4.0/lib/cucumber/multiline_argument/data_table.rb:355:in `diff!'
```

This ends up boiling down to `C::MA::DT#pad_and_match` calling `Array#transpose` on the internal 2d array, and both `[[]].transpose` and `[].transpose` both return `[]`, which causes the DataTable implementation to crash with a nil error, because its expecting a 2d array, but it only has a 1d array to work with.

## Expected Behavior

One might argue that the bug is truly in Ruby's `Array#transpose` function (confirmed same behavior in ruby-head, btw), but regardless, I think we can agree that `#diff!` shouldn't nil crash in this circumstance. I would like to see at least one of those two invocations succeed, and possibly both.

## Context & Motivation

The motivation behind this request is that I am heavily using DataTables in my cucumber scenarios, and have even written a library named [chop](https://github.com/botandrose/chop) that makes DataTables much more powerful when used in conjunction with ActiveRecord and Capybara. The specific use-case is as follows:

I have a method that creates a data structure based off of the current state of the Capybara page, lets say its a table of users, and I want to use the `#diff!` method to assert that it matches an expected state, which could include being completely empty.

```ruby
Then "I should see the following users:" |table|
  table.diff! users_table
end

Then "I should see no users"
  empty_table.diff! users_table
end

def users_table
  all("table.users tr").map do |row|
    row.all("td").map do |cell|
      cell.text
    end
  end
end

def empty_table
  Cucumber::MultilineArgument::DataTable.from([[]]) # or [], perhaps
end
```

This may seem a bit awkward, like why not just do `expect(users_table).to eq([])`, but this is a gross simplification of the actual system. Suffice it to say that using `C::MA::DT#diff!` in both circumstances is greatly preferable. One could look at the implementation and usage of `Chop::Diff` for a better understanding of the actual system, if desired.

## Next Steps

Would a PR addressing this issue be considered for merging? If so, I am happy and eager to do the work.

botandrose:
## Summary
Fixes the following:
```ruby
DataTable.from([[]]).diff! [[]] #=> NoMethodError: undefined method `length' for nil:NilClass
```
See #1096 for much more details.

## Details
Added a private `#ensure_2d` method to DataTable, and used it internally in `#pad_and_match` to make sure we're always dealing with an array of arrays, which the code had previously been assuming.

## Motivation and Context
Fixes #1096

## How Has This Been Tested?
Added unit test to expose bug, then wrote patch to fix it.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/multiline_argument/data_table.rb
+++ b/lib/cucumber/multiline_argument/data_table.rb
@@ -589,7 +589,7 @@ module Cucumber
             matched_cols << unmatched_cols.delete_at(mapped_index)
           else
             mark_as_missing(cols[i])
-            empty_col = other_cell_matrix.collect {SurplusCell.new(nil, self, -1)}
+            empty_col = ensure_2d(other_cell_matrix).collect {SurplusCell.new(nil, self, -1)}
             empty_col.first.value = v
             matched_cols << empty_col
           end
@@ -601,7 +601,11 @@ module Cucumber
           cols << empty_col
         end
 
-        return cols.transpose, (matched_cols + unmatched_cols).transpose
+        return ensure_2d(cols.transpose), ensure_2d((matched_cols + unmatched_cols).transpose)
+      end
+
+      def ensure_2d(array) #:nodoc:
+        Array === array[0] ? array : [array]
       end
 
       def ensure_table(table_or_array) #:nodoc:
--- a/spec/cucumber/multiline_argument/data_table_spec.rb
+++ b/spec/cucumber/multiline_argument/data_table_spec.rb
@@ -332,6 +332,12 @@ module Cucumber
           }
         end
 
+        it 'should allow diffing empty tables' do
+          t1 = DataTable.from([[]])
+          t2 = DataTable.from([[]])
+          expect{ t1.diff!(t2) }.not_to raise_error
+        end
+
         context 'in case of duplicate header values' do
           it 'raises no error for two identical tables' do
             t = DataTable.from(%{


lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -601,7 +601,11 @@ def pad_and_match(a_cell_matrix, other_cell_matrix) #:nodoc:
           cols << empty_col
         end
 
-        return cols.transpose, (matched_cols + unmatched_cols).transpose
+        return ensure_2d(cols.transpose), ensure_2d((matched_cols + unmatched_cols).transpose)
+      end
+
+      def ensure_2d(array) #:nodoc:
```
danascheider: I'm wondering if we should make this method private.

spec/cucumber/multiline_argument/data_table_spec.rb
```diff
@@ -332,6 +332,12 @@ module MultilineArgument
           }
         end
 
+        it 'should allow diffing empty tables' do
+          t1 = DataTable.from([[]])
+          t2 = DataTable.from([[]])
+          t1.diff!(t2)
+        end
```
danascheider: Maybe use something like
```
expect{ t1.diff!(t2) }.not_to raise_error
```
would be better?

botandrose: Technically redundant, but more explicitly conveys the intent. I like! 👍

lib/cucumber/multiline_argument/data_table.rb
botandrose: It's already `protected`... is that good enough?

danascheider: Oh yeah, that's great. I didn't see that.

spec/cucumber/multiline_argument/data_table_spec.rb
danascheider: Cool, if you could just change that and push it up I'll merge :) Thanks!

botandrose:
Okay! Pushed, and all :green_heart: here. Thank you for your time and attention!

I'll be submitting fixes for one or two more DataTable-related bugs over the coming days, BTW.

danascheider:
Thank you @botandrose - we really appreciate your help with the project!

mattwynne:
Thanks for handling it @danascheider 👌

PR #1102, opened 2017-04-01
Fix incorrect DataTable#diff! report

botandrose:
## Summary

This fixes a bug in `#diff!` that was causing incorrect values to be displayed for the header row in the diff report.

## Details

If columns were missing in the first table, they would be filled in with instances of `SurplusCell`. However, the same instance of `SurplusCell` was being used for every missing column in the header column, so when the value was being set on one cell, it was being set on them all. This led to the last value "winning," and the diff report reporting an incorrect diff summary. For example:

```ruby
DataTable.from([[]]).diff! DataTable.from(%{
  |a|b|c|
  |d|e|f|
  |g|h|i|
})

            | (+) c | (+) c | (+) c |
            | (+) d | (+) e | (+) f |
            | (+) g | (+) h | (+) i |
```

The fix is simply to generate a new instance of SurplusCell for each cell.

## How Has This Been Tested?

Added two tests to expose the bug, and added the patch to fix.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/multiline_argument/data_table.rb
+++ b/lib/cucumber/multiline_argument/data_table.rb
@@ -596,8 +596,8 @@ module Cucumber
         end
 
 
-        empty_col = cell_matrix.collect {SurplusCell.new(nil, self, -1)}
         unmatched_cols.each do
+          empty_col = cell_matrix.collect {SurplusCell.new(nil, self, -1)}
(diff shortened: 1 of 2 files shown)


danascheider:
Great work on this, @botandrose! Thanks for your contribution!

PR #1103, opened 2017-04-05
Refactor DataTable#diff!

botandrose:
## Summary

While fixing some bugs in `DataTable#diff!`, I noticed that its pretty unwieldy, and could benefit from some refactoring. Code Climate agrees with me, giving the class an [F](https://codeclimate.com/github/cucumber/cucumber-ruby/Cucumber::MultilineArgument::DataTable#source). This refactoring PR aims to improve the code quality.

## Details

If one is interested in the the details, they are probably best elucidated by reviewing the diff of each commit. In broad strokes, this PR aims to extract all the diff-related complexity from `DataTable` into a new private class: `DataTable::DiffMatrices`. There is certainly more work that could be done in this vein. `#pad_and_match` and `#perform_diff` are still rather complex, and I have a hunch that the overall complexity could be reduced further, but I think this is a good first step, and also I want to get some feedback before spending too much time on this. :)

What do y'all think?


botandrose:
I pushed a few more commits. 2dbaccf could be considered a breaking change, in case someone was relying upon the diffing table being mutated by `#diff!`. Seems very unlikely, since one would have to first rescue from the diff exception to even notice the mutation. I'd describe this more as clarifying undefined behavior, since it seems pretty unlikely that anyone would expect this, or even encounter it, much less rely upon it. However, it appears you folks are gearing up for a 3.0.0 release, so I thought I'd try to get this in, now, if its desirable, to be safe.

So yeah, overall, I think things are improved by lowered complexity and reduced mutation. Any thoughts here? How does this look?

danascheider:
Thanks for this contribution @botandrose! Any chance you could separate out the new class you added into a separate file? We have a lot of long files and I think they make the codebase less maintainable.

botandrose:
@danascheider No problem! Thanks for taking a look at this. I'll push out a commit shortly.

botandrose:
I think the AppVeyor failure is unrelated to this change:

`undefined method `scenario_outline' for #<Cucumber::RunningTestCase::Scenario:0x49e70f8> (NoMethodError)`

Looks like AppVeyor is broken on master, too. :/

danascheider:
Thanks @botandrose! Really appreciate the help!

botandrose:
:tada:

PR #1104, opened 2017-04-06
History: change bad character to space

olleolleolle:
## Summary

Render markdown in the changelog correctly, by replacing a non-space character with a space.


nodo:
Good catch @olleolleolle , thanks! 👍

PR #1114, opened 2017-05-26
Travis: use jruby-9.1.10.0

olleolleolle:
This PR updates the CI matrix to use latest JRuby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     - rvm: 2.3.4
     - rvm: 2.2
     - rvm: 2.1
-    - rvm: jruby-9.1.8.0
+    - rvm: jruby-9.1.10.0
       env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ matrix:
     - rvm: 2.3.4
     - rvm: 2.2
     - rvm: 2.1
-    - rvm: jruby-9.1.8.0
+    - rvm: jruby-9.1.10.0
       env:
```
nodo: Thanks @olleolleolle, what about adding `jruby-1.9.1` instead? So we can keep track of the latest major version?

olleolleolle: @nodo If that would work, which it does not, that would be cool.

nodo: Right, that is unfortunate. Anyway, thanks for the PR, looks good to me! Possibly we can keep track of `jruby` head if it adds value.

Issue #1044, opened 2016-10-04
When running with --retry, exit code should be 0 if a failed scenario passes on retry


rishi-freshbooks:
## Summary

When running cucumber with the --retry flag, the process will exit with a non-zero code if the scenario fails on first try but passes on a retry.
## Expected Behavior

Running cucumber with --retry flag should exit with code 0 if scenarios pass on first try or within the specified number of retries .
## Context & Motivation

The non-zero exit code causes our build job in JenkinsCI to fail when it should report as success.

rishi-freshbooks:
@mattwynne does this capture the issue appropriately? Is this something that can be addressed in v3.0.0.pre.1?

mattwynne:
That's great thanks @rishi-freshbooks. We'll do our best to fix this in a future pre-release of v3, yes.

streetlogics:
@mattwynne any progress on this?  Just encountering this issue myself - I love that I finally got the --retry flag working in circleci (note: I had to put the command at the end of the line otherwise a --format option that had an --out param seemed to swallow the retry attempt), however, once the retry started "working", circle still saw the run as a failure. 😞

streetlogics:
@mattwynne - I took a really ugly pass at this just to get something working.  Definitely open to any thoughts you might have on improving this for the actual implementation

mattwynne:
Awesome, thanks @streetlogics one of us will take a look ASAP

garie:
Any update? My team is also interested in this fix as a non-zero exit code causes our Bamboo build to fail.

garside:
Just ran into the `--retry` flag and the non-zero exit code for circleci. Just hoping to get a pulse check on the status of this retry-success feature that @streetlogics proofed.

@mattwynne, Is there an ETA for it and/or is there anything we could do to help push this across the finish line?

brasmusson:
Yes, in non-strict mode the exit code should be 0 is all scenarios passes on retry, but in strict mode (option `--strict`) the exit code should not be 0 is any scenarios were retried.

I think a that first the summary should  be updated with a new flaky category (as in `6 Scenarios (1 Passed, 1 Failed, 1 Undefined, 1 Pending, 1 Skipped, 1 Flaky)`), then it is straight forward to implement the exit code logic. In non-strict mode the exit code should be 0 if and only if there no failed scenarios. In strict mode the exit code should be 0 if and only if there are no failed, flaky, pending or undefined scenarios.

[...]

PR #1121, opened 2017-06-13
error code 0 on successful `--retry`

gondalez:
## Summary

A basic working version of --retry based on work by @streetlogics, in the time I had available.

The logging could use some work and it would be nice to get feedback from the maintainers.

I figured I would put this here in case it helps someone else out and maybe it can go part-way towards a fix for #1044 :)

## Details

I cherry picked the commits https://github.com/cucumber/cucumber-ruby/compare/master...streetlogics:issue/retry-success-exit-code-0-1044 on top of master.
See https://github.com/cucumber/cucumber-ruby/issues/1044#issuecomment-278864246.

## Motivation and Context

I tried using @streetlogics branch but I got an ArgumentError when a feature is retried.
This seems to have been fixed by 31968ae09ae5ff8898cbc571ac9a932c98f0bbbd, that was not yet in the remote branch.

## How Has This Been Tested?

I tested quickly locally and can confirm the error code is now 0 if a step succeeds on subsequent retry.

### DIFF
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -231,9 +231,10 @@ module Cucumber
       elsif @configuration.retry_attempts > 0
         if summary_report.test_cases.total_passed == @configuration.total_cases
           Cucumber.logger.info "All retried test cases passed!\n" if summary_report.test_cases.total_failed > 0
-          return false
+          false
+        else
+          true
         end
-        true
       else
         summary_report.test_cases.total_failed > 0 || summary_report.test_steps.total_failed > 0 ||
           (@configuration.strict? && (summary_report.test_steps.total_undefined > 0 || summary_report.test_steps.total_pending > 0))
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -228,6 +228,8 @@ module Cucumber
     def failure?
       if @configuration.wip?
         summary_report.test_cases.total_passed > 0
+      elsif @configuration.retry_attempts > 0
+        summary_report.test_cases.total_passed != @configuration.total_cases
       else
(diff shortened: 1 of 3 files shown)


danascheider:
Thanks @gondalez! If you can rebase this off of master I'll help figure out what's breaking in jruby.

gondalez:
@danascheider no probs; rebase is done and the jruby build passed this time 🎉

lib/cucumber/runtime.rb
```diff
@@ -228,6 +228,12 @@ def legacy_formatter?(factory)
     def failure?
       if @configuration.wip?
         summary_report.test_cases.total_passed > 0
+      elsif @configuration.retry_attempts > 0
+        unless summary_report.test_cases.total_passed != @configuration.total_cases
```
olleolleolle: Golfing the code: 

- an `unless` with `!=` could become an `if` with a `==`

Would that read better?

gondalez: @olleolleolle 👍 fixed :)

brasmusson:
As I commented on the [original PR](https://github.com/cucumber/cucumber-ruby/issues/1044#issuecomment-306458646), I think the exit code should be non-zero in `--strict` mode also when all scenarios are passed on rerun, that the summary should list the number of scenarios passed on rerun in the new `flakey` category, and that the exit code should be based on the summary (zero in `--strict` mode if all scenarios are `passed` or `skipped`, zero in non-strict mode if no scenarios are `failed` or `ambiguous`). I can't see that an implementation adding `Configuration#total_cases` is the implementation we want.

gondalez:
@brasmusson That makes sense. I like the idea of a new category.
I'm happy to try my hand at an implementation over the coming weeks. Can you give me any helpful pointers?

brasmusson:
Superseded by cucumber/cucumber-ruby-core#141 and #1159.

PR #1130, opened 2017-06-24
Step: Rename name to text

olleolleolle:
This PR is a follow-up to (the now-merged) https://github.com/cucumber/cucumber-ruby-core/pull/137 

It

- [x] renames `Step`'s "name" to "text"
- [x] calculates widths of lines using the brand-new `to_s` implementation from Core


olleolleolle:
This is not all-right yet, I'm missing something, and the `./spec/cucumber/formatter/json_spec.rb` does not do the right thing, yet.

olleolleolle:
Rebased on top of current master; sorry for churn.

PR #1131, opened 2017-06-25
Remove programming language abstraction

tooky:
The old programming language abstraction is confusing. This starts to improve the names and lay the groundwork to get rid of all the mess it hides.

There's lots of global state and mutation that we can clean up.


mattwynne:
@tooky I think we should marge this. It's already making things loads better.

PR #1116, opened 2017-05-29
Expanded ci

enkessler:
## Summary

Adds OSX and Windows to the CI process for the project.

This PR still needs work. The wiring for getting the builds running on OSX and Windows is in place but the project does not actually build cleanly on those platforms.

## Details

Updated the Travis-CI configuration so that it runs tests against OSX and added a configuration file for running the build on AppVeyor (which provides Windows CI)

## Motivation and Context

Fixes #1115

## How Has This Been Tested?

It's a change to the CI process. It is inherently 'tested' with every build.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1133, opened 2017-06-26
Fixing the profile loader

enkessler:
Fixed a profile parsing bug that was only affecting Windows.

## Summary

A bug in the profile loader was discovered while working on #1116. This PR fixes that bug.

## Details

Bug fix + tests. Rather than try to re-implement the parsing being done by Shellwords, I simply _temporarily_ adjusted the input data so that Shellwords could handle it as usual.
 
## Motivation and Context

The problem was discovered during #1116 but it was decided to handle it separately rather than making it a blocking problem on that PR.

## How Has This Been Tested?

New spec added for the profile loader.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/cli/profile_loader_spec.rb
+++ b/spec/cucumber/cli/profile_loader_spec.rb
@@ -43,8 +43,17 @@ default: <%= x %>
         expect(loader.args_from('default')).to eq ['--format','pretty','features/sync_imap_mailbox.feature:16:22']
       end
 
-      it 'correctly parses a profile that uses tag expressions' do
+      it 'correctly parses a profile that uses tag expressions (with double quotes)' do
         given_cucumber_yml_defined_as({'default' => '--format "pretty" features\sync_imap_mailbox.feature:16:22 --tags "not @jruby"'})
+        if(Cucumber::WINDOWS)
+          expect(loader.args_from('default')).to eq ['--format','pretty','features\sync_imap_mailbox.feature:16:22','--tags','not @jruby']
+        else
+          expect(loader.args_from('default')).to eq ['--format','pretty','featuressync_imap_mailbox.feature:16:22','--tags','not @jruby']
+        end
+      end
+
+      it 'correctly parses a profile that uses tag expressions (with single quotes)' do
+        given_cucumber_yml_defined_as({'default' => "--format 'pretty' features\\sync_imap_mailbox.feature:16:22 --tags 'not @jruby'"})
         if(Cucumber::WINDOWS)
           expect(loader.args_from('default')).to eq ['--format','pretty','features\sync_imap_mailbox.feature:16:22','--tags','not @jruby']
         else
--- a/spec/cucumber/cli/profile_loader_spec.rb
+++ b/spec/cucumber/cli/profile_loader_spec.rb
@@ -48,7 +48,7 @@ default: <%= x %>
         if(Cucumber::WINDOWS)
           expect(loader.args_from('default')).to eq ['--format','pretty','features\sync_imap_mailbox.feature:16:22','--tags','not @jruby']
         else
-          expect(loader.args_from('default')).to eq ['--format','pretty','featuressync_imap_mailbox.feature:16:22']
+          expect(loader.args_from('default')).to eq ['--format','pretty','featuressync_imap_mailbox.feature:16:22','--tags','not @jruby']
         end
       end
 
--- a/spec/cucumber/cli/profile_loader_spec.rb
+++ b/spec/cucumber/cli/profile_loader_spec.rb
@@ -42,6 +42,16 @@ default: <%= x %>
         given_cucumber_yml_defined_as yml
         expect(loader.args_from('default')).to eq ['--format','pretty','features/sync_imap_mailbox.feature:16:22']
       end
+
+      it 'correctly parses a profile that uses tag expressions' do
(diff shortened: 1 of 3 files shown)


spec/cucumber/cli/profile_loader_spec.rb
```diff
@@ -42,6 +42,16 @@ def loader
         given_cucumber_yml_defined_as yml
         expect(loader.args_from('default')).to eq ['--format','pretty','features/sync_imap_mailbox.feature:16:22']
       end
+
+      it 'correctly parses a profile that uses tag expressions' do
```
brasmusson: I would also add an example using `'` around the argument containing blanks, for instance `"--format pretty features --tags 'not @jruby'"` (which also avoids the path separator problem), to be very sure that both `"` and `'` can be used to define arguments with blanks in them.

enkessler: Done.

enkessler: I don't know why this PR isn't noticing the new commit that I made. The new test has been added to this branch but it's not showing up...

enkessler: There it is. GitHub is slow today. Anything else @brasmusson?

brasmusson: Now, it looks good to me.

olleolleolle:
@enkessler _Maybe_ there is a quick way to green about these RuboCop warnings:

- `rubocop --auto-correct spec/cucumber/cli/profile_loader_spec.rb` would get rid of the `{ ... }` complaints
- and the one about `Exception` being too wide a rescue, it can be ` # rubocop:disable Lint/RescueException`

enkessler:
Strange. The rubocop task doesn't have any issues when I run it locally.

nodo:
@enkessler It's because the file is ignored by rubocop. You need to change `.rubocop_todo.yml` first.

enkessler:
Ah, right. I read your comment too quickly.

enkessler:
@olleolleolle I opted to just go ahead and change the exception type. Not catching _everything_ is probably good advice and `StandardError` is what is being caught in other areas of the code. I checked out the history and nothing indicates that it was a broader net intentionally. Maybe that's just how the contributor wrote code 8 years ago.

spec/cucumber/cli/profile_loader_spec.rb
```diff
@@ -42,6 +42,25 @@ def loader
         given_cucumber_yml_defined_as yml
         expect(loader.args_from('default')).to eq ['--format','pretty','features/sync_imap_mailbox.feature:16:22']
       end
+
+      it 'correctly parses a profile that uses tag expressions (with double quotes)' do
```
mattwynne: I'm wondering if this test name is misleading. It seems to me that the bug we've found is to do with parsing *any* switches given after the feature file paths, and that tag expressions just happen to be the example we've found of that general defective behaviour. Am I right?

spec/cucumber/cli/profile_loader_spec.rb
```diff
@@ -42,6 +42,25 @@ def loader
...
+        given_cucumber_yml_defined_as('default' => '--format "pretty" features\sync_imap_mailbox.feature:16:22 --tags "not @jruby"')
+        if(Cucumber::WINDOWS)
+          expect(loader.args_from('default')).to eq ['--format','pretty','features\sync_imap_mailbox.feature:16:22','--tags','not @jruby']
+        else
+          expect(loader.args_from('default')).to eq ['--format','pretty','featuressync_imap_mailbox.feature:16:22','--tags','not @jruby']
+        end
+      end
+
+      it 'correctly parses a profile that uses tag expressions (with single quotes)' do
+        given_cucumber_yml_defined_as('default' => "--format 'pretty' features\\sync_imap_mailbox.feature:16:22 --tags 'not @jruby'")
```
mattwynne: I don't really understand what's going on with the backslashes in these tests! 😁 

This one uses two, but the one on line 47 uses one. And the expected parsed output on non-windows OSes has no slashes in the path?!

I realise this is code we've inherited from our ancestors, but now that you've looked at it, are you able to explain what it means @enkessler?

brasmusson: @mattwynne From the irb:
```
irb(main):004:0> puts 'dir\file'
dir\file
=> nil
irb(main):005:0> puts "dir\file"
dir
   ile
=> nil
irb(main):006:0> puts "dir\\file"
dir\file
=> nil
```
Within `""` the `\` is interpreted as "escape" so if you literally mean `\` you have to use `\\`, which when the `""` expression is "evaluted" is "converted" to `\`. (`\f` in the `""` expression above means ["formfead"](https://stackoverflow.com/a/31791519))

mattwynne: Thanks @brassmusson, that makes sense.

I'm still puzzled why the expected output on non-windows has no slashes in the path at all though. What's going on there?!

spec/cucumber/cli/profile_loader_spec.rb
enkessler: I think you're right.

spec/cucumber/cli/profile_loader_spec.rb
enkessler: I don't know why the output is different either but I just copy/paste/tweaked for the new test.

As for why a certain number of backslashes are needed...well...they are basically [magic](https://xkcd.com/1638/). ;)

PR #1135, opened 2017-06-27
RuboCop setting: add name of failed cop rule; URL

olleolleolle:
## Summary

This PR enables displaying RuboCop **cop names** and **styleguide URLs** for each failed cop.

## Details

To help people trying to fix or disable cop warnings, the name and URL helps.

We are currently on RuboCop 0.40.0, and [these are that version's configuration option defaults](https://github.com/bbatsov/rubocop/blob/v0.40.0/config/default.yml).

## Motivation and Context

I saw someone struggling with cop output. The tool ought to present all it knows.

## Screenshot

Here is a picture with some failures that I added to one file. 

In practice, this is what it looks like **with extra information:**

screenshot 2017-06-27 11 11 00

Contrast this with the before **without extra information:**

screenshot 2017-06-27 11 18 18

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,13 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
+  # Cop names are not displayed in offense messages by default. Change behavior
+  # by overriding DisplayCopNames, or by giving the `-D/--display-cop-names`
+  # option.
+  DisplayCopNames: true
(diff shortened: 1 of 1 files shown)


olleolleolle:
@nodo Is this "Developer Experience"-Label-worthy? If yes, yay!

nodo:
Good for me :) Thanks for merging, feel free to update HISTORY.md

Issue #916, opened 2015-09-23
Dev environment should be friendly to Windows users


mattwynne:
In https://github.com/cucumber/cucumber-ruby/issues/915#issuecomment-142438357 @enkessler noted that Cucumber's build doesn't work on Windows.

This excludes Windows users from contributing to Cucumber.

Is anyone willing to help fix this? We can use https://ci.appveyor.com to add CI once we have a build that works.

tomatobrown:
I have the cucumber-ruby gem working on Windows, maybe I'm not understanding your problem

aslakhellesoy:
@tomatobrown this is about running cucumber's own build on Windows. Clearer?

enkessler:
@mattwynne @aslakhellesoy PR #1116 has added a Windows build to the CI process. There is some remaining work in getting that build green, however.

All specs pass but there are still [failing features](https://ci.appveyor.com/project/enkessler/cucumber-ruby-u5iqh/build/1.0.1/job/1euuunkgnb6h2i7a).

brasmusson:
It looks that some (all?) of the failing features would be fixed if Cucumber-Ruby also on Windows understood that `/` in the feature path argument means the path separator. Luckily for Cucumber-JVM, Java on Windows understands that, therefore Cucumber-JVM always output the `uri` with `/` as the path separator and it works on all platforms (for instance uri's with `/` in the `rerun.txt` file).

enkessler:
@brasmusson Wouldn't that mean that we only use path separators in 1 in 3 of our tests? Additionally, several of those path errors go away when that command is invoked via backticks instead of through Aruba. Shouldn't the path error still be there if it were a Ruby on Windows problem?

aslakhellesoy:
We have an Appveyor build, it just isn't passing. Closing this in favour of #1215 which has a more specific acceptance criterion (make the build pass)

PR #1138, opened 2017-06-29
Extract HtmlBuilder class

danascheider:
## Summary

Our HTML formatter class has an F rating in CodeClimate, partially due to implementing a lot of logic that should probably belong to an HTML builder class. This PR extracts a `Cucumber::HtmlBuilder` class that subclasses `Builder::XmlBuilder` and handles some of the tasks of building elements the formatter uses.

## Details

This PR extracts the `Cucumber::HtmlBuilder` class, which defines methods that build HTML nodes used by the HTML formatter to embed text and images, create an HTML document, add inline CSS and JavaScript, and other actions.

This is a work in progress and still needs work to make sure the code is of high quality and decoupled from the HTML formatter. Feedback is welcome with the understanding that this is not yet a finished product.

## Motivation and Context

CodeClimate has identified the HTML formatter as one of the most problematic parts of our code base. In order to make it more maintainable, I have extracted this class as the first step in a larger refactor, as discussed with @mattwynne at CukenFest.

## How Has This Been Tested?

This refactor has not affected existing tests covering the HTML formatter, but I will add tests for the HtmlBuilder class before removing the WIP label.

## Types of changes

- [x] Refactor (change improves maintainability without affecting functionality)

## Checklist:


danascheider:
@enkessler I could use some help with the AppVeyor failures - any idea what's going on with this?

brasmusson:
@danascheider I suspect that part of the AppVeyor issues has to do with that [Cucumber-Ruby on Windows does not understand the `features/file.feature` means the file `file.feature` in the directory `features`](https://github.com/cucumber/cucumber-ruby/issues/916#issuecomment-311107215)

brasmusson:
One other thing we really want to do with the HTMLFormatter is to re-write it to use the new event API.

danascheider:
Thanks @brasmusson!

enkessler:
@danascheider I put in a PR that turns off the existing Windows failures so that we have a green starting point. Pull in the changes from master and just don't cause any new failures.

lib/cucumber/html_builder.rb
```diff
@@ -73,7 +74,8 @@ def prepend_to_span(span_class, content)
 
     def inline_css
       style(type: 'text/css') do
-        self << ::File.read(::File.dirname(__FILE__) + '/cucumber.css')
+        pn = ::Pathname.new(::File.dirname(__FILE__) + '/cucumber.css')
+        self << ::File.read(pn)
```
olleolleolle: Golf note: One cool thing about Pathname objects is that they can be READ:

```ruby
self << pn.read
```

lib/cucumber/formatter/html.rb
```diff
@@ -25,88 +26,56 @@ def initialize(runtime, path_or_io, options)
...
-            type = mime_type =~ /;base[0-9]+$/ ? mime_type : mime_type + ';base64'
-            src = 'data:' + type + ',' + src
-          end
-          embed_image(src, label)
-        when /^text\/plain/
-          embed_text(src, label)
+        if image?(mime_type)
+          src = src_is_file_or_data?(src) ? src : "data:#{standardize_mime_type(mime_type)},#{src}"
+
+          @builder.embed_image(src: set_path(src), label: label, id: "img_#{@img_id}")
```
mattwynne: I'd use a private `attr_reader` for `builder` here. Sandi Metz taught me to basically never use instance variables.

lib/cucumber/formatter/html.rb
```diff
@@ -25,88 +26,56 @@ def initialize(runtime, path_or_io, options)
...
+          @img_id += 1
```
mattwynne: How about extracting a `next_id` method so we don't need to see this here?

e.g. `next_id(:img)`

```
def next_id(type)
  @indices ||= Hash.new { 0 }
  @indices[type] += 1
  "#{type}_#{@indices[type]}"
end
```

lib/cucumber/html_builder.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+require 'builder'
+
+module Cucumber
+  class HtmlBuilder < Builder::XmlMarkup
```
mattwynne: I think it would be best if this lives in the `Cucumber::Formatter` namespace so it sits alongside `formatter/html.rb`

lib/cucumber/inline-js.js
```diff
@@ -0,0 +1,30 @@
+SCENARIOS = "h3[id^='scenario_'],h3[id^=background_]";
```
mattwynne: I think this file should also be tidied away to sit next to `formatter/html.rb`

lib/cucumber/formatter/html.rb
danascheider: Love that idea. I'm not a huge fan of instance variables myself.

lib/cucumber/html_builder.rb
danascheider: I debated that with myself and don't feel that strongly one way or the other so I'll go ahead and change it.

lib/cucumber/inline-js.js
danascheider: Sounds good, I just moved it so I'll move it back.

lib/cucumber/formatter/html.rb
danascheider: Will do :)

lib/cucumber/html_builder.rb
danascheider: Oh cool, good to know.

danascheider:
@enkessler I got the same AppVeyor failures on a couple other PRs I made that didn't even touch the same parts of the app. The failing scenario appears to deal with internationalization and I can't figure out how that would've been affected by any of this. Given that I don't have access to a Windows computer, would you mind taking a look when you get a chance?

enkessler:
@danascheider You are getting the same, single test failure that master has on Windows. So I'd say that you are fine.

danascheider:
@brasmusson @mattwynne @enkessler Any objections to me merging this? I want to get this merged before I start working on a new event-based formatter since that will result in merge conflicts for sure.

enkessler:
@danascheider So this changes requires a documentation update but is not a breaking change? It is still backwards compatible?

mattwynne:
@danascheider if you're happy with it, merge away 👍

danascheider:
@enkessler It would potentially require a change to contributor documentation but tbh I'm not really sure why I checked that.

mattwynne:
👍  for merging your own PR @danascheider! It's good to see they're not piling up just because I'm away on vacation.

PR #1141, opened 2017-06-30
Add script to update history.md

mattwynne:
As discussed at CukenFest, this will avoid manual error when updating
history. It still needs to be run manually, and requires that one of the
`type: *` labels has been applied to the PR.

It's still going to save me time!


aslakhellesoy:
❤️️ this!

tooky:
Sweet!

PR #1147, opened 2017-07-04
Update cucumber yaml to use new tag syntax

danascheider:
This PR updates the Cucumber YAML file to use the new tag syntax, such as `"not @wire"` instead of the old syntax `~@wire`.

## Motivation and Context

This change will remove deprecation warnings that are currently appearing in CI.

## Types of changes

## Checklist:

### DIFF
--- a/cucumber.yml
+++ b/cucumber.yml
@@ -9,11 +9,11 @@ wip_opts = "--color -r features"
 wip_opts << " --tags @wip" if !defined?(JRUBY_VERSION)
 wip_opts << " --tags @wip,@wip-jruby" if defined?(JRUBY_VERSION)
 %>
-default:     <%= std_opts %> --tags ~@jruby
-jruby:       <%= std_opts %> --tags ~@wire
-jruby_win:   <%= std_opts %> --tags ~@wire CUCUMBER_FORWARD_SLASH_PATHS=true
-windows_mri: <%= std_opts %> --tags ~@jruby --tags ~@wire --tags ~@needs-many-fonts --tags ~@todo-windows CUCUMBER_FORWARD_SLASH_PATHS=true
-ruby:        <%= std_opts %> --tags ~@todo-windows --tags ~@jruby
+default:     <%= std_opts %> --tags "not @jruby"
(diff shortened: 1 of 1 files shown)


mattwynne:
Thanks @danascheider. We needed to merge #1133 before doing this, as the profile loader was totally borked on Windows without that.

That's merged so we can merge this now.

enkessler:
@danascheider @mattwynne The new tag syntax doesn't work on Windows even with #1133. That PR just makes it no longer explode. It still fails to function correctly.

mattwynne:
> @danascheider @mattwynne The new tag syntax doesn't work on Windows even with #1133. That PR just makes it no longer explode. It still fails to function correctly.

Ah I wasn't aware of that, so I've merged this prematurely, sorry. What's the problem behaviour we need to fix next then to get Windows back to green @enkessler?

mattwynne:
@enkessler can you establish whether the tests for https://github.com/cucumber/cucumber/tree/master/tag-expressions/ruby pass on windows?

enkessler:
I'll check when I get home today.

enkessler:
@mattwynne I cloned down the tag-expression project and all of the tests pass for me on my Windows machine. However, I don't see that even the bleeding edge of Cucumber uses the tag-expression gem.

mattwynne:
@enkessler the tag filter lives in the core: https://github.com/cucumber/cucumber-ruby-core/blob/master/cucumber-core.gemspec

enkessler:
Okay. When I use the bleeding edge version of cucumber-core then I am getting tag-expressions and the tests for my project are running normally. So...once the next release happens and tag-expressions is actually available, things will be okay. Until then, the pre-release version of cucumber-ruby doesn't fully work on Windows. But that's okay because it is a pre-release. ;)

mattwynne:
Thanks, I hadn't realised we had unreleased changes in core. If you keep a copy of `cucumber-ruby-core` in a folder next to `cucumber-ruby` the Gemfile should pick up the dev version of core automatically.

PR #1148, opened 2017-07-04
Fix Rubocop violations in the legacy API adapter

danascheider:
This PR fixes the Rubocop violations in the legacy API adapter and reconfigures Rubocop not to ignore this file.

## Summary

Removed the legacy API adapter file from the .rubocop-todo.yml file and then fixes the Rubocop violations that occur in that file.

## Details

There were two method calls where the args needed to be enclosed in parentheses; I added parens to both.

## Types of changes

## Checklist:

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -9,7 +9,6 @@
 # Offense count: 8
 Lint/AmbiguousOperator:
   Exclude:
-    - 'lib/cucumber/formatter/legacy_api/adapter.rb'
     - 'lib/cucumber/formatter/legacy_api/ast.rb'
     - 'lib/cucumber/multiline_argument/data_table.rb'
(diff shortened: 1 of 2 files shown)


nodo:
Actually, sorry @danascheider, if I remember correctly in cucumber codebase there is the convention to avoid parenthesis when the result of a method call is discarded. I wanted to write a custom cop for that, I will try to find some time for it.

olleolleolle:
@nodo An issue with precisely these parens missing is that they yield Ruby warnings. Perhaps the convention would opt for no-warnings, when there's no other way?

olleolleolle:
@nodo "I think", I should add.

brasmusson:
Let's not waste any time on the (hopefully soon) to be removed legacy_api, see #1230.

PR #1153, opened 2017-07-08
Rake task fix_cr_lf: exclude iso-8859-1 files

olleolleolle:
This PR changes the `fix_cr_lf` Rake task: it now fixes CRLF **on the right files**.

The previous exclusion rules were for files with `dos` in the filename – these files are no longer a part of the codebase.

Also: tell Git how to clone files using the [Dealing with line endings](https://help.github.com/articles/dealing-with-line-endings/) documentation that GitHub wrote.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,32 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text eol=lf
+
+# Explicitly declare text files to always be normalized and converted
+# to native line endings on checkout.
+*.css text eol=lf
+*.feature text eol=lf
+*.gemspec text eol=lf
+*.js text eol=lf
+*.md text eol=lf
+*.rake text eol=lf
+*.rb text eol=lf
+*.sass text eol=lf
+*.tcl text eol=lf
+*.textile text eol=lf
+*.txt text eol=lf
+*.xml text eol=lf
+*.yml text eol=lf
+Gemfile text eol=lf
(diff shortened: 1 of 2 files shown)


enkessler:
The changes themselves look fine. What's been happening with that `iso-8859-1` file before this fix?

olleolleolle:
@enkessler My guess is "task has not been run" so, nothing. 

In order to be more explicit, we could exclude specifically the TWO other-encoded files.

 Perhaps give them their own FileList to prep for an impl which actually CRLF-fixes those two files as well. And, I _think_ exclude can be done using the filelist objects with subtraction.

enkessler:
Out of curiosity, is there a setting that we can simply adjust so that all files get saved with LF instead of CRLF?

olleolleolle:
@enkessler There is. A gitattributes dotfile can assist with that with an auto-rule matching file patterns. 

Life-learning: not all files like being transformed. So, some file patterns to be transformed may be called for. 

Maintaining the `.gitattributes` file is less troublesome than having CRLF issues, I guess.

olleolleolle:
@enkessler I now see that GitHub's Help pages have documented this to a high degree: https://help.github.com/articles/dealing-with-line-endings/

olleolleolle:
Eh, @aslakhellesoy, did you author the original Rake task?

Would you say that the above  Git attributes solution makes the Rake task superfluous?

PR #1169, opened 2017-08-03
Handle selective strict options.

### LINKED ISSUE #1160: Cucumber --retry option status 1 with strict mode causes --retry option to not be useful
@brasmusson - re: 
https://github.com/cucumber/cucumber-ruby-core/pull/141 Tying this to `--strict` mode is overloading what `--strict` mode means and how developers are going to want to use the `--retry` option.  

Personally speaking I am trying to tie this in with CI because I don't want to have to rebuild 30+ minutes of feature files for 1 scenario that happened to randomly fail.  The only way I see that happening is by turning off strict mode since CI relies on a 0 or 1 status code to report success/failure.  However, I also don't want someone to be able to put `pending` or worse yet forget to define a step in one of their new scenarios and have CI report a "pass" because `--strict` mode was disabled, which is the whole reason I enabled `--strict` mode in CI to begin with.

Since `--retry` isn't a default option, if I pass `--retry` and `--strict`, that means I *know* there's a chance something may need to be retried but I want `--strict` enforcement of rules around undefined or pending tests.  Backing up to a higher level, since `--retry` isn't a default option, if I wanted a single "failure" to fail my whole test suite and report a non-zero status code, I would just leave off the `--retry` option all together since 1 fail is the same as 1 fail and 1 pass retry in the way it's currently designed.

If the current functionality is absolutely desired (although I honestly can't think of a case where I would want a non-zero exit code if retries were successful), maybe we could use a new option instead, something like `--retry-strict`

@mattwynne @garie @garside @gondalez @danascheider @rishi-freshbooks  - would love any thoughts on this as well.  Was just trying to get this latest code updated in my CI and couldn't figure out why it wasn't passing until I realized I put `--strict` in our cucumber.yml.

references: 
https://github.com/cucumber/cucumber-ruby/pull/1121
https://github.com/cucumber/cucumber-ruby/issues/1044
https://github.com/cucumber/cucumber-ruby-core/pull/141

brasmusson:
## Summary

Let the result types affected the the strict settings be controlled individually, so only a subset of the can be set to be handled strict.

## Details

Add the options 
* `--[no-]strict-undefined`, 
* `--[no-]strict-pending`, and
* `--[no-]strict-flaky`, 

so each of the result type affected by the strict settings be controlled individually. 
The `--[no-]strict` option is equivalent to `--[no-]strict-undefined --[no-]strict-pending --[no-]strict-flaky`. 
Also make sure that it is possible from the command line to adjust the strict setting from a profile, for instance:
Given that a profile includes the `--strict` option
When using `--no-strict-flaky` on the command line
Then only `undefined` and `pending` will be handled strict

Depends on cucumber/cucumber-core#143.

## Motivation and Context

Fixes #1160.

## How Has This Been Tested?

The automated test suite has been updated to verify this behaviour.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly (the `--help` option will describe the new options).

### DIFF
--- a/lib/cucumber/cli/configuration.rb
+++ b/lib/cucumber/cli/configuration.rb
@@ -26,7 +26,7 @@ module Cucumber
         @args = args
         @options.parse!(args)
         arrange_formats
-        raise("You can't use both --strict and --wip") if strict? && wip?
+        raise("You can't use both --strict and --wip") if strict.strict? && wip?
         set_environment_variables
       end
 
@@ -42,7 +42,7 @@ module Cucumber
         Integer(@options[:seed] || rand(0xFFFF))
       end
 
-      def strict?
+      def strict
         @options[:strict]
       end
 
--- a/lib/cucumber/cli/options.rb
+++ b/lib/cucumber/cli/options.rb
@@ -3,6 +3,7 @@ require 'cucumber/cli/profile_loader'
 require 'cucumber/formatter/ansicolor'
 require 'cucumber/glue/registry_and_more'
 require 'cucumber/project_initializer'
+require 'cucumber/core/test/result'
 
 module Cucumber
   module Cli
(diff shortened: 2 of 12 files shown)


streetlogics:
@brasmusson Thanks for this!  Plugged it in and works like a charm.

PR #1170, opened 2017-08-07
Do not by default delay the execution of test cases

### LINKED ISSUE #1167: Make the TC count event optional
It is not a good idea to always stop all test cases in a filter before passing them on, because maybe a formatter that is interested in the test case count is used (from the [cukenspace workshop](https://cucumberbdd.slack.com/archives/C5B8A9Q3H/p1498313949091649)), see also https://github.com/cucumber/cucumber-ruby/pull/1082#issuecomment-310316166

brasmusson:
## Summary

Do not by default delay the execution of test cases until all have been processed through the filter chain.

## Details

* Do not by default delay the execution of test cases until all have been processed through the filter chain.
* Change the name of the TestRunStarted event to TestCount event.
* Add the option --count-first which ensures that the TestCount event is issued before any test case is executed.

Depends on cucumber/cucumber-ruby-core#147

## Motivation and Context

See https://github.com/cucumber/cucumber-ruby/pull/1082#issuecomment-310316166.
Fixes #1167.

## How Has This Been Tested?

The automatic test suite has been updated to verify this behaviour.

## Types of changes

## Checklist:


features/docs/events/test_count_event.feature
```diff
@@ -11,31 +11,55 @@ Feature: Test Run Started Event
...
+      Given /a turtle/ do
+        puts "turtle!"
+      end
       """
     And a file named "features/support/events.rb" with:
       """
       AfterConfiguration do |config|
-        config.on_event :test_run_started do |event|
-          config.out_stream.puts "test run started"
+        config.on_event :test_count do |event|
```
mattwynne: I think I'd like this to be called `test_case_count`

WDYT?

mattwynne:
Thanks for pushing away at this @brasmusson, I'll be glad to have the code back to normally not processing all the test cases before execution starts.

Two big thoughts here:

1. Rather than having a configuration switch, could the `BroadcastTestCountEvent` filter interrogate the config to see if there are any listeners subscribed to the `test_count` event, and act as a no-op if there aren't any?

2. Wouldn't it be more useful to have an event that surfaces a complete list of all the test cases, rather than just a count?

What do you think?

brasmusson:
There are a couple of choices here:
* Optional events?
  Should a "test count" event be optional, and only issued when batching up the test cases before execution? So far we do not have any optional events. On the other hand issuing the "test count" event after the test case has executed, as currently in the PR, does not seem very useful.
* The possibility to query the event bus for handlers, or the existence of handlers, for events?
  So far the event sources are ignorant of the existence of, or the identity of any handlers for the events they issue. Since handlers can register themselves after events have started to be issued, if an event should only be issued if there are any handlers registered, then we have to consider when the latest time to register that handler and still trigger the issuing of the optional event. Should that time be at the time of the `afterConfiguration`-hook, or immediately before the issuing of the optional event?

In this RP I choose not to introduce any new concepts (optional events of querying the event bus for handlers), therefore a configuration switch is used, and the result is a change in the execution order and a reordering of events.

brasmusson:
@mattwynne I've update the PR so that the `test_case_count` event is only fired if a listener has registered itself or the event.

mattwynne:
Nice one @brasmusson!

I was thinking it would be more generally useful if it send a list of all the test case `file:line` locations. You could count that, but you could also use it to look up (and render details about) the forthcoming test cases' scenarios from Gherkin documents.

Do you think it makes sense to replace this event with that one now, or as part of another iteration?

PR #1173, opened 2017-08-10
Fix Lint/DuplicateMethods

jaysonesmith:
## Summary

Investigated the Lint/DuplicateMethods offenses and need guidance on how best to address them.

## Details

* Offense count: 2
* Added review note as I'm unsure how these should be handled.
* Review note: In the two instances of duplication, both methods are under MultilineArgument, but one self method calls another that belongs to the private Builder class.
* [Here's the code in question](https://github.com/jaysonesmith/cucumber-ruby/blob/1021-fix-style-violations-2/lib/cucumber/multiline_argument.rb#L31*L58)

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -22,11 +22,6 @@ Lint/DeprecatedClassMethods:
     - 'lib/cucumber/project_initializer.rb'
 
 # Offense count: 2
-Lint/DuplicateMethods:
-  Exclude:
-    - 'lib/cucumber/multiline_argument.rb'
-
-# Offense count: 2
 Lint/Eval:
   Exclude:
     - 'lib/cucumber/formatter/ansicolor.rb'
--- a/lib/cucumber/multiline_argument.rb
+++ b/lib/cucumber/multiline_argument.rb
@@ -11,14 +11,14 @@ module Cucumber
         builder.wrap(node)
       end
 
-      def from(argument, location=nil)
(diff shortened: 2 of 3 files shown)


mattwynne:
I think you could safely inline those two method calls on lines 32 and 36, and construct the types directly there. Leave the `Builder` to just be used for the `from_core` method.

jaysonesmith:
So, in getting more familiar with the code and realizing that builder.doc_string and builder.data_table are pretty widely used concepts, it makes removing them from the Builder class not a good idea, IMO. Here's some info on what I did, and why:

*There's only a single point that calls `MultilineArgument.doc_string` directly when running the rake tests, so I've changed it to `MultilineArgument.doc_string_builder` and modified the method outside Builder in `multiline_argument.rb` to match. 
* While there seem to be no tests that touch lines 35-37 of `multiline_argument.rb` aka `data_table` I've updated it to match the dos_string_builder naming scheme, too. This section of code potentially has a coverage issue as it seems like we're not testing arrays. _I can comment out the code called here, and rake doesn't break_
* I've also changed the #from method for both string and array argument types, so that code that utilizes #from will work accordingly too.

Running `bundle exec rake` succeeds without any errors due to this change.

Commit inc!

jaysonesmith:
Inlined the function so things should look/feel better.

PR #1177, opened 2017-08-11
Fix Style/AlignArray

jaysonesmith:
## Summary

Partially Fixing Metrics/AlignArray

## Details

* All offending directories except 'spec/cucumber/formatter/legacy_api/adapter_spec.rb' which has 1106 offenses

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -146,16 +146,6 @@ Style/Alias:
   Exclude:
     - 'lib/cucumber/formatter/ansicolor.rb'
 
-# Offense count: 1197
-# Cop supports --auto-correct.
-Style/AlignArray:
-  Exclude:
-    - 'lib/cucumber/cli/options.rb'
-    - 'lib/cucumber/rake/task.rb'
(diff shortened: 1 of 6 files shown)


jaysonesmith:
Before I get started on (the adapter_spec file)[https://github.com/cucumber/cucumber-ruby/blob/master/spec/cucumber/formatter/legacy_api/adapter_spec.rb#L114-L151], it seems like the indentation here was done for visibility, but not all arrays are done this way?

Should I still fix this?

mattwynne:
Yes the indentation on the adapter spec is done for visibility. We intend to remove that code before too long anyway, so I think we could put inline exceptions for those and leave them as they are.

jaysonesmith:
That makes sense and is doable. Thanks!

jaysonesmith:
Finished the inline exception work. This cop should be all set now!

jaysonesmith:
History.md has been updated. PR coming soon. (Including my other PRs)

Issue #1182, opened 2017-08-22
scripts/update-history generates bad links


aslakhellesoy:
2 problems:

* The Markdown is wrong (missing `]`)
* The URL to the issue is wrong (points to API)

Both the script and the `History.md` needs updating.

/cc @mattwynne @jaysonesmith

mattwynne:
Gah! That'll teach me to write untested code late at night while on whisky

jaysonesmith:
Hah! GG @mattwynne :p It should be easy enough to fix for now until the [bash version](https://github.com/cucumber/cucumber/issues/255) comes out.

jaysonesmith:
@mattwynne You fixed this in master, yeah?

mattwynne:
I did, in 25b33b0e8c6e39099bb1a9e76986fef07bd4dded

mattwynne:
Oh wait, actually I didn't fix the URL to the issue thing.

mattwynne:
links fixed in eed8b2d3e0d77dedc2dd8269b2b53b402bdd4d65

olleolleolle:
The only remaining thing is to get History.md in line with those two changes.

jaysonesmith:
History.md is no more @olleolleolle ! :p Are you referring to adding the changes that @mattwynne made to the update script as needing to be added to the changelog or something different?

olleolleolle:
@jaysonesmith The Description field mentioned the History.md as a last step; so, I guess, in order to take this to _done_ as defined in there, the changelog needs to avoid the issues, as well.

jaysonesmith:
@olleolleolle Ah, I get what you mean. Pretty sure I nabbed things. I'd like to also update usernames to point to git profiles, but perhaps that'll be something for another day.

PR #1186, opened 2017-08-24
Fix links and other updates

jaysonesmith:
Minor changes. Manual, partial fix for #1182


jaysonesmith:
Hey! So, as a heads up, I got a bit eager in making some doc updates and moved forward without being fully aware of various processes. With that said, if my merging and deleting branches too soon has caused issues and you'd like me to fix or flat out revert this PR, happy to do so as I definitely don't want to cause issues/step on toes. If you have a preferred fix method you'd like me to run, please send it my way.

Thanks and sorry if this has caused any troubles!

Check out the #committers channel in Slack for more.

PR #1197, opened 2017-09-15
Publish to Cucumber Pro (from Travis)

mattwynne:
This will automatically push our docs to Cucumber Pro after a succesful build, meaning they're always up to date.

Based on https://docs.travis-ci.com/user/deployment/custom/ and
https://docs.travis-ci.com/user/encrypting-files/

Let's see if this works...


olleolleolle:
@mattwynne Perhaps this is _exactly_ what [Build Stages](https://docs.travis-ci.com/user/build-stages/define-steps/) could be good for?

olleolleolle:
@mattwynne Also: COOL CHANGE, you!

mattwynne:
@olleolleolle I was just thinking to myself "this is going to run for all those different Ruby versions in the matrix, that's stupid". Would build stages fix that then?

olleolleolle:
@mattwynne The example they put in their thing was `TEST_SUITE=none` (being _another_ ENV variable which wouldn't add to "matrix width" (if that's a word)).

mattwynne:
Once this build has run and I've got some feedback I'll have a play with it.

maxmeyer:
Interesting. @mattwynne When this works, I'm interested to use this for aruba as well.

brasmusson:
Cucumber-JVM has for some time now used Build Stages to publish snapshot releases; [.travis.yml](https://github.com/cucumber/cucumber-jvm/blob/master/.travis.yml#L19-L65)

mattwynne:
Won't work now we're on Circle!

PR #1205, opened 2017-09-29
CircleCI build

aslakhellesoy:
The travis build currently takes over 40 minutes to give us feedback.

I set up [CircleCI](https://circleci.com/gh/cucumber/cucumber-ruby) and feedback time is now down to a couple of minutes.

As part of doing this I also simplified the `Gemfile` (changed the mechanism for using local dependencies).

The `Gemfile.lock` is now checked in, which allows CircleCI to cache gems and speed up the build. It's excluded from the built gem.

The `cucumber.gemspec` no longer runs git to list all files. The jruby docker image doesn't have git, and we shouldn't assume anyone has git. Not running git is also faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "Gemfile.lock" }}
-          - v1-dependencies-
+          - bundle-{{ checksum "Gemfile.lock" }}
       - run:
           name: install dependencies
           command: |
@@ -22,7 +21,7 @@ jobs:
       - save_cache:
           paths:
             - ./vendor/bundle
-          key: v1-dependencies-{{ checksum "Gemfile.lock" }}
+          key: bundle-{{ checksum "Gemfile.lock" }}
       - run:
           name: Run RSpec
           command: bundle exec rake spec
@@ -38,8 +37,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "Gemfile.lock" }}
-          - v1-dependencies-
+          - bundle-{{ checksum "Gemfile.lock" }}
       - run:
           name: install dependencies
           command: |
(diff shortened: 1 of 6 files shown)


mattwynne:
I'm absolutely delighted with this.

I think we might be able to replace *all* of that Gemfile stuff with this instead: http://bundler.io/v1.3/man/bundle-config.1.html#LOCAL-GIT-REPOS

mattwynne:
@aslakhellesoy I know you don't like rebase, but if we want to merge this into master it would be nice if the commits were clean. Could you rebase -i to make them tell a more tidy story?

aslakhellesoy:
It's primarily on master I think rebase should be avoided. I'll squash.

aslakhellesoy:
Not sure we can use that Gemfile local stuff because `Gemfile.lock` is checked in to speed up the CI build (using caching)

mattwynne:
Alright well we can play about with the Gemfile stuff, let's get this merged!

botandrose:
@aslakhellesoy Wow! A speed up by an order of magnitude is very significant! What is it about Circle CI that makes the full CI run so much faster than Travis?

brasmusson:
@botandrose From what I know about Travis, the Travis configuration of Cucumber-Ruby is not designed to give fast feedback. AFAIK:
* Travis will not run more than 5 jobs from a build in parallel, so with more than 5 builds jobs start to run (somewhat) sequential. The Cucumber-Ruby Travis configuration has 12 jobs.
* Travis is primarily set up to run Linux. There seems to be fewer OSX executors, so OSX-job seems to have a tendency not be be started immediately after they are created. The Cucumber-Ruby Travis configuration has 6 OSX jobs.

There seems certainly to be potential to change the Cucumber-Ruby Travis configuration to achieve faster feedback (although less comprehensive in terms of tested Ruby versions and platforms).

brasmusson:
For instance after 15 minutes all Linux jobs of https://travis-ci.org/cucumber/cucumber-ruby/builds/282345779 have finished, not it just waiting for the OSX jobs.

PR #1208, opened 2017-10-06
Review and handle excluded/ignored cops

jaysonesmith:
- Moved to rubocop.yml

## Summary

Moving style decisions we're not going to follow to the main `rubocop.yml` file improves the quality of life for those working on style changes.

## Details

This is for [1021](https://github.com/cucumber/cucumber-ruby/issues/1021) and the future of the repo's style direction.

When I started working on style issues, I started at the top of the todo file which happened to be the AmbiguousOperator cop. Unfortunately, it wasn't until after I finished the work that I found [1022](https://github.com/cucumber/cucumber-ruby/pull/1022) in which we had determined we wouldn't be following that style guide. 

This change will accomplish a few things:

1. A central area for us all to use to track styles/cops that we've decided not to abide by.
2. Remove styles we're not following from the _todo file, preventing duplicate work.
3. Prevent any styles we're not following from being wiped out when --auto-gen-config is rerun to get an updated list.

If possible, I'd move that we comment before the disabled/excluded cop a link to an issue or some documented text where each particular decision was made.

## How Has This Been Tested?

`bundle exec rake` ✅ 

## Types of changes

## Checklist:


jaysonesmith:
I'm not certain if there's a particular place we should document this direction change. Perhaps under the ('First Timer')[https://github.com/cucumber/cucumber-ruby/blob/master/CONTRIBUTING.md#first-timer-no-problem] heading?

jaysonesmith:
So it seems to make sense to ignore inline specific cops that are not going to be ignored widespread.

@mattwynne After inlining some ignore statements, I've wondered how we'll notify future style folks that we're intentionally excluding specific cops without having to dig through all the source. Since you've hopped in a little about this, do you have any input?

jaysonesmith:
Thanks for the feedback @olleolleolle! I agree it helps them understand things faster. In my opinion, I'd like to keep Rubocop exclusions in the rubocop file rather than inline. Inline, to me, buries the style choices that we've made and if a given committer hasn't worked on a file with that particular exclusion, it could prompt confusion? 

Given that @mattwynne 's mentioned and asked for inlining on a couple occasions, I'd still like to get his input/sign off. I know he's planning to, just been busy.

mattwynne:
@jaysonesmith I support your idea of moving global exclusions like the `Lint/AmbiguousOperator` out of the TODO file, and documenting them like this, good idea!

I think there are still cases where inline exclusions make sense. For example, [this inline exclusion](https://github.com/cucumber/cucumber-ruby/blob/master/spec/cucumber/formatter/legacy_api/adapter_spec.rb#L114) from the `AlignArray` rule. We have a good reason to keep it like that (it makes the test more readable) and we won't be keeping this code forever anyway.

So I think we can have both. Certainly, inline exclusions should be a last resort, and the TODO file should only contain things that are actually actionable.

Does that all make sense?

jaysonesmith:
`bundle exec rake` ✅

PR #1220, opened 2017-10-27
Fix Performance/RedundantBlockCall

jaysonesmith:
## Details

- Rubocop prefers the use of `yield` over `block.call` as it's much, much faster. It also doesn't like having `&block` in method arguments when using `yield` as it considers it unused, hence the use of `*`.
- May be a change that some don't agree with stylistically. If so, happy to revert!

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes
- [x] Rubocop style fixes


jaysonesmith:

PR #1226, opened 2017-10-28
Fix Lint/HandleExceptions

jaysonesmith:
## Details

- Rubocop doesn't like it when you block hole exceptions. Since we're not wanting to error out on the load error here, I've just added a brief warn message.

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes
- [x] Rubocop style fixes


jaysonesmith:
This one definitely needs to get squashed. Ended up basing it off of the todoregen branch erroneously and now it's a hot mess but should still be good to go overall.

jaysonesmith:

lib/cucumber/rake/task.rb
```diff
@@ -5,6 +5,7 @@
   # Support Rake > 0.8.7
   require 'rake/dsl_definition'
 rescue LoadError
+  warn 'LoadError for rake/dsl_definition'
```
olleolleolle: This is one of those place where we _want_ to silence the LoadError and not caring about it.

For ancient Rake versions before that version, this file won't exist (and won't be necessary) and Ruby will raise the LoadError. We catch that and silence that error, telling Ruby: "This is normal."

So, we want to silence RuboCop's complaining about this specific LoadError silencing.

jaysonesmith: I've pushed a new version without this. :)

PR #1227, opened 2017-10-28
Fix Lint/IneffectiveAccessModifier

jaysonesmith:
## Details

Rubocop pointed out that we had two points in our code where we had methods under the private designation that actually weren't private. It seems that for one instance we actually want it to be private and another we don't. I'll make comments below to explain some.

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes
- [x] Rubocop style fixes

### DIFF
--- a/lib/cucumber/glue/registry_and_more.rb
+++ b/lib/cucumber/glue/registry_and_more.rb
@@ -155,6 +155,14 @@ module Cucumber
         raise ArgumentError.new('Expression must be a String or Regexp')
       end
 
+      def self.cli_snippet_type_options
+        registry = CucumberExpressions::ParameterTypeRegistry.new
+        cucumber_expression_generator = CucumberExpressions::CucumberExpressionGenerator.new(registry)
+        Snippet::SNIPPET_TYPES.keys.sort_by(&:to_s).map do |type|
+          Snippet::SNIPPET_TYPES[type].cli_option_string(type, cucumber_expression_generator)
+        end
+      end
+
       private
 
       def available_step_definition_hash
@@ -167,14 +175,6 @@ module Cucumber
 
       def hooks
         @hooks ||= Hash.new{|h,k| h[k] = []}
-      end
-
-      def self.cli_snippet_type_options
-        registry = CucumberExpressions::ParameterTypeRegistry.new
-        cucumber_expression_generator = CucumberExpressions::CucumberExpressionGenerator.new(registry)
-        Snippet::SNIPPET_TYPES.keys.sort_by(&:to_s).map do |type|
-          Snippet::SNIPPET_TYPES[type].cli_option_string(type, cucumber_expression_generator)
-        end
       end
     end
 
--- a/lib/cucumber/glue/snippet.rb
+++ b/lib/cucumber/glue/snippet.rb
@@ -81,6 +81,8 @@ module Cucumber
           new(cucumber_expression_generator, 'Given', 'I have 2 cukes', MultilineArgument::None.new).step
         end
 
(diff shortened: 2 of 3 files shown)


lib/cucumber/glue/registry_and_more.rb
```diff
@@ -155,6 +155,14 @@ def create_expression(string_or_regexp)
         raise ArgumentError.new('Expression must be a String or Regexp')
       end
 
+      def self.cli_snippet_type_options
```
jaysonesmith: Moved this above the private section as there are [places in the repo we call it directly](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/options.rb#L329)

lib/cucumber/glue/snippet.rb
```diff
@@ -81,6 +81,8 @@ def self.example(cucumber_expression_generator)
           new(cucumber_expression_generator, 'Given', 'I have 2 cukes', MultilineArgument::None.new).step
         end
 
+        private_class_method :example
```
jaysonesmith: Actually made the example method private with this line. Test still pass so ideally we're all good.

jaysonesmith:

lib/cucumber/glue/snippet.rb
olleolleolle: The only usage _here_ seems to be in the CLI options-creating one-liner.

I'm not fully familiar with the `glue/` code - I can not say whether this is used anywhere else. What the code _does_ indicates that it's quite local. 

Alternative: If it truly is local, it could be made a local variable instead of its own method.

Alternative 2: Do not change the visibility of this method in this PR.

jaysonesmith: With the way it was written before, it was written under the `private` section, but self methods aren't private unless specific set private. One would assume that if it was under private that it should be private, yeah? [Info here](http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Lint/IneffectiveAccessModifier)

olleolleolle: That's a good indication, supporting "it's local".

jaysonesmith: @olleolleolle Interestingly enough, if I change step 47 to `format('%-7s: %-28s e.g. %s', type, description, "")` and comment the rest out, the tests pass still.

jaysonesmith: This makes me wonder if perhaps there's a coverage issue?

jaysonesmith: @olleolleolle Any direction given my most recent comments above? Perhaps I should just leave this one be for now and create another issue to investigate it further? /shrug

olleolleolle: Let's leave this out of this change - you're right.

olleolleolle:
@jaysonesmith Apologies for slow movement, but I'm in favor of moving carefully when we have no coverage on this thing.

lib/cucumber/glue/snippet.rb
jaysonesmith: Will do!

jaysonesmith:
@olleolleolle That's alright, I appreciate the time you've spent! Thank you!

jaysonesmith:
Also, once the above is reverted, are you good with me merging?

olleolleolle:
@jaysonesmith Yes, let's get this merged; you're improving the developer experience!

jaysonesmith:
Sounds good, thanks!

PR #1228, opened 2017-10-28
Fix Style/AccessorMethodName

jaysonesmith:
## Details

- Rubocop says don't use things like `set` before write methods. A writer method's behavior should be implied/known.

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes
- [x] Rubocop style fixes


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Changed
 
+* Fix Style/AccessorMethodName ([#1228](https://github.com/cucumber/cucumber-ruby/pull/1228) [@jaysonesmith](https://github.com/jaysonesmith))
```
brasmusson: Updating the Changelog in a pull-request will most likely lead to merge conflict (with Changelog changes in other pull-requests). AFAIK out guideline is that the person the merges a pull-request updates the Changelog, instead of the pull-request author (in the pull-request).

brasmusson: I really feel that "Changed" is the wrong heading in the Changelog for these kind of changes. I think "Changed" should mean "different behavior compared to the previous release" - think that a user need to examine carefully before starting to use a new version. Things not affecting the behavior should be placed under other headings like "Refactoring / Developer Experience" as in v3.0.0.pre.2 of "Improved" as in v3.0.0.

jaysonesmith: @brasmusson Unfortunately keepachangelog doesn't seem to indicate any way of dealing with these types of changes. Is this issue only our Ruby repo deals with or does this type of thing happen elsewhere, too? If elsewhere, should we discuss on the monorepo to draw a wider audience?

Ultimately I can definitely move things, it just puts us out of sync with the other repos. Also, once all of this style work is complete, we can enforce the rubocop style set that we end up with so that we won't have any more of these types of PRs, right?

jaysonesmith:
@brasmusson As I've been going through and updating my current PRs, I've been moving the changelog entries to the Improved section. That work for you? :)

lib/cucumber/cli/options.rb
```diff
@@ -101,19 +101,19 @@ def parse!(args) # rubocop:disable Metrics/AbcSize
 
           opts.on("#{RETRY_FLAG} ATTEMPTS", *retry_msg) {|v| set_option :retry, v.to_i }
           opts.on('--i18n-languages', *i18n_languages_msg) { list_languages_and_exit }
-          opts.on('--i18n-keywords LANG', *i18n_keywords_msg) {|lang| set_language lang }
+          opts.on('--i18n-keywords LANG', *i18n_keywords_msg) {|lang| language lang }
```
luke-hill: If these are methods with argument(s), it's actually better to call them inside parentheses. Again, not sure if you wanna fix here. There's a few examples of why this gets ambiguous with multiple args. With 1 there's no real issue, so up to you really.

jaysonesmith: Thanks for the eye @luke-hill!

jaysonesmith:

lib/cucumber/cli/options.rb
jaysonesmith: @luke-hill There are many, many of those in that file. Thank you for calling them out though! Upon further review/thought, I'm going to leave them for now for two reasons:

1. This PR is for the Rubocop requirement around [Accessor Method Names](
http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/AccessorMethodName)
2. The style decision made in [this PR.](https://github.com/cucumber/cucumber-ruby/pull/1022)

I appreciate you calling it out though! :D

CHANGELOG.md
```diff
@@ -57,6 +58,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 ### Fixed
 
 *  Fix the handling of failed hooks in the JUnit Formatter ([@brasmusson](https://github.com/brasmusson))
+>>>>>>> master
```
olleolleolle: TODO: This is a conflict marker, which should not be in the resulting file.

lib/autotest/cucumber_mixin.rb
```diff
@@ -26,7 +26,7 @@ def run
 
     loop do # ^c handler
       begin
-        get_to_green
+        to_green
```
olleolleolle: An issue with this rename: the name is not the name of a getter, but the expression "Get to green", which is lingo from the Agile community about TDD. 

Example of this lingo in use: https://blog.carbonfive.com/2011/01/12/tdd-getting-to-green-one-small-step-at-a-time/

lib/cucumber/cli/options.rb
luke-hill: Re-read #2 - Figure if people have made this decision elsewhere then there's no reason to alter it back. Good for digging up the refs though

luke-hill:
All happy with the stuff I mentioned :+1:

CHANGELOG.md
jaysonesmith: All set on this. Thanks for the eye!

lib/autotest/cucumber_mixin.rb
jaysonesmith: I've fixed this. :)

jaysonesmith:
@olleolleolle LMK if things LGTY

jaysonesmith:
Before I fix the conflict and merge, I need to do some investigation as for some reason, rubocop isn't working as expected any longer. May be a local issue.

PR #1230, opened 2017-10-29
Remove the legacy_api

### LINKED ISSUE #839: Re-write formatters against new API

brasmusson:
## Summary

Remove the legacy_api.

## Details

As soon as all formatter have been re-written to use the new formatter api (see #839), the legacy_api should be removed.

Closes #839.

## Motivation and Context

The legacy_api was only intended as a short term fix in v2.0.0.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


mattwynne:
\o/

@brasmusson how about we set the master branch to target v4.0.0 and try and just rip out all this old code, including the pretty/html formatters, then rebuild from there?

My only concern is that there may be quite a few scenarios that depend on the pretty formatter, though I could be wrong.

brasmusson:
Around half of the scenarios fail when the `pretty` and `html` formatters do not work. A bit less than half if the default formatter is changed to `progress`, otherwise more than half (137 out of 210) of the scenarios fails. So yes, a lot of our feature files depends on the `pretty` formatter.

olleolleolle:
I’m interested in this being open.

mattwynne:
💖 hooray!

PR #1233, opened 2017-11-16
Fix Junit formatter when Around hook throws an exception

conf:
## Summary

`Junit` formatter doesn't handle correctly cases when exceptions are thrown from `Around` hook.

## Details

Junit formatter incorrectly assumes that failing scenario always has a
failing step, while failure can come from a hook and all steps would be
successful.

Let's use the information from `@failing_step_source` only if it exists.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/formatter/junit.rb
+++ b/lib/cucumber/formatter/junit.rb
@@ -103,7 +103,7 @@ module Cucumber
       def create_output_string(test_case, scenario, result, row_name)
         output = "#{test_case.keyword}: #{scenario}\n\n"
         return output if result.ok?(@config.strict)
-        if test_case.keyword == 'Scenario'
+        if test_case.keyword == 'Scenario' && @failing_step_source
           output += @failing_step_source.keyword.to_s unless hook?(@failing_step_source)
           output += "#{@failing_step_source}\n"
         else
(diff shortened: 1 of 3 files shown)


conf:
Hello, guys 👋 !
I'm not sure if the fix is optimal, but at least, it works for me. I'll be glad to hear any suggestions how to improve this.
Thanks!

conf:
@brasmusson Please, take a look.

lib/cucumber/formatter/junit.rb
```diff
@@ -103,7 +103,7 @@ def end_feature(feature_data)
       def create_output_string(test_case, scenario, result, row_name)
         output = "#{test_case.keyword}: #{scenario}\n\n"
         return output if result.ok?(@config.strict)
-        if test_case.keyword == 'Scenario'
+        if test_case.keyword == 'Scenario' && @failing_step_source
```
brasmusson: Entering the else clause will print "Example row" in the report, so that is wrong, we need to enter the then clause - but not use `@failing_step_source` unless it exists.

brasmusson:
@conf Thanks for identifying the problem. Failing around hooks are the only case where `@failing_spec_source` is not set, all other hook types are represented as test steps in the test case. The proposed solution will treat scenarios with around hook failures as scenario outlines, so is does not work correctly, therefore I made a different fix in 9605e88

conf:
Thank you for looking into this! I'll try harder next time 😄

PR #1236, opened 2017-11-21
[WIP] Scenario result is unavailable in around hooks

p0deje:
## Summary

Currently, any around hook yields instance of scenario. However, after the scenario is run, this instance is not updated to reflect the result of scenario execution. This makes it impossible to run around hooks checking for scenario errors, like this:

```ruby
Around do |scenario, block|
  block.call
  if scenario.failed?
    # do something
  end
end
```

## Details

This PR merely adds a regression test demonstrating the problem. If you think that I'm right and the behavior should be fixed, I can try to implement that, though I haven't figured out how to do it so far, so any guidance is appreciated.

## Motivation and Context

I understand that such around hooks can be decomposed into before + after, but I also think that current behavior is confusing and can be fixed. For example, RSpec properly propagates example errors.

## Types of changes

## Checklist:


p0deje:
It's still a valid issue, please keep it open.

xtrasimplicity:
@p0deje It looks like test cases [trigger a  `Cucumber::Events::TestCaseFinished` event on completion](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/events/test_case_finished.rb), but the [around hook is defined in `cucumber-ruby-core`](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/around_hook.rb). 

The other hooks are defined in this project:
https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/hooks.rb

Hope this helps!

p0deje:
I still don't really understand how to propagate result to scenario.

xtrasimplicity:
So, I've made a little progress on this. It turns out, the `Around` block is _not_ called immediately prior to a step's completion. To demonstrate this, I updated the feature to print out "This text was rendered within the around block...", without any conditions.
```
[...]
And a file named "features/support/hooks.rb" with:
"""
Around do |scenario, block|
  puts "This text was rendered within the around block..."
  puts "Around hook knows about failure" if scenario.failed?
  block.call
end
"""
[...]
```
As you can see, the text is only printed once, towards the start of the step.
```
+Feature: Around hooks
+
+  Scenario: scenario is failed in hook # features/f.feature:2
+      This text was rendered within the around block...
+    When scenario fails                # features/step_definitions/steps.rb:1
+      Something went wrong within the scenario (RuntimeError)
+      ./features/step_definitions/steps.rb:2:in `/^scenario fails$/'
+      ./features/support/hooks.rb:5:in `block in <top (required)>'
+      features/f.feature:3:in `When scenario fails'
+
+Failing Scenarios:
+cucumber features/f.feature:2 # Scenario: scenario is failed in hook
+
+1 scenario (1 failed)
+1 step (1 failed)
```

Theoretically, if we execute this hook again at the end of each scenario (as with an `after` hook), this should resolve the issue. However, this will effectively turn an `around` block into an `around + after` block. @aslakhellesoy, @mattwynne Do you foresee this being an issue? Would you accept a PR to resolve this?

mattwynne:
@xtrasimplicity there are two things I would not trust here, unfortunately.

1) try using `$stderr.puts` instead of `puts`. For historical reasons, Cucumber [currently overrides `puts` in step definitions (and possibly hooks) and swallows the output](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/glue/proto_world.rb#L156) for printing later in the formatters.

2) the pretty formatter is [hideously complex](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/formatter/legacy_api/adapter.rb), again for historical reasons, and I would not trust whether it's reliably reflecting what's going on beneath the covers. I'd try using the `summary` formatter instead which is [a lot simpler](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/formatter/summary.rb).

mattwynne:
Maybe you've already seen all that and borne it in mind in your research, but I wanted to check.

xtrasimplicity:
Thanks @mattwynne, I really appreciate the input! I take it you guys would be open to merging a PR which resolves this issue (i.e. the scenario result never being returned to the `Around` hook)? Assuming it isn't an absolute mess, of course. :)

Thanks again!

mattwynne:
Absolutely, that sounds like a useful feature if you could figure out how to do it.

xtrasimplicity:
I've merged master into this branch and have taken another look this morning. 

I've tried to explicitly invoke the `around` hooks from within `Cucumber::Filters::ApplyAfterHooks#test_case`, but it looks like the around hooks aren't being registered at this point. Will need to investigate further.

```bash
$ bundle exec cucumber features/docs/writing_support_code/around_hooks.feature:269
Return value is: nil

[8, 17] in /home/xtrasimplicity/Development/Projects/cucumber-ruby/lib/cucumber/filters/apply_after_hooks.rb
    8: 
    9:         require 'byebug'
   10:         debugger
   11: 
=> 12:       end
   13:     end
   14:   end
   15: end
(byebug) hooks.find_around_hooks(test_case)
[]
```

luke-hill:
Devils Advocate Question.

What does an Around hook give you that an AfterStep hook doesn't?

cc/ @p0deje / @xtrasimplicity

p0deje:
Around runs around scenario, not step. To answer the question, it is just more convenient sometimes:

```ruby
Around do |scenario, block|
  Features::Helpers::Screencast.start
  block.call
  Features::Helpers::Screencast.discard unless scenario&.failed?
end
```

luke-hill:
Yep, that's a pretty good use case. Fair does.

xtrasimplicity:
Good news, @p0deje! I've made some changes in a local working copy and have managed to get this working. I'll need to clean it up a little and write some more specs - but we should have this sorted soon. :)

luke-hill:
Given more recent conversations indicate we're removing these, and you can achieve these using a combination of other hooks. I'll close this long-standing issue off.

PR #1271, opened 2018-01-24
Fix Multiline Violations: Braces, Blocks, Method call work, etc.

jaysonesmith:
## Details

- Layout/MultilineArrayBraceLayout
- Layout/MultilineBlockLayout
- Layout/MultilineMethodCallBraceLayout
- Layout/MultilineMethodCallIndentation
- Layout/MultilineOperationIndentation
- Metrics/ClassLength exclusion for the CLI's options also.

## Motivation and Context

## How Has This Been Tested?

`bundle exec rake` :+1:

## Types of changes
- [x] Rubocop style fixes

### DIFF
--- a/scripts/invite-collaborator
+++ b/scripts/invite-collaborator
@@ -19,7 +19,7 @@ class Team
 
   def team
     @team ||= gh.organization_teams('cucumber')
-      .find { |team| team['name'] == name } || raise("Unable to find a team named #{name}")
+                .find { |team| team['name'] == name } || raise("Unable to find a team named #{name}")
   end
 end
 
--- a/scripts/invite-collaborator
+++ b/scripts/invite-collaborator
@@ -18,8 +18,9 @@ class Team
   end
 
   def team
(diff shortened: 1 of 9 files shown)


scripts/invite-collaborator
```diff
@@ -19,7 +19,7 @@ class Team
 
   def team
     @team ||= gh.organization_teams('cucumber')
-      .find { |team| team['name'] == name } || raise("Unable to find a team named #{name}")
+                .find { |team| team['name'] == name } || raise("Unable to find a team named #{name}")
```
luke-hill: This is a preference (And it isn't in rubocop), but I'd start a newline after `||=` which means less tabbing

rubocop will not complain about :point_up:

jaysonesmith: I see what you mean. I don't know if I agree here, but I'll take a look later. :) Ultimately the way it is now satisfies Rubocop which is the goal. Thanks for the input!

jaysonesmith: Fixed it. :)

Issue #1285, opened 2018-03-26
The cucumber --format stepdefs stack trace is too harsh and doesn't provide a solution


mbigras:
## Summary

Thank you for this project!

I'm having an issue when using `cucumber --format stepdefs` when there is a typo in my step.
The cucumber --format stepdefs stack trace is too harsh and doesn't provide guidance to a solution.
Issues #953 and #955 point out a problem with using the `--format stepdefs` flag but don't clarify the problem or provide a solution.

The work around solution is:

* remove the step
* comment out the step
* make a stub definition of the step

## Main idea

```
Given I first installed cucumber and aruba
And I'm not an expert at debugging Ruby applications
And I make a typo in my step
And I want to check out the avaiable step definitions to see what my typo was
When I run `cucumber --format stepdefs`
Then the command should fail
And I shouldn't be shown a 40 line Ruby stack trace
And there should be a helpful error message that looks something like:
  """
  You can't use the `--format stepdefs` flag with undefined steps!
  Implement the steps and try again.

  You can implement step definitions for undefined steps with these snippets:

  Then("the command should pass with:") do |string|
    pending # Write code here that turns the phrase above into concrete actions
  end
  """

Or, there should be a feature to just preview all the steps.
Or, the `--dry-run` switch should work.
```

## Steps to Reproduce (for bugs)

```
gem install cucumber:3.1.0 aruba:0.14.5
mkdir cukular_bomb
cd cukular_bomb/
cucumber --init
cat <<'EOF' > features/support/env.rb
require 'aruba/cucumber'
EOF
cat <<'EOF' > features/helloruby.feature
Feature: Ruby's command line interface
  Scenario: Using the -e flag
    When I run `ruby -e 'print "hello world!"'`
    Then the command should pass with:
      """
      hello world!
      """
EOF
cucumber
cucumber --format stepdefs
cucumber --format stepdefs 2>&1 | awk 'END { print NR }'
```

## Context & Motivation

Showing a user a 40 line stack trace for making a typo in the step definition is too harsh and not helpful.

## Your Environment

```
sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1036

gem list cucumber aruba

*** LOCAL GEMS ***

cucumber (3.1.0)
cucumber-core (3.1.0, 2.0.0)
cucumber-expressions (5.0.13)
cucumber-tag_expressions (1.1.1)
cucumber-wire (0.0.1)

[...]

PR #1286, opened 2018-03-29
Fixed lengthy stacktrace when running `cucumber -f stepdefs` when steps aren't defined

xtrasimplicity:
## Summary
Currently, when running `cucumber -f stepdefs` against features which don't include step definitions, a lengthy stacktrace is shown to the user. This is thoroughly described [in issue #1285](https://github.com/cucumber/cucumber-ruby/issues/1285).

## Details
I've simply added a condition to prevent a `NoMethodError` from being raised, if a step hasn't been defined for a test case. When an error isn't raised, example snippets are printed to stdout, as shown in the following test case:

```
Scenario: Run with --format stepdefs when some steps are undefined
  Given a file named "features/calculator.feature" with:
  """
  Feature: Calculator
    Scenario: Adding numbers
      When I add 4 and 5
      Then I should get 9
  """
  When I run `cucumber -f stepdefs features/calculator.feature`
  Then it should pass with:
  """
  You can implement step definitions for undefined steps with these snippets:

  When("I add {int} and {int}") do |int, int2|
  end

  Then("I should get {int}") do |int|
  end
  """
```     

## Motivation and Context
As described in issue #1285, it can be difficult for new users to understand what went wrong, when they mis-type a step, or fail to define one, and are shown a 40 line stacktrace.

## How Has This Been Tested?
I've added an additional feature to `features/docs/formatters/usage_formatter.feature`.

## Types of changes

## Checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/cucumber/cucumber-ruby/1286)

### DIFF
--- a/features/docs/formatters/usage_formatter.feature
+++ b/features/docs/formatters/usage_formatter.feature
@@ -99,3 +99,25 @@ Feature: Usage formatter
         11 steps (11 skipped)
 
         """
+    
+    Scenario: Run with --format stepdefs when some steps are undefined
+      Given a file named "features/calculator.feature" with:
+      """
+      Feature: Calculator
+        Scenario: Adding numbers
+          When I add 4 and 5
+          Then I should get 9
+      """
+      When I run `cucumber -f stepdefs features/calculator.feature`
+      Then it should pass with:
+      """
+      You can implement step definitions for undefined steps with these snippets:
+
+      When("I add {int} and {int}") do |int, int2|
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+
+      Then("I should get {int}") do |int|
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+      """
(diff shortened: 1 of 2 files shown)


xtrasimplicity:
Thanks for reviewing these PRs, @olleolleolle! :)

mbigras:
Nice @xtrasimplicity!

Question:

Do you think the exit status should be non-zero?

I think there shouldn't be a long stack trace but the exit status should be non-zero because all the steps weren't successfully run. Also, there are probably tools make decisions about what to do next based on cucumber's exit status. Exiting 0 for an undefined step might not be what folks expect.

xtrasimplicity:
@mbigras, The fix I implemented changed the `on_step_finished` event, which I suspect may be called from other areas of the code. My changes won't have made any impact on the exit codes that cucumber returns. 

 Whether or not a test suite with undefined steps should yield a non-zero exit code is something that I would consider to be a breaking change. 

I'm also not sure whether returning a successful exit code when undefined steps are present was the desired functionality, or whether it just turned out to be that way. @olleolleolle/@mvz

mvz:
I haven't been around long enough to know the original intent, but to me it makes sense that output format and exit code are orthogonal concerns, so the exit code with `-f stepdeps` should be the same as, say `-f pretty` or `-f progress`.

Issue #1240, opened 2017-12-12
Error when html formatter tries to render error


fschwahn:
## Summary

When `Cucumber::Formatter::Html` tries to render an error where `snippet_for` goes to the `else`-branch (line 581)
https://github.com/cucumber/cucumber-ruby/blob/84bc859a62220290383a0a8c80ebf6820a426b4c/lib/cucumber/formatter/html.rb#L575-L583

it returns a frozen string (because of `frozen_string_literal: true`). When the syntax gem is not installed, modifying the frozen string in the `snippet`-method (line 571) raises an error (`can't modify frozen String`).
https://github.com/cucumber/cucumber-ruby/blob/84bc859a62220290383a0a8c80ebf6820a426b4c/lib/cucumber/formatter/html.rb#L568-L573

## Possible Solution

Either `highlighted += "..."` to copy the frozen string instead of mutating it, or `[String.new("# Couldn't get snippet for #{error_line}"), 1]` in `snippet_for` to create a mutable string from the start.

I would have submitted a PR, but couldn't find an easy way to add a regression spec for this.

## Steps to Reproduce (for bugs)

This is hard to reproduce, because I'm not sure how to reliably trigger the `else`-branch in `snippet_for`. It happened on our CI, as well as locally, and was connected with timeouts from `poltergeist`-gem.

## Your Environment

* Version used: 3.1.0 (but every version with frozen_string_literals: true is affected)

fschwahn:
This is still relevant.

sciros:
I see it consistently when hitting a particular kind of error with Watir, usually when an element is not clickable at (x,y) because of an unexpected modal dialog or some-such.
I've been able to replicate it by simply adding 
`raise(Selenium::WebDriver::Error::UnknownError.new('something something')`
into a step definition.

newtewt:
I'm am seeing it consistently as well when the button is visible but not enabled. I see the error below then the can't modify frozen string. Causes my subsequent tests to fail. Installing syntax gem is a work around to fix the error. 

Stack trace showing 
`~/.rvm/gems/ruby-2.4.1/gems/cucumber-3.1.0/lib/cucumber/formatter/html.rb:571:in 'snippet'`

`element present, but timed out after 30 seconds, waiting for #<Watir::Button: located: true; {:class=>"qa-forward-stepper", :tag_name=>"button"}> to be enabled (Watir::Exception::ObjectDisabledException)`

xtrasimplicity:
[...]

PR #1287, opened 2018-03-30
Improve HTML formatting when `frozen_string_literal` is enabled and the `Syntax` gem is not installed.

xtrasimplicity:
## Summary 
_This has been copied directly from [the original issue](https://github.com/cucumber/cucumber-ruby/issues/1240)._

When `Cucumber::Formatter::Html` tries to render an error where `snippet_for` goes to the `else`-branch (line 581)
https://github.com/cucumber/cucumber-ruby/blob/84bc859a62220290383a0a8c80ebf6820a426b4c/lib/cucumber/formatter/html.rb#L575-L583

it returns a frozen string (because of `frozen_string_literal: true`). When the syntax gem is not installed, modifying the frozen string in the `snippet`-method (line 571) raises an error (`can't modify frozen String`).
https://github.com/cucumber/cucumber-ruby/blob/84bc859a62220290383a0a8c80ebf6820a426b4c/lib/cucumber/formatter/html.rb#L568-L573

## Details

To resolve this, I simply created a new string from the original one, rather than mutating the frozen string.

## Motivation and Context
When the `@@converter` class variable (under `Cucumber::Formatter::Html::SnippetExtractor`) is assigned an instance of `Cucumber::Formatter::Html::SnippetExtractor::NullConverter` (or more simply, when the `Syntax` gem is not installed), and `frozen_string_literal` is enabled, `Cucumber::Formatter::Html::SnippetExtractor#snippet` raises an error when the `gem install syntax to get syntax highlighting` comment is appended to the error message.

This issue is thoroughly defined in issue #1240.

## How Has This Been Tested?
I've added a Spec to `spec/cucumber/formatter/html_spec.rb`, to test `Cucumber::Formatter::Html::SnippetExtractor#snippet`

## Types of changes

## Checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/cucumber/cucumber-ruby/1287)

### DIFF
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -566,10 +566,7 @@ module Cucumber
           raw_code, line = snippet_for(error[0])
           highlighted = @@converter.convert(raw_code, false)
 
-          if @@converter.is_a? NullConverter
-            highlighted = String.new(highlighted) if highlighted.frozen? # Ensure that a "can't modify frozen String" RuntimeError isn't raised when frozen_string_literal is enabled.
-            highlighted << "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>"
-          end
+          highlighted += "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>" if @@converter.is_a? NullConverter
 
           post_process(highlighted, line)
         end
--- a/lib/cucumber/formatter/html.rb
+++ b/lib/cucumber/formatter/html.rb
@@ -565,7 +565,12 @@ module Cucumber
         def snippet(error)
           raw_code, line = snippet_for(error[0])
           highlighted = @@converter.convert(raw_code, false)
-          highlighted << "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>" if @@converter.is_a?(NullConverter)
+
+          if @@converter.is_a? NullConverter
+            highlighted = String.new(highlighted) # Ensure that a "can't modify frozen String" RuntimeError isn't raised when frozen_string_literal is enabled.
+            highlighted << "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>"
+          end
+
           post_process(highlighted, line)
         end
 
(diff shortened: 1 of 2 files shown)


olleolleolle:
This is a helpful bugfix.

lib/cucumber/formatter/html.rb
```diff
@@ -565,7 +565,9 @@ class NullConverter; def convert(code, _pre); code; end; end #:nodoc:
         def snippet(error)
           raw_code, line = snippet_for(error[0])
           highlighted = @@converter.convert(raw_code, false)
-          highlighted << "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>" if @@converter.is_a?(NullConverter)
+
+          highlighted += "\n<span class=\"comment\"># gem install syntax to get syntax highlighting</span>" if @@converter.is_a? NullConverter
```
olleolleolle: Question: Can change be made smaller by not changing the use of parentheses? Does that change anything in warnings output or otherwise?

spec/cucumber/formatter/html_spec.rb
```diff
@@ -567,5 +567,24 @@ module Formatter
...
+        let(:instance) { Html::SnippetExtractor.new }
+
+        context 'when the error string is frozen' do
+          before { Html::SnippetExtractor.class_variable_set(:@@converter, Html::SnippetExtractor::NullConverter.new) }
+
+          context 'the error line doesnt end with a colon and a line number' do
+            let(:error_line) { 'path/to/file'.freeze }
+            let(:error) { [error_line, nil] }
+
+            it do
```
olleolleolle: Perhaps this `it` could be described, so that a rspec output using “-f documentation” has more complete sentences?

xtrasimplicity: Yep, I'll flesh this out. Thanks!

lib/cucumber/formatter/html.rb
xtrasimplicity: Yep - I'll do this, too. It has no impact on anything at all. :)

PR #1293, opened 2018-04-26
 Add read access to task_name of Cucumber::Rake::Task

entwanderer:
## Summary

When running tests, you may need to store the name of the task in the log file to make reproducing failures easier.

## Details

So with this commit we give read access to task_name.

## Motivation and Context

I run tests with different profiles on jenkins. For reporting and logging i needed to store the task name, to have a mapping to the task with which the test run.

## How Has This Been Tested?

Added test to retrieve the task_name.

I logged the task_name in my project successfully.

## Screenshots (if appropriate):
-

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] I've added tests for my code

### DIFF
--- a/spec/cucumber/rake/task_spec.rb
+++ b/spec/cucumber/rake/task_spec.rb
@@ -7,6 +7,15 @@ require 'rake'
 module Cucumber
   module Rake
     describe Task do
+      describe '@task_name' do
+        context 'has read access to task name' do
+          it { expect(subject.respond_to? :task_name).to be true }
+        end
+        context 'has no write access to task name' do
+          it { expect(subject.respond_to? "#{:task_name}=").to be false }
+        end
+      end
+
       describe '#cucumber_opts' do
         before { subject.cucumber_opts = opts }
 
--- a/lib/cucumber/rake/task.rb
+++ b/lib/cucumber/rake/task.rb
@@ -133,6 +133,9 @@ module Cucumber
       #
       # Note that this attribute has no effect if you don't run in forked mode.
       attr_accessor :bundler
+      
+      # Name of the running task
+      attr_reader :task_name
 
       # Define Cucumber Rake task
       def initialize(task_name = 'cucumber', desc = 'Run Cucumber features')


entwanderer:
I use it to define the name of the report file names:
```ruby
Cucumber::Rake::Task.new(:my_rake_task) do |t|
  t.profile = 'my_profile'
  t.cucumber_opts += ["--format html","--out","'out/report/#{Env.iso_timestamp}_#{t.task_name}.html'"]
end
```

I use it for logging the used task in `env.rb`:
```ruby
MyLogger.log "task", Config.task_name
```

I use it for naming screenshots, taken in `hooks.rb`:
```ruby
After('@browser-test') do |scenario|
  if scenario.failed?
    screenshot_filename = "out/report/screenshots/#{Env.iso_timestamp}_#{Config.task_name}.png"
    @browser.save_screenshot(screenshot_filename)
    embed(screenshot_filename, "image/png")
  end
  @browser.close
end
```

`Rake::Task` provides the `#name` as well, where I'm already get used to read it.

entwanderer:
@olleolleolle 
Only as long as we use namespaces the string way, e.g.:
```ruby
Cucumber::Rake::Task.new('my_namespace:my_task)'
```

`Cucumber::Rake::Task` does not consider namespaces that way in its task_name:
```ruby
namespace :my_namespace do
  Cucumber::Rake::Task.new(:my_task)
    # something to do
  end
end
```

If we would provide this feature as well, we may think about renaming `@task_name` to `name` or at least provide a `#name` method, returning `@task_name` to use the same method name in `Cucumber::Rake::Task` and `Rake::Task`

As `Cucumber::Rake::Task` does only support a fraction of `Rake::Task`, it may also be worth thinking about only providing `@task_name` without namespace support. But providing namespaces would be a nice to have for sure.

spec/cucumber/rake/task_spec.rb
```diff
@@ -7,6 +7,15 @@
 module Cucumber
   module Rake
     describe Task do
+      describe '@task_name' do
+        context 'has read access to task name' do
+          it { expect(subject.respond_to? :task_name).to be true }
```
luke-hill: this can be slightly simplified to `expect(subject).to respond_to(:task_name)` And the one below can be `.not_to`

spec/cucumber/rake/task_spec.rb
```diff
@@ -7,6 +7,15 @@
 module Cucumber
   module Rake
     describe Task do
+      describe '@task_name' do
```
luke-hill: As this would be an instance method, tradition dictates this would probably be `#task_name` (Possibly just a typo on your behalf)

xtrasimplicity:
Bump.

luke-hill:
I'll pick this up and fix it. As it's been around a while and it's a tiny amount of fixes needed

PR #1298, opened 2018-05-26
Use single quotes for snippets

### LINKED ISSUE #1291: Wrong step definition created when having "("
_From @christianbaumann on April 19, 2018 9:43_

## Summary

I have a step definition which contains "(" and ")", eg. 
> And the room information (room code and availability) is correct

## Expected Behavior

When I run cucumber and have no step definition defined, I would expect a matching step definition to be defined
> And(/^the room information \(room code and availability\) is correct$/) do

## Current Behavior

Instead, following step definition is created:
> And("the room information \(room code and availability) is correct") do

## Your Environment

* Ruby: ruby 2.3.3p222 (2016-11-21 revision 56859) [x64-mingw32]
* Cucumber:
cucumber (3.1.0)
cucumber-core (3.1.0)
cucumber-expressions (5.0.17, 5.0.15)
cucumber-tag_expressions (1.1.1)
cucumber-wire (0.0.1)
* Operating System and version: Windows 8.1 Enterprise

_Copied from original issue: cucumber/cucumber#364_

danascheider:
This PR addresses #1291 by changing snippets to use single quotes rather than double quotes.

## Details

Previously, snippets looked like this:
```
Given("a user named {string}") do
  ...
end
```
Now, this snippet would look like this:
```
Given('a user named {string}') do
  ...
end
```

## Motivation and Context

This PR fixes #1291, in which parentheses and brackets are not escaped correctly using double quotes. Following @aslakhellesoy's suggestion, instead of adding an extra escape character, I simply changed the snippets to use single quotes instead of double

## How Has This Been Tested?

All Cucumber features and RSpec examples have been updated to incorporate the single quotes. I also added a Cucumber scenario testing that the characters are escaped correctly.

## Types of changes

## Checklist:


luke-hill:
Seems legit: Only thing I'd mention is this line probably needs rewriting to specify in a bit more detail what the unit test does. But it's not a red/green line. So seems cool from me :+1:

xtrasimplicity:
The documentation PR has been merged, so we should probably merge this one, too. Perhaps the test case should be reviewed and another PR submitted, as per @luke-hill's feedback?

lib/cucumber/glue/snippet.rb
```diff
@@ -88,7 +88,7 @@ def typed_pattern
         def to_s
           header = generated_expressions.each_with_index.map do |expr, i|
             prefix = i == 0 ? '' : '# '
-            "#{prefix}#{code_keyword}(\"#{expr.source}\") do#{parameters(expr)}"
+            "#{prefix}#{code_keyword}('#{expr.source}') do#{parameters(expr)}"
```
aslakhellesoy: Have you considered the case where `expr.source` contains a single quote?

features/docs/defining_steps/snippets.feature
```diff
@@ -49,8 +49,31 @@ Feature: Snippets
...
+    When I run `cucumber features/snippets.feature`
+    Then it should pass with:
+      """
+      You can implement step definitions for undefined steps with these snippets:
+
+      Given('a step with \(parentheses)') do
+        pending # Write code here that turns the phrase above into concrete actions
+      end
+
+      Given('another step with \{brackets}') do
```
aslakhellesoy: I think it's called curly braces, or just braces

lib/cucumber/glue/snippet.rb
aslakhellesoy: Or, if we keep the double-quoted strings, the case where `expr.source` contains a `\` character.

danascheider: Ooh! Good point. I actually agree with everything you’re saying and would also prefer the double quotes to stay. I just misread your other comments and took them to mean you preferred this way, and deferred to that even though I didn’t agree 😂

danascheider: I’m making myself a note to submit new documentation changes tomorrow morning and change this PR right away.

danascheider:
@aslakhellesoy It looks to me like implementing the escape will involve changes to Cucumber Expressions since `expr.source` might contain params in curly braces. I might be able to find a workaround for this using `expr.parameter_names` but it might be easier to add the escape character in Cucumber expressions before the params are interpolated. WDYT?

aslakhellesoy:
Bump :-)

danascheider:
Bump. This is still on my radar, I'm just moving to Australia next Sunday so things have been hectic.

aslakhellesoy:
Take your time @danascheider - good luck with the move. We'll keep this one waiting for you [down under](https://github.com/cucumber/cucumber/blob/2558787723fb18f41e5d69624d1a7f44079685be/gherkin/gherkin-languages.json#L828-L869).

danascheider:
@aslakhellesoy I'm closing this PR since we decided on a different implementation. I'll open a new one when I have that implementation fleshed out more fully.

PR #1321, opened 2018-10-05
Remove deprecated method

orien:
## Summary

Fix two tests that were failing.

## Details

The test suite was failing with:

```
  1) Cucumber::Formatter::Interceptor::Pipe#unwrap! disables the pipe bypass
     Failure/Error: raise "This method is due for removal after version #{remove_after_version}" if Cucumber::VERSION > remove_after_version

     RuntimeError:
       This method is due for removal after version 3.99
     # ./lib/cucumber/deprecate.rb:20:in `call'
     # ./lib/cucumber/deprecate.rb:28:in `deprecate'
     # ./lib/cucumber/formatter/interceptor.rb:24:in `buffer'
     # ./spec/cucumber/formatter/interceptor_spec.rb:88:in `block (3 levels) in <module:Formatter>'

  2) Cucumber::Formatter::Interceptor::Pipe#write adds the buffer to its stored output
     Failure/Error: raise "This method is due for removal after version #{remove_after_version}" if Cucumber::VERSION > remove_after_version

     RuntimeError:
       This method is due for removal after version 3.99
     # ./lib/cucumber/deprecate.rb:20:in `call'
     # ./lib/cucumber/deprecate.rb:28:in `deprecate'
     # ./lib/cucumber/formatter/interceptor.rb:24:in `buffer'
     # ./spec/cucumber/formatter/interceptor_spec.rb:112:in `block (3 levels) in <module:Formatter>'
```

The method `Cucumber::Formatter::Interceptor::Pipe#buffer` has been scheduled for removal before version 4.0.0 is released.

I removed this method and migrated tests to use the replacement method: `Cucumber::Formatter::Interceptor::Pipe#buffer_string`.

## Motivation and Context

This moves us one step closer to having a passing CI build.

## How Has This Been Tested?

The method has been removed, allowing tests to pass.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/formatter/interceptor_spec.rb
+++ b/spec/cucumber/formatter/interceptor_spec.rb
@@ -5,13 +5,7 @@ require 'cucumber/formatter/interceptor'
 
 module Cucumber::Formatter
   describe Interceptor::Pipe do
-    let(:pipe) do
-      pipe = double('original pipe')
-
-      allow(pipe).to receive(:instance_of?) { true }
-
-      pipe
-    end
+    let(:pipe) { instance_spy(IO) }
 
     describe '#wrap!' do
       it 'raises an ArgumentError if its not passed :stderr/:stdout' do
@@ -56,6 +50,7 @@ module Cucumber::Formatter
     describe '#unwrap!' do
       before :each do
         @stdout = $stdout
+        $stdout = pipe
         @wrapped = Interceptor::Pipe.wrap(:stdout)
       end
 
@@ -84,10 +79,9 @@ module Cucumber::Formatter
         buffer = '(::)'
         Interceptor::Pipe.unwrap! :stdout
 
-        expect(@wrapped).to receive(:write).with(buffer)
-        expect(@wrapped.buffer).not_to receive(:<<)
-
         @wrapped.write(buffer)
+
+        expect(@wrapped.buffer_string).not_to end_with(buffer)
       end
 
       after :each do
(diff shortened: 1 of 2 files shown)


spec/cucumber/formatter/interceptor_spec.rb
```diff
@@ -20,7 +14,7 @@ module Cucumber::Formatter
 
       context 'when passed :stderr' do
         before :each do
-          @stderr = $stdout
+          @stderr = $stderr
```
orien: We should save `$stderr` so we can put it back how we found it after the spec run.

spec/cucumber/formatter/interceptor_spec.rb
```diff
@@ -5,13 +5,7 @@
...
 module Cucumber::Formatter
   describe Interceptor::Pipe do
-    let(:pipe) do
-      pipe = double('original pipe')
-
-      allow(pipe).to receive(:instance_of?) { true }
-
-      pipe
-    end
+    let(:pipe) { instance_spy(IO) }
```
orien: We can use a simple instance spy here.

spec/cucumber/formatter/interceptor_spec.rb
luke-hill: We don't need an each declaration when we're inside an area with one test. But that's probably splitting hairs.

jaysonesmith:
@enkessler Looks like the Windows builds are failing on some formatting stuff for the /branch build. The /pr build looks to have jumped the gun on the exe's being available or something along those lines? I can't retry the latter, so I'm not sure if it'd end up failing for the same reason as the /branch one.

Any idea here?

orien:
The Windows build failures are affecting the master branch also.
https://ci.appveyor.com/project/enkessler/cucumber-ruby-u5iqh/build/job/dxauxu8fnx24vcrq

The problem is not introduced by this PR.

orien:
Is anyone opposed to me merging this in?

jaysonesmith:
Go ahead @orien ! Thanks for the help!

orien:
Thanks 😄

Issue #1324, opened 2018-11-05
--require not being honoured with --dry-run


dpsi:
## Summary

Cucumber does not honour the -r flag, and Cucumber provides no way load *.env.rb files during dry runs. From what I understand, cucumber ignores env files due to interactions with Rails. Why this option is not configurable is a mystery as not every Ruby project is a Rails project. 

The --require option has the following help text: (emphasis mine)
> Require files before executing the features. If this option is not specified, all *.rb files that are siblings or below the features will be loaded auto-matically. Automatic loading is disabled when this option is specified, and **all loading becomes explicit**. Files under directories named "support" are always loaded first.

I am not sure how much more explicit I can be. Sure I could add another rb file and require my env file, but that brings up 2 more problems:
1) I need to then keep track of every env file and its location in the directory structure.
2) Cucumber uses load instead of require, so I will get spammed with re-declaration warnings. 

## Expected Behavior

Given I start cucumber is started with the following:
`cucumber -d -r features/support/env.rb`
Then env.rb should be loaded.

## Current Behavior
Cucumber does not honour the -r option: 
~~~
cucumber -d -r features/support/env.rb
Code:

Features:
  * features/alarms.feature
  * features/.... etc.
~~~

## Possible Solution
Cucumber should honour the -r option.
Cucumber should allow a configurable option for the env.rb loading behaviour for dry runs.
If neither of the above is acceptable, then Cucumber should use requires instead of loads as per #1043 as a work around.

## Steps to Reproduce (for bugs)

1. Run cucumber with the dry run option, and the -r option set to an env.rb file
2. See cucumber not loading the file.

## Context & Motivation
I just want to do a dry run so I can find unused steps. I don't see why Rails specific behaviour should affect non-Rails usage.
< Providing context helps us come up with a solution that is most useful in the real world>
Cucumber should be a general purpose BDD testing suite, not a Rails specific one. 
## Your Environment

* Version used: 3.1.2
* Operating System and version: Ubuntu 14.04 x64

## Other notes

In general the loading of files by cucumber is not very well defined. We know that files in support folders are loaded first, but beyond that it is a mystery. I have had issues with tests randomly starting to fail because cucumber decided to load xyz.rb before env.rb. I can't require env.rb because of #1043

dpsi:
Here is an example: https://github.com/dpsi/cucumber-test

```
~/cucumber-test$ cucumber
You should see this when env.rb is loaded
Feature: This feature should pass

[...]

PR #1326, opened 2018-11-07
Update code to include env files in dry run

danascheider:
## Summary

Causes Cucumber to load `env.rb` files even on dry run.

## Details

This PR removes a conditional in `configuration.rb` that prevents env files from loading when the `--dry-run` option is turned on. It also updates the help message for `--dry-run` to omit the part about env files not loading.

## Motivation and Context

In #1324, @dpsi brings up an issue they're having loading constants and libraries when Cucumber runs under the `--dry-run` option, pointing out that this is [not the first time](https://groups.google.com/d/msg/cukes/Kucz1T_VBss/VkEUBpY6MQAJ) users have brought up this problem with us. The issue stems from the `env.rb` files not being loaded.

## How Has This Been Tested?

The test ensuring that env files are not loaded on dry runs has been removed. I've ensured that env files do still load first in all cases.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/cli/options.rb
+++ b/lib/cucumber/cli/options.rb
@@ -189,10 +189,7 @@ Specify SEED to reproduce the shuffling from a previous run.
       end
 
       def dry_run_msg
-        [
-          'Invokes formatters without executing the steps.',
-          'This also omits the loading of your support/env.rb file if it exists.'
-        ]
+        ['Invokes formatters without executing the steps.']
       end
 
       def exclude_msg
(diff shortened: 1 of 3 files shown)


danascheider:
@mattwynne I also just noticed that CircleCI sent me an e-mail saying my access was revoked. Do you know why that could be?

aslakhellesoy:
Please hold off on merging this until we’ve agreed on the purpose of a dry run. I have reopened #1324 - let’s continue the discussion there.

PR #1329, opened 2018-12-07
Fix randomization seed

deivid-rodriguez:
## Summary

While debugging some apparently order dependent failures on my cucumber scenarios, I noticed that it's currently impossible to reproduce order dependent failures, because the seed cucumber says it has used is not the seed it has really used.

## Details

My changes make sure that the seed that cucumber uses to randomize the scenarios is the same that cucumber informs in the UI, so the same scenario order can actually be reproduced in subsequent runs.

In order to test this, I needed to use some stuff from newer versions of `aruba`, so I upgraded the scenarios to use that first. Happy to extract that to a separate PR, of course.

## Motivation and Context

This change makes it possible to reproduce order dependent failures when running scenarios in random order. Previously, it was not working.

## How Has This Been Tested?

I added a test that runs cucumber in random order and with summary format, then runs cucumber again with the seed the previous command informed, and then checks that both commands give the same output (excluding runtime).

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 
+* Fix seed printed in cucumber UI to match the seed that was actually used.
+  ([#1329](https://github.com/cucumber/cucumber-ruby/pull/1329)
+   [deivid-rodriguez](https://github.com/deivid-rodriguez))
+
 ### Added
 
 ### Improved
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I splitted this PR into #1330, #1331 and #1332. Once those PRs are merged and reviewed, I'll rebase this one :+1:.

deivid-rodriguez:
I rebased this PR that should be fully focused now and I added a changelog entry for it.

deivid-rodriguez:
It's the information about the running time, something like "0m1.234s", which obviously might differ for both runs.

deivid-rodriguez:
Ok, going in! Thanks for your review @luke-hill!

PR #1339, opened 2019-01-07
Feature/add task name to rake

luke-hill:
Re-do of stale PR #1293 

cc/ @xtrasimplicity for review/merge


luke-hill:
Top 2 failures can be ignored (2.6 stuff), the Appveyor failure looks legit though. But I've no idea how this would have caused it. It's 3 feature based tests are giving different lines for different cukes.

luke-hill:
What do you want to do @olleolleolle ??

luke-hill:
Given the discussions around JRuby and getting this fixed now being a non-priority I'm going to merge this and suggest we put jruby on the accepted failures list in travis.

Issue #1043, opened 2016-10-03
Use `require` to load support files instead of `load`


roberts1000:
I've run into a situation where I would like to control the load order for files in the `features/support` folder.  

I got the impression, from reading The Cucumber Book, that Cucumber would behave similar to RSpec and only autoload `features/support/`.  I tried moving the files into `features/setup/` and manually requiring the files from `env.rb`.  I quickly discovered that Cucumber wants to take over and load everything, regardless of the sub folder.  (To be precise, RSpec doesn't autoload files in `spec/support` unless it's specifically configured to do so.)

Next I tried moving all the files back into the `features/support` folder, but having File A explicitly require File B t to make sure File B is loaded before File A is executed.  Unfortunately, this doesn't work either because Cucumber uses the `load` method to load the support files.  **Is there any reason Cucumber couldn't use `require` instead of `load` in this method?**

https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/rb_support/rb_language.rb#L107

That would give us much more flexibility in controlling the order of how support files are loaded.  `require` would guarantee the file is only loaded once.  It wouldn't matter if Cucumber got to it first or I manually required it somewhere.

As it stands now, my only options are to combine the two dependent files into a single file, OR, move files outside of the feature folders; neither of which are good options.  The use of `load` blocks any attempt to establish order because it will blindly load the file even if it has already been loaded.

mattwynne:
I don't believe there's a good reason for using `load`, no. Probably just an accident of history. I'll give it some thought.

I'm not near the code but do all the tests pass if you change it?

roberts1000:
It looks like there are about 70 failures.

mattwynne:
Ah interesting. The test failures seem to mostly be down to [this](https://github.com/cucumber/cucumber-ruby/blob/master/features/lib/support/env.rb#L11) which allows all Cucumber's own tests to be run in the same Ruby process. This is for performance - it's much slower to run the test suite when we work with files and processes for each scenario.

So Cucumber itself might work fine, but our tests are relying on the use of `load`, I think. We could build a new [strategy for the Aruba process](https://github.com/cucumber/aruba/tree/master/lib/aruba/processes) that uses `fork` which would be faster than the spawning strategy, but that might be too much of a yak-shave.

What do you think @roberts1000? Are you keen to dig any deeper?

roberts1000:
So you're saying there's a chance.... :)

[...]

PR #1349, opened 2019-05-03
Spike/altered loading strategy

luke-hill:
## Summary

Provide a way for cucumber to only load files once.

## Details

Using a config switch, decide on whether cucumber should recurse through files potentially `n` times, or just once only.

## Motivation and Context

This fixes an old problem here: https://github.com/cucumber/cucumber-ruby/issues/1043 - Where the user had procedural code which was being instantiated twice. A user should be able to specify the load order of a file, and "not" have the file loaded twice.

## How Has This Been Tested?

Unit Tests have been added in the appropriate file.

## Types of changes

## Checklist:

Outstanding questions FYI: @mattwynne / @roberts1000
- Should we use the top level config or the config level which requires flags.
- How should we advertise this?
- Is there anything missing?


luke-hill:
I've coded this in two slightly different ways in different SHA's. Will remove the one / both we like/dislike

jaysonesmith:
WRT to your questions:
* I think something like a --load-once flag could be suitable for invoking this.
* If you're down for writing a blog post about the solution and how you did it, we could advertise it that way. :)
* I saw that in Slack you said you were unsure if it was an issue for other languages: Is there any way you could find out? @mpkorstanje do you happen to know off-hand if this is something Java could benefit from as well?

mpkorstanje:
No. We have nothing comparable in Java.

alexJunger:
Is there a reason for keeping `load`, given that this will make it to the new major (I assume)?
If configurable, I think the non-default compatibility option should be `load`, not `require`.

luke-hill:
I think that whilst this will be in a new major cut, there was no notice of deprecations beforehand, and this is quite a big change to the autoloader. I would rather keep the default as is and add a new config for v4 and then for v5 alter / remove / switch the default

mattwynne:
Looks good to me @luke-hill and I second @jaysonesmith's suggestion of a blog post when we do the release. Just make sure the changelog is up to date when you merge so you/we know what to build the blog post around.

luke-hill:
As an update the following is required

A rebase (10min)
A review of my blog post (30min)
An update to the blog post, make it less like a German newspaper (1-2hours)
A post on slack/linkedin promoting the changes
A cut for a cucumber `4.0.0.rc2` - @mattwynne how would I go about this.
A changelog update (5mins)

luke-hill:
Couple of reasons
- It felt right based on the original situation - It is something thats quite global to cucumber running (for ruby)
- Using a `Configuration` object doesn't allow you set it in time, it kind of avoids it
- I didn't try an `AfterConfiguration` block, but I have a feeling if it's similar to the Configuration objects it won't work (Guessing by the name)

luke-hill:
Code wise this is all good now.

Issues outstanding
- Tweak to blog from old keyword to new keyboard
- Final review of blog
- Altering the link in the changelog here

cc/ @mattwynne / @aslakhellesoy

luke-hill:
I'm going to merge this in now. And as/when we cut a version or release my blog post, we can quickly amend the changelog link

PR #1350, opened 2019-05-03
Remove cucumber pro publish job

luke-hill:
## Details

Removing a failing task from circleCI

## Motivation and Context

It's long been failing, and was added a while back: https://github.com/cucumber/cucumber-ruby/commit/24e3a2ec659a02ee7797375178de19eac7e41d0f

In the SHA @mattwynne you mention adding an SSH key. Was this done? As it keeps failing for invalid authentication

Just trying to tidy up the build process with @olleolleolle to get everything green and trustworthy again

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,12 +5,6 @@
 version: 2.0
 
 jobs:
-  publish:
-    machine: true
-    steps:
-      - checkout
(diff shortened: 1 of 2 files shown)


luke-hill:
@mattwynne Are we ok to merge this?

mattwynne:
For sure! It would be good to set up publishing to Jam but we don't need to do it like this anymore.

Issue #1066, opened 2016-12-23
Report Formatter Json embed files missing labels 


vveliev:
## Summary
When using json formatter label of the embed file names are ignored, is there any practical reason to do so !? 

## Expected Behavior
```embeddings[{"mime_type":"image/png","lable":"Login_page_HTML","data":"iVBORw0KGgoAAAANSUhEUgAAB2wAAAPdCAYAAABGI5gD"}]```

## Current Behavior
```_label``` ignored
## Possible Solution
```rb
module Cucumber
  module Formatter
    class Json
      def embed(src, mime_type, label)
        if File.file?(src)
          content = File.open(src, 'rb') { |f| f.read }
          data = encode64(content)
        else
          if mime_type =~ /;base64$/
            mime_type = mime_type[0..-8]
            data = src
          else
            data = encode64(src)
          end
        end
        test_step_embeddings << { mime_type: mime_type,label: label, data: data }
      end
    end
  end
end
```

badobernard94:
That would be nice

PR #1352, opened 2019-05-14
Add support for embed labels on json report formatter

magynhard:
Add support for embed labels on json report formatter

## Details

- Added support for embed labels in json formatter
- Extended rspec test
- Make label optional

## Motivation and Context

I want to add labels to my embeddings, as i can do in other formatters.
So I'm able to categorize several string based embeddings.

#1066

## How Has This Been Tested?

Rspec tests have been extended.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/formatter/json_spec.rb
+++ b/spec/cucumber/formatter/json_spec.rb
@@ -533,7 +533,7 @@ module Cucumber
           FEATURE
 
           define_steps do
-            Given(/^there are bananas$/) { embed('YWJj', 'mime-type;base64') }
+            Given(/^there are bananas$/) { embed('YWJj', 'mime-type;base64', 'embed_label') }
           end
 
           it 'includes the data from the step in the json data' do
@@ -556,7 +556,7 @@ module Cucumber
                       "name": "there are bananas",
                       "line": 4,
                       "embeddings": [{"mime_type": "mime-type",
-                                      "data": "YWJj"}],
+                                      "data": "YWJj", "label": "embed_label"}],
                       "match": {"location": "spec/cucumber/formatter/json_spec.rb:536"},
                       "result": {"status": "passed",
                                  "duration": 1}}]}]}]))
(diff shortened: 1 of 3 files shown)


spec/cucumber/formatter/json_spec.rb
```diff
@@ -556,7 +556,7 @@ module Formatter
                       "name": "there are bananas",
                       "line": 4,
                       "embeddings": [{"mime_type": "mime-type",
-                                      "data": "YWJj"}],
+                                      "data": "YWJj", "label": "embed_label"}],
```
luke-hill: minor nit, can we keep formatting consistent, stick this on a newline

aslakhellesoy:
Let's keep this moving. It's also relevant to cucumber/cucumber#642.

@magynhard - does it look like the proposed [Attachment](https://github.com/cucumber/cucumber/blob/master/cucumber-messages/messages.md#io.cucumber.messages.Attachment) message will cover your use case? Do you think you could refactor your code to use the new protocol messages? It's the way going forward. Happy to answer any questions you might have.

aslakhellesoy:
Would it make sense to add an `Attachment#meta` field? What should its properties be? A `label` field could be useful, but what about other fields such as `browser`, or `hostname`. Do we try to come up with some standardised fields, or do we leave this to the application layer, which could use it to hold a JSON document with application-specific fields. That begs a couple of questions: Do we want components to exchange implementation-specific fields? What are the benefits and drawbacks? Could we let this evolve gradually, based on pull by consumers? It would be easy to release backward and forward compatible revisions of the protocol to accommodate for new fields.

magynhard:
@aslakhellesoy as the proposed attachment message lets me include custom information of every kind, it seems to cover my use case. I was suprised, that there hasn't been implemented a feature like that already as cucumber is no young project.

A additional label field, used for report generators, might be useful.
Otherwise some type information could be added to the content type MIME
`content/type;some-additional-information`

So translations of the appropiate labels might map with that information using it just as a key of its dictionary.

I used that strategy to workaround my initial problem here so far.

PR #1354, opened 2019-06-15
Support embeddings stored in step results

Sigill:
## Summary

Support embeddings listed in step results.

## Motivation and Context

I'm looking into cucumber-cpp for my company.
We need to be able to embed text or images into the reports, but that feature is not available when using the wire protocol.

This patch is the 1st of a series a 4, respectively impacting _cucumber-ruby_, _cucumber-ruby-core_, _cucumber-ruby-wire_ and _cucumber-cpp_.

Basically, the idea is to allow embeddings to be specified in the success or fail responses sent through the wire protocol. E.g.:
```
["success",{"embeddings":[{"label":"Embedded text","mime_type":"text/plain","src":"Some text"}]}]
["fail",{"embeddings":[{"label":"Embedded image","mime_type":"image/png;base64","src":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="}],"message":"There was no  banana"}]
```

The embeddings will then be stored in the step Result in order to be accessible by the formatter.

The patch on _cucumber-ruby_ (cucumber/cucumber-ruby#1354) will let formatters access embeddings stored in a step result.
The patch on _cucumber-ruby-core_ (cucumber/cucumber-ruby-core#170) will store in the step result the embeddings returned by the step invocation.
The patch on _cucumber-ruby-wire_ (cucumber/cucumber-ruby-wire#20) will allow the embeddings stored in the wire response to be returned by the wire-based step invocation process.
The patch on _cucumber-cpp_ (cucumber/cucumber-cpp#223) will allow C++ steps to send embeddings through wire responses.

Note: embeddings were not considered to be useful for pending steps, but support should be possible if required.

## Details

In the current implementation, it's not possible to access data from the wire response as nothing is returned from the wire invocation when a step succeed (a step is considered successful by default, an exception needs to be raised in case of failing or pending step).

**Successful step**
```
Cucumber::Core::Test::Runner::RunningTestCase::Status::Base.execute
  Cucumber::Core::Test::Step.execute
    Cucumber::Core::Test::Action.execute
      block -> Cucumber::StepMatch.invoke
        Cucumber::Wire::StepDefinition.invoke
          Cucumber::Wire::Protocol.invoke
            Cucumber::Wire::Protocol::Requests::Invoke.execute
              Cucumber::Wire::RequestHandler.execute
                Cucumber::Wire::Connection.call_remote
                  Cucumber::Wire::DataPacket reponse = // parsed packet
                  Cucumber::Wire::DataPacket.handle_with
                    Cucumber::Wire::DataPacket.handle_success // noop
                  // response discarded
      Cucumber::Core::Test::Action.passed -> return Cucumber::Core::Test::Result::Passed
  Cucumber::Core::Test::Result::Passed.describe_to(Cucumber::Core::Test::Runner::RunningTestCase)
    Cucumber::Core::Test::Runner::RunningTestCase.passed
      Cucumber::Core::Test::Runner::RunningTestCase@status = Cucumber::Core::Test::Runner::RunningTestCase::Status::Passing
```

**Failing step**
```
Cucumber::Core::Test::Runner::RunningTestCase::Status::Base.execute
  Cucumber::Core::Test::Step.execute
    Cucumber::Core::Test::Action.execute
      block -> Cucumber::StepMatch.invoke
        Cucumber::Wire::StepDefinition.invoke
          Cucumber::Wire::Protocol.invoke
            Cucumber::Wire::Protocol::Requests::Invoke.execute
              Cucumber::Wire::RequestHandler.execute
                Cucumber::Wire::Connection.call_remote
                  Cucumber::Wire::DataPacket reponse = // parsed packet
                  Cucumber::Wire::DataPacket.handle_with
                    Cucumber::Wire::RequestHandler.handle_fail -> throw Cucumber::Wire::Exception // error message (from response) embedded in exception
                  // response discarded
      Rescue in Cucumber::Core::Test::Action.execute
        Cucumber::Core::Test::Action.failed -> return Cucumber::Core::Test::Result::Failed
  Cucumber::Core::Test::Result::Failed.describe_to(Cucumber::Core::Test::Runner::RunningTestCase)
    Cucumber::Core::Test::Runner::RunningTestCase.failed
      Cucumber::Core::Test::Runner::RunningTestCase@status = Cucumber::Core::Test::Runner::RunningTestCase::Status::Failing
```

This patch allow two things:
- Visitors of `Result` to be notified of embeddings (if any).
- The json formatter (who does not use the visitor pattern) to read embeddings (if any) in Result.

Those changes should be backward compatible.

## How Has This Been Tested?

No test in that repository, but I just discovered there is a `FakeWireServer`, I need to see if I can do something with it. Suggestions are welcomed.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/formatter/json.rb
+++ b/lib/cucumber/formatter/json.rb
@@ -190,6 +190,7 @@ module Cucumber
       def add_match_and_result(test_step, result)
         @step_or_hook_hash[:match] = create_match_hash(test_step, result)
         @step_or_hook_hash[:result] = create_result_hash(result)
+        result.embeddings.each { |e| embed(e['src'], e['mime_type'], e['label']) } if result.respond_to?(:embeddings)
       end
 
       def add_failed_around_hook(result)
--- a/lib/cucumber/formatter/duration_extractor.rb
+++ b/lib/cucumber/formatter/duration_extractor.rb
@@ -24,6 +24,8 @@ module Cucumber
       def duration(duration, *)
         duration.tap { |dur| @result_duration = dur.nanoseconds / 10**9.0 }
       end
+
+      def embed(*) end
     end
   end
 end
--- a/lib/cucumber/formatter/junit.rb
+++ b/lib/cucumber/formatter/junit.rb
@@ -238,6 +238,8 @@ module Cucumber
       def duration(duration, *)
         duration.tap { |dur| @test_case_duration = dur.nanoseconds / 10**9.0 }
       end
+
+      def embed(*) end
     end
   end
 end


lib/cucumber/formatter/json.rb
```diff
@@ -190,6 +190,7 @@ def create_data_table_value(data_table)
       def add_match_and_result(test_step, result)
         @step_or_hook_hash[:match] = create_match_hash(test_step, result)
         @step_or_hook_hash[:result] = create_result_hash(result)
+        result.embeddings.each { |e| embed(e['src'], e['mime_type'], e['label']) } if result.respond_to?(:embeddings)
```
luke-hill: This is insanely nested and I may be barking up the wrong tree here. But do we care about the return value of this method.

Would there be any harm / change in code if we moved this method to in between the other 2.

I know this one looks like it's mutating each result to have new data attached to it (So the return in theory doesn't matter right?)

Sigill: Agreed, this formatter might benefit from being a visitor for results.

Swapping the lines should be possible, but I don't think we need to: `add_match_and_result()` is only called once and its return value is ignored.

Also, I guess any issue should have been catched by the tests, the json formatter looks well covered.

luke-hill:
@Sigill in what order do we need to deal with these things. Is this something you would like merged in now? Or does it rely on the other items.

Sigill:
@luke-hill The patches must be integrated in the following order:
- _cucumber-ruby_ (cucumber/cucumber-ruby#1354)
This patch allow the json formatter to use embeddings stored in `Passed`/`Failed` step results. Nothing happen if the `embeddings` field is not present (it is added by the 2nd patch).
- _cucumber-ruby-core_ (cucumber/cucumber-ruby-core#170)
This patch allow whatever is invoking a step to return an object (`PassedInvokeResult`, `FailedInvokeResult`) describing the result of the invocation (status, embeddings), and `Action` to create the resulting `Passed`/`Failed` object. The previous logic is still supported.
- _cucumber-ruby-wire_ (cucumber/cucumber-ruby-wire#20)
This patch make the wire protocol request handler responsible of step invocations return a `PassedInvokeResult`/`FailedInvokeResult`. This change support Wire json responses not containing an `embeddings` field (added by the 4th patch).
- _cucumber-cpp_ (cucumber/cucumber-cpp#223)
This patch adds embeddings to the Wire json response.

Each patch is supposed to be backward compatible with respect to the dependencies, we should therefore be able to integrate them independently (but in that order).

The feature is complete with those patches. I just don't have any test for this patch (I havn't yet looked at `FakeWireServer` to see if it could help me about that). Suggestions are welcome.

Otherwise, they are ready to be merged.

luke-hill:
Perfect answer. I'll get on with getting the first patch in in the next few days.

lib/cucumber/formatter/json.rb
luke-hill: So I've re-looked at this and have 2 new points (Again might be redundant).

From the method signature we are adding these result cases to the iVar `@step_or_hook_hash`

Now when we attach the result into 191 / 192, it is the result that is passed in. However here you are mutating the result (Altering an embeddings property).

I think it would be more prudent to move it up to 191 mutate it, then pass this new object into wherever it goes.

2nd point, can we alter the temp value of the block. Is e the embedding or the result.

I appreciate doing `embedding.embed` sounds a bit bad, but I think with something quite deeply nested having more information is better than less information

Sigill: I'm not sure to get your 1st point. Are you suggesting to write something like:
`@step_or_hook_hash[:embeddings] = create_embeddings(result)`

luke-hill: Point 1 is you're attaching metadata to something, but after it's been attached somewhere else.

I think you should re-order it so it attaches metadata, then attaches it somewhere else.

Sigill: `embed()` pushes embeddings to `@step_or_hook_hash[:embeddings]`, it's not mutating `@step_or_hook_hash[:result]`.
I guess those two structures are required, because when testing Ruby code (no Wire protocol), when a step calls `embed()`, it is directly forwarded to the formatters' `embed()` method, filling
`@step_or_hook_hash[:embeddings]` before the call to `add_match_and_result()` (which will occur when the step ends).

luke-hill:
Merged this in as I can follow the flow as described and as you've explained.

aslakhellesoy:
Hi @Sigill,

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.

Aslak Hellesøy
Creator of Cucumber

PR #1358, opened 2019-07-31
Feature/remove ruby 2 2

luke-hill:
## Summary

Remove any Ruby 2.2 logic / tests from repo

## Details

As written above

## Motivation and Context

It's been EOL for ages and no cucumber gem supports 2.2 anywhere

## How Has This Been Tested?

CI

## Types of changes

## Checklist:


luke-hill:
2 key review points
- I've not added tests for 2.6 yet, there's a branch in progress from Aslak that I may hijack after merging this in
- Rubocop has been bumped from `0.58.x` to `0.64.x` - This is sufficient a bump to be meaningful whilst not forcing people extremely high up the tool-chain. But we probably should look to bump this up another bunch sooner rather than later.

luke-hill:
Will merge this in as it's passed CI.

As aslak is off on his summer holidays I may have some time in a week or so to begin having a look at the Ruby 2.6 stuff. But it did seem as though some additional stuff was needed. But no idea.

PR #1360, opened 2019-08-15
Gherkin 7

aslakhellesoy:
Upgrade to gherkin 7


CHANGELOG.md
```diff
@@ -15,15 +15,22 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 ### Added
 
 ### Changed
-* Going forward the minimum ruby version for all cucumber based gems is 2.3
-([luke-hill](https://github.com/luke-hill))
+* Going forward the minimum ruby version for all cucumber based gems is 2.4
```
luke-hill: Is this strictly true for all cucumber gems (Are we going to make the minimum supported ver 2.4?

I don't mind either way, but I didn't feel comfortable making that decision as I know a reasonable amount of people use 2.3 and I'm still new-ish here.

I'm definitely all for it, as it makes support 10* easier. But just wanted to confirm this was for *all* cucumber gems, not just the main one.

cucumber.gemspec
```diff
@@ -18,34 +18,34 @@ Gem::Specification.new do |s|
...
 
   # For maintainer scripts
-  s.add_development_dependency 'octokit'
+  s.add_development_dependency 'octokit', '~> 4.14', '>= 4.14.0'
 
   # Needed for examples (rake examples)
-  s.add_development_dependency 'capybara', '>= 2.1'
-  s.add_development_dependency 'rack-test', '>= 0.6.1'
-  s.add_development_dependency 'sinatra', '>= 1.3.2'
+  s.add_development_dependency 'capybara', '~> 3.28', '>= 3.28.0'
```
luke-hill: this is a sizeable bump. Are we sure we're happy bumping everything immediately to their latest versions

aslakhellesoy: Yes we are, as long as it doesn't break the build.

It looks like none of the `/examples` run in CI. Nor can I run them locally. I'll see if I can make them work. If not we might as well remove the capybara dependency

CHANGELOG.md
aslakhellesoy: I reverted to 2.3

PR #1367, opened 2019-10-16
Gherkin 8

vincent-psarga:
## Summary

Preparation for release 4.0
Remove `dots-formatter` as this uses the Go implementation

This PR goes alongside:
 - [cucumber-ruby-core#178](https://github.com/cucumber/cucumber-ruby-core/pull/178)
 - [cucumber-ruby-wire#23](https://github.com/cucumber/cucumber-ruby-wire/pull/23)

## Types of changes

## Checklist:


vincent-psarga:
> I agree, but the purpose here is just to illustrate/test the syntax, not necessarily to give a good example of use.

Well, if we can do both at once, I'd say why not :)

It makes me think, I'm wondering about refactoring the existing features inside the `features/docs` folder so it uses the new syntax.
It can be pretty confusing for the users to have features using the `Scenario` keyword and the other one using the `Example` one. Which one is correct ? (well, both of them but still, being consistent would be great).
Same thing for the `Scenario Outline`.

PR #1368, opened 2019-10-17
Refactor Circle CI config

vincent-psarga:
## Summary

 - less copy/paste for the commands
 - allow failure for JRuby (same as for cucumber-core)
 - add test with latest Ruby (allow failure here too)

## Checklist:

Not applicable

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,22 @@ jobs:
 
 workflows:
   version: 2
+
+  nightly-build:
+    triggers:
+      - schedule:
+          cron: "0 5 * * *"
+          filters:
+            branches:
+              only:
+                - master
+
+    jobs:
+      - build-ruby-2_3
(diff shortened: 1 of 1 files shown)


vincent-psarga:
That's the same Rubocop errors I'm getting on the `gherkin-8` branch (that's fixed there), but I don't see them on `master` ...

I there some dark magic somewhere or is it just due to the fat that the build on `master` has not run since almost a month ?

luke-hill:
It could be due to the fact the versions of rubocop were different when they ran.

Rubocop is still in 0.x phase, so they do break it quite a bit.

vincent-psarga:
Okay that explains it :)

vincent-psarga:
The rubocop issues should be fixed by #1369
I think I'll add a nightly build here too (as for the monorepo) to avoid this kind of surprises in the next Rubocop update ;)

PR #1370, opened 2019-10-18
Build on Windows with CircleCI

vincent-psarga:
## Summary

This is a try for building cucumber-ruby on Windows using CircleCI.

## Checklist:


luke-hill:
Hi @vincent-psarga 

I hate to add extra weight (And it can be left until after), but once we've got this working, we can remove appveyor as it'll still be causing issues I imagine.

If there's any tests you need me to look at I have a bit of free time this week. I know a few of them were flaky on windows. But not sure if you're at that phase yet

vincent-psarga:
@luke-hill no worries adding extra weight ;)
Yes that's the plan, if we got build working on CircleCI, we'll drop AppVeyor (for this repo and the other ones using Windows, I'm thinking `cucumber-ruby-core` for example). This way we'll have a single CI for everything.

So far it's not an issue with tests themselves (not yet at least :D ) but with the underlying tooling (I can't get `rikd` installed and it's necessary to get the `bundle` running).
We're in touch with the CircleCI team to see if that can be fixed.

Gemfile
```diff
@@ -13,9 +13,18 @@ end
...
 gem 'cucumber-expressions', path: ENV['CUCUMBER_EXPRESSIONS_RUBY'] if ENV['CUCUMBER_EXPRESSIONS_RUBY']
 gem 'cucumber-messages', path: ENV['CUCUMBER_MESSAGES_RUBY'] if ENV['CUCUMBER_MESSAGES_RUBY']
 gem 'gherkin', path: ENV['GHERKIN_RUBY'] if ENV['GHERKIN_RUBY']
+
+require 'rbconfig'
+# rubocop:disable Style/DoubleNegation
+is_windows = !!(RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/)
+# rubocop:enable Style/DoubleNegation
+
+install_if -> { !is_windows } do
```
luke-hill: does rubocop not like building on windows? I wasn't aware of any issues when I developed on windows using it.

vincent-psarga: Rubocop relies on the [`json` gem](https://rubygems.org/gems/json/versions/2.0.3) (not the built-in one) and it causes issues when running `bundle install` on the Windows host of Circle CI (seem to work fine on AppVeyor that said).

As our goal is to have a tool as simple as possible to install, we're doing some cleanup in the gems to ensure we don't run into issues with gems that use native code.

In the case of Rubocop, as the rules are already executed 5 times (one for each supported Ruby), I think we can skip in on Windows :D

luke-hill: I've no problem with that at all. But where does it rely on `json` I checked the deps and my local installs and I can't trace it.

Am I being an idiot?

aslakhellesoy: @vincent-psarga wasn't the problem with `rubocop` rather that it depends on `jaro_winkler`, which requires a C compiler in order to install?

aslakhellesoy: There was a PR to remove it - https://github.com/rubocop-hq/rubocop/pull/7673 but it's been rejected/postponed.

vincent-psarga: yes indeed, my bad, got mixed between the various dependence failing to build there.

luke-hill: All makes sense now, I have some experience of these with `curb` another native gem. No probs my side then

vincent-psarga:
Failing Scenarios:
 - [ ] cucumber features/docs/api/list_step_defs_as_json.feature:12 # Scenario: Two Ruby step definitions, in the same file
 - [ ] cucumber features/docs/api/list_step_defs_as_json.feature:39 # Scenario: Non-default directory structure
 - [ ] cucumber features/docs/api/run_cli_main_with_existing_runtime.feature:9 # Scenario: Run a single feature
 - [ ] cucumber features/docs/cli/randomize.feature:110 # Scenario: Run scenarios randomized with some skipped
 - [ ] cucumber features/docs/cli/retry_failing_tests.feature:73 # Scenario: Flaky scenarios gives exit code zero in non-strict mode
 - [ ] cucumber features/docs/defining_steps/ambiguous_steps.feature:13 # Scenario: Ambiguous steps
 - [ ] cucumber features/docs/defining_steps/ambiguous_steps.feature:57 # Scenario: Ambiguous steps with guess mode
 - [ ] cucumber features/docs/defining_steps/nested_steps.feature:16 # Scenario: Use #steps to call several steps at once
 - [ ] cucumber features/docs/defining_steps/nested_steps.feature:36 # Scenario: Use #step to call a single step
 - [ ] cucumber features/docs/defining_steps/nested_steps.feature:111 # Scenario: Backtrace doesn't skip nested steps
 - [ ] cucumber features/docs/defining_steps/nested_steps_i18n.feature:18 # Scenario: Use #steps to call several steps at once
 - [ ] cucumber features/docs/events/gherkin_source_read_event.feature:9 # Scenario: Read two documents
 - [ ] cucumber features/docs/exception_in_around_hook.feature:16 # Scenario: Exception before the test case is run
 - [ ] cucumber features/docs/exception_in_before_hook.feature:18 # Scenario: Handle Exception in standard scenario step and carry on
 - [ ] cucumber features/docs/formatters/rerun_formatter.feature:75 # Scenario: Exit code is not zero, regular scenario
 - [ ] cucumber features/docs/formatters/summary_formatter.feature:10 # Scenario: A couple of scenarios
 - [ ] cucumber features/docs/formatters/usage_formatter.feature:34 # Scenario: Run with --format usage
 - [ ] cucumber features/docs/formatters/usage_formatter.feature:60 # Scenario: Run with --expand --format usage
 - [ ] cucumber features/docs/getting_started.feature:7 # Scenario: Run Cucumber in an empty directory
 - [ ] cucumber features/docs/getting_started.feature:16 # Scenario: Accidentally run Cucumber in a folder with Ruby files in it.
 - [ ] cucumber features/docs/gherkin/expand_option_for_outlines.feature:8 # Scenario: 
 - [ ] cucumber features/docs/gherkin/outlines.feature:76 # Scenario: Run scenario with filtering on the scenario outline name
 - [ ] cucumber features/docs/gherkin/using_descriptions.feature:19 # Scenario: Everything with a description
 - [ ] cucumber features/docs/writing_support_code/after_step_hooks.feature:16 # Scenario: Access Test Step object in AfterStep Block
 - [ ] cucumber features/docs/writing_support_code/hook_order.feature:50 # Scenario: Around hooks cover background steps
 - [ ] cucumber features/docs/writing_support_code/hook_order.feature:58 # Scenario: All hooks execute in expected order

vincent-psarga:
So, current status: I tried to add `--retry 4` when running the tests and bumped (in an ugly way ...) `Aruba` to 1.0 on Windows ([see build result](https://app.circleci.com/jobs/github/cucumber/cucumber-ruby/4822))

The idea now is to first fix those 6 scenarios and then the flaky ones.
The issue with Aruba 1.0 is that it apparently does not support Ruby 2.3 (which EOL was [march 2019](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/)).

luke-hill:
Cucumber 2.4 - Jun 2016 - Min version 1.9
Cucumber 3.0 - Sep 2017 - Min version 1.9
Cucumber 3.1.2 - Jul 2018 - Min version 2.2
Cucumber 4.rc2 - Oct 2019 - Min version 2.3
Cucumber 4(full) - Mid 2020 - Realistically could use 2.4 as a minimum? It would also tidy up the gemspec code as in ruby 2.4 we introduced `.match?` boolean

vincent-psarga:
I was more thinking about deprecating Ruby 2.3 for 4.0 and not supporting it anymore on 5.0

That said, I tried again with retry and Aruba 0.14, there's only 1 scenario fixed by this upgrade (and one more failing on Ruby/Linux). So not sure it's that interesting.

vincent-psarga:
So, the tests are now passing (with `--retry 4 --no-strict-flaky` set). I'm not really fan of that but that's better than nothing.

I'll keep investigating to see why the clean-up of `tmp/aruba` is not always done properly (when we look at the failed test, they generally come from the fact that there's extra data in the output).
I'll also try to see if it's possible that the `stdout` is not properly reset between tests.

The PR will be squashed to avoid bringing in a shitload of commits that were just there for testing on Windows...

luke-hill:
Top tekkers vincent. If you're able to as well possibly see if we can remove the appveyor checks here now we're building on circle using windows. As they're redundant.

vincent-psarga:
>If you're able to as well possibly see if we can remove the appveyor checks here now we're building on circle using windows. As they're redundant.

I thought removing `appveyor.yml` would be sufficient but apparently not. I'll see about that once we merge.

luke-hill:
Yeh I removed that in a PR ages ago (Possibly linked to this), and it did nothing.

vincent-psarga:
Note to future-me: one thing to investigate, some files can't be removed as there's still a process using it.
It might be worth adding an `After` hook killing all processes started by aruba. This may even remove the need for the `Before` hook.

vincent-psarga:
After some discussions this morning with @aslakhellesoy, we agreed on merging this and we will consider the `--retry 8` to be good enough.

The various investigations shown that the problem comes more from `aruba` on Windows (no issues on OSX/Linux) and data which are not cleaned up in `tmp/aruba`.

vincent-psarga:
Note: I also removed AppVeyor web hooks (a bit too late that said) but we should not see it anymore failing in the future.

PR #1373, opened 2019-10-30
Rename obsolete rubocop directive

vincent-psarga:
## Summary

Nightly build has failed today with a new version of Rubocop. This should fix the error.

## Types of changes

## Checklist:

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -26,7 +26,7 @@ Layout/EmptyLineAfterGuardClause:
 
 # Offense count: 5
 # Cop supports --auto-correct.
-Lint/UnneededCopDisableDirective:
+Lint/RedundantCopDisableDirective:
   Exclude:
     - 'lib/cucumber/errors.rb'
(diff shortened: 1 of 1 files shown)


luke-hill:
This is actually highlighting an issue with our new updating shell script.

We shouldn't ever lock rubocop to `x.y` it should always be locked to `x.y.z` and then manually updated and checked.

Because its at a `0.x` release. Items are often backwards incompatible. I suggest alongside the work you've done in this PR, we lock rubocop to an `x.y.0` release. This then also gives us a good stable state to try fix up the cop issues in the todo file (Not sure what's in it)

luke-hill:
Also on a side-note. The last time rubocop was ran and tested against was 0.64 (Which was quite a long time ago). So it's likely the autogen file is also severely outdated.

We should probably live on an old(er), version for the time being until we fix stuff up.

vincent-psarga:
I'll (hopefully soon) push a commit in which Rubocop version will be pinned (0.75.1).
This should avoid this kind of issue and we'll let Dependabot try to update it to newer version.

PR #1374, opened 2019-11-04
Protobuf formatter

vincent-psarga:
## Summary

Allow cucumber-ruby to produce `protocol-buffer`  messages that can be used with our stand-alone formatters.

## Types of changes

## Checklist:


lib/cucumber/glue/hook.rb
```diff
@@ -1,14 +1,17 @@
...
 
 module Cucumber
   module Glue
     # TODO: Kill pointless wrapper for Before, After and AfterStep hooks with fire
     class Hook
-      attr_reader :tag_expressions, :location
+      attr_reader :id, :tag_expressions, :location
 
       def initialize(registry, tag_expressions, proc)
+        @id = SecureRandom.uuid
```
aslakhellesoy: We should use `IdGenerator` here, so we can plug in the predictable one. This will be necessary for the cucumber compatibility kit, which will compare generated and expected ndjson.

I think it's best to pass the `id` through the constructor, and let the caller delegate to the `IdGenerator` to generate it. The generator type should be configurable via a `--predictable-ids` command-line option.

lib/cucumber/glue/step_definition.rb
```diff
@@ -62,16 +65,40 @@ def parse_target_proc_from(options)
         end
       end
 
-      attr_reader :expression, :registry
+      attr_reader :id, :expression, :registry
 
       def initialize(registry, expression, proc)
         raise 'No regexp' if expression.is_a?(Regexp)
+        @id = SecureRandom.uuid
```
aslakhellesoy: Pass `id` via constructor

vincent-psarga:
Most of this have been passed as smaller PRs, can be merged now.

PR #1377, opened 2019-12-04
Remove JRuby from supported platforms

PragTob:
Saw that in the latest commit e34971024b850541 it was stated that JRuby was no longer supported (which is sad) but I guess then it also shouldn't be part of the supported platforms.

Thanks for cucumber!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for info on contributing to Cucumber.
 * Ruby 2.5
 * Ruby 2.4
 * Ruby 2.3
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
We might be able to support JRuby if https://github.com/cucumber/cucumber/pull/813 is completed and merged.

Let’s hold off on this PR and give it some time.

PragTob:
Might be obsolete now that https://github.com/cucumber/cucumber/pull/813 is merged?

aslakhellesoy:
Obsolete! We can support JRuby :-)

Issue #1362, opened 2019-09-24
Deprecate "calling steps from step definitions" functionality


aslakhellesoy:
## Summary

Calling steps from step definitions is one of the features I regret having added to Cucumber. It's like a poor man's implementation of subroutines (which all languages support natively) with the following drawbacks:

* Bad stacktraces
* Editor/IDE cannot spot errors
* Refactoring is painful

Using this functionality should print:

```
Calling steps from step definitions is deprecated and will be removed in the next major version. For a better alternative, follow this link: XXX
```

The XXX link should point to a page in the documentation explaining in more details how to use fiunctions/methods with Cucumber. It can be based on the explanation I gave in https://github.com/cucumber/cucumber-js/issues/1232

marnen:
@luke-hill:

> Use ruby's own `send` method if you want to metaprogram or program on the fly.

> Use Helpers or Classes for specific pieces of reusable code

I’m quite familiar with both these techniques, but I don’t see how either would be useful in the case I described. Do you?

> Use case / if logic to decide based on what text which class / helper to instantiate and use

The Gherkin parser already does pretty much that, and I would rather reuse it than reimplement it. This idea seems like a step backwards to me. 

If you need the problem stated in a more generic form: I have a modifier that I would like to be able to apply to any arbitrary Gherkin step from within the scenario. I only see one maintainable way to do that, and it’s this:

```ruby
Then /^(.+) with modifier$/ do |step_text|
  with_modifier { step step_text }
end
```

How else would you propose to implement this, without maintaining a separate table of step definitions? I don’t understand how I can do this if `step` is removed, and I’d appreciate specific ideas, not just generic suggestions.

luke-hill:
`Helpers::ReUsableSteps.step_one`
`Helpers::ReUsableSteps.step_two`
`Helpers::ReUsableSteps.step_three`

That's a primitive one that would allow you to call them. If they were module functions mixed into the World.
You could then call your steps based on the `step_text` so you just `send` call them.

Also I would advocate having one "mega-step", that does 10 different things, and actually have 10 small steps, even if portions are re-used. It makes debugging easier. And in-fact most of our open source stuff we're un-coupling steps, i.e. removing the areas that do 2 different things for 2 different steps that do 1 thing.

for your example you could do.

```
Then /^(.+) with modifier$/ do |step_text|
  with_modifier { Helpers::ReUsableSteps.send(step_text) }
end
```

marnen:
[...]

PR #1383, opened 2020-01-13
Build on JRuby again

vincent-psarga:
## Summary

Thanks to the generation of Protobuffers in pure Ruby, we should be able to build cucumber on JRuby again.

## Types of changes

## Checklist:


olleolleolle:
(I made edits to the title and description, to account for the J of JRuby.)

vincent-psarga:
> (I made edits to the title and description, to account for the J of JRuby.)

Thanks, I did that PR a bit too quickly :/

luke-hill:
This PR also lends itself to perhaps (either alongside or simultaneously), removing appveyor and getting circleci to build windows. That's where the failures are most prevalent, JRuby and windows.

vincent-psarga:
Note: I disabled one of the scenario to make this work ([`features/docs/defining_steps/nested_steps.feature:111`](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/defining_steps/nested_steps.feature#L111) which involves using `step`.
As the long term idea is to disable nesting `steps`(see #1362), spending time fixing this seems counter productive. That said, if someone knows how to get this scenario pass again, feel free :)

I also had to tweak a bit one test. If seems that Ruby does not set `RbConfig::CONFIG['rubyarchdir']` which is used by `BacktraceFilter`. It seemed logical that the unit test also checked if this was set.

vincent-psarga:
> This PR also lends itself to perhaps (either alongside or simultaneously), removing appveyor and getting circleci to build windows. That's where the failures are most prevalent, JRuby and windows.

Getting the build running on CircleCI/Windows (and removing AppVeyor) is planned yes, but not in this PR.
The idea here is simply to activate JRuby again (as the `cucumber-messages` gem now uses a pure Ruby Protobuffers handling).

luke-hill:
It's failing for russian letters being converted somewhere. 

The good news is all of the main scenarios pass. It's the language specific ones that come up smelly fishy

vincent-psarga:
Ok, I've spent a fair amount of time on trying to figure out why this fails but don't get anywhere. If anyone has a deeper knowledge of JRuby and/or why this may happen, have fun :)

The only way I add to get it working was to replace content of `examples/i18n/ru/step_definitions/calculator_steps.rb` with:

```ruby
Given('я ввожу число {int}') do |number|
  calc.push number
end

Given('затем ввожу число {int}') do |number|
  calc.push number
end

When('нажимаю {string}') do |operation|
  calc.send operation
end

When('я нажимаю {string}') do |operation|
  calc.send operation
end

Then('результатом должно быть число {float}') do |expected|
  calc.result.should == expected
end

Given('я сложил {int} и {int}') do |first_number, second_number|
  calc.push first_number
  calc.push second_number
  calc.send '+'
end
```

But that's definitely cheating :/

vincent-psarga:
Oh and if someone want to figure that out, bumping to `9.2.9` does not help (the same error is there)

vincent-psarga:
Ok, this I've disabled a few examples for i18n and JRuby (namely 'ru', 'uk' and 'uz') as we can't define step definitions there (note that Greek or Japanese work).

I think it's more an issue with JRuby than with Cucumber. I was thinking about adding a file under 'docs/jruby-limitations.md' (which would be linked from the README) telling that we can not use keywords as step definitions for those languages (in the ruby code, the features files still support those)

Would that be enough to merge this PR ?

docs/jruby-limitations.md
```diff
@@ -0,0 +1,40 @@
+# Cucumber and JRuby limitations
+
+`cucumber` can be executed on `JRuby` (tested with `9.2.8.0`), allthough some of the features
```
olleolleolle:
```suggestion
`cucumber` can be executed on `JRuby` (tested with `9.2.8.0`), although some of the features
```

docs/jruby-limitations.md
```diff
@@ -0,0 +1,40 @@
...
+are not available on this platform.
+
+## Defining steps with native languages:
```
olleolleolle:
```suggestion
## Defining steps with native languages
```

vincent-psarga:
Thanks for the corrections :)
(and I'm scared to be able to do one click commits from Github interface :D )

examples/i18n/Rakefile
```diff
@@ -4,29 +4,45 @@ task :cucumber do
...
 
 task default: :cucumber
 
+def examples_disabled?(lang)
+  return make_warning("SKIPPING #{lang} (The examples are out of date - please help update them)") unless examples_working?(lang)
+  return make_warning("SKIPPING #{lang}: examples have been disabled for JRuby.") if jruby_disabled_examples?(lang)
+end
+
+def jruby_disabled_examples?(lang)
+  return unless RUBY_PLATFORM == 'java'
```
luke-hill: rubocop will flag this. Linespace after guard clause.

README.md
```diff
@@ -28,7 +28,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for info on contributing to Cucumber.
 * Ruby 2.5
 * Ruby 2.4
 * Ruby 2.3
-* JRuby 9.1
+* JRuby 9.1 (with [some limitations](https://github.com/cucumber/cucumber-ruby/blob/master/docs/jruby-limitations.md))
```
olleolleolle: Perhaps "JRuby 9.1, 9.2"?

```suggestion
* JRuby 9.1, JRuby 9.2 (with [some limitations](https://github.com/cucumber/cucumber-ruby/blob/master/docs/jruby-limitations.md))
```

vincent-psarga: It may be worth checking it still build with JRuby 9.1, but that's just a task to add to CircleCI, should not be much of a trouble.
I'll add that :)

docs/jruby-limitations.md
```diff
@@ -1,6 +1,6 @@
 # Cucumber and JRuby limitations
 
-`cucumber` can be executed on `JRuby` (tested with `9.2.8.0`), although some of the features
+`cucumber` can be executed on `JRuby` (tested with `9.1`and `9.2`), although some of the features
```
olleolleolle:
```suggestion
`cucumber` can be executed on `JRuby` (tested with `9.1` and `9.2`), although some of the features
```

vincent-psarga: well, it seems we have even more problems with JRuby 9.1 (apparently [it's failing with language `he`](https://app.circleci.com/jobs/github/cucumber/cucumber-ruby/4384)).
I guess it's safer to just indicate `9.2`

vincent-psarga:
> Is there a reason we document 9.2 but run CI against 9.1?? I don't use JRuby so there may be an obvious reason

I bumped to 9.2 but forgot to update the documentation (which is done now).
No better reason than human error :D

PR #1389, opened 2020-02-19
Deprecate json formatter

vincent-psarga:
## Summary

As we now have a `message`formatter working and that has been tested with the new stand-alone `json-formatter`, we can deprecate the old json-formatter.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/deprecate_spec.rb
+++ b/spec/cucumber/deprecate_spec.rb
@@ -4,13 +4,12 @@ require 'spec_helper'
 
 module Cucumber::Deprecate
   describe CliOption do
-    let(:subject) { Cucumber::Deprecate::CliOption }
     let(:error_stream) { double }
 
     context '.deprecate' do
       it 'outputs a warning to the provided channel' do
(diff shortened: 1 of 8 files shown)


spec/cucumber/deprecate_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+module Cucumber::Deprecate
+  describe CliOption do
+    let(:subject) { Cucumber::Deprecate::CliOption }
```
olleolleolle: If you want to, you can use `described_class` instead of defining a `subject` with the same value.

vincent-psarga: nice, thanks for the tip :)

I'm often confused with `subject {...}`, `let(:subject) { ...}` or not defining any subject at all to be honest :D

olleolleolle: Glad it helped!

PR #1396, opened 2020-03-09
Fix CCK errors

vincent-psarga:
Fix the build errors we currently have when running Cucumber against the Compatibility Kit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,13 +5,6 @@
 version: 2.1
 
 commands:
-  add_nokogiri_dependencies:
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
Did you consider adding `libxml2-dev` and `libxslt-dev` to the image instead?

vincent-psarga:
> Did you consider adding `libxml2-dev` and `libxslt-dev` to the image instead?

That's indeed what I did in [cucumber#927](https://github.com/cucumber/cucumber/pull/927)

PR #1400, opened 2020-03-18
Remove Aruba

vincent-psarga:
The idea is to remove [aruba](https://github.com/cucumber/aruba) as the automation layer for our test suite.

Here's a non-exhaustive list of the pros and cons for doing so:

**Pros:**

 - it's faster, almost 10 times faster :) I did a benchmark using [hyperfine](https://github.com/sharkdp/hyperfine). The simplest way I admit, just to get a rough idea of the benefits:
   - with Aruba:
      ```shell
      hyperfine 'bundle exec cucumber'
      Benchmark #1: bundle exec cucumber
        Time (mean ± σ):     63.973 s ±  3.243 s    [User: 43.448 s, System: 15.650 s]
        Range (min … max):   58.784 s … 69.282 s    10 runs
      ```
   - without Aruba:
      ```shell
      hyperfine 'bundle exec cucumber'
      Benchmark #1: bundle exec cucumber
        Time (mean ± σ):      6.742 s ±  0.790 s    [User: 4.896 s, System: 1.326 s]
        Range (min … max):    5.931 s …  8.387 s    10 runs
      ```
  Note: for both benchmarks, I disabled the test "Rerun scenarios randomized" in [`randomize.feature:104`](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/cli/randomize.feature#L104) which I can't get passing yet ...

 - it's more stable on Windows: I did a simple experiment, run the tests 100 times in a row on Windows. [It worked !](https://app.circleci.com/pipelines/github/cucumber/cucumber-ruby/411/workflows/83d8702a-6073-4dbf-a51e-2a3618ce532e/jobs/6543/steps) - okay one experiment its definitely not enough to say we are more stable, but since  I started this PR, the failing tests on Windows are the same than on Linux. I did [the same experiment for JRuby](https://app.circleci.com/pipelines/github/cucumber/cucumber-ruby/411/workflows/83d8702a-6073-4dbf-a51e-2a3618ce532e/jobs/6537/steps) and we have the same results.

 - the control of the steps_definitions is in the same repo. I think it may help people contributing too (when I started working on `cucumber-ruby`, I found it a bit hard to understand which steps were available until I discovered where they were located).

 - the comparisons on the output is a bit more clever. Currently in Aruba, it expects the trailing spaces to be included in the `feature` files as they would be in the real output. Personally, I have my editor configured to trim trailing whitespaces (because they are pointless) and I lost count of the number of times I had to revert some feature files I had modified to had back those. The comparison now trims trailing whitespaces on both real and expected output to avoid this kind of trouble.

**Cons:**

 - some tests are less representative of the real way `cucumber-ruby` works. For example, the tests in [`rake_task.feature`](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/rake_task.feature) and [`raketask.feature`](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/raketask.feature) are now using in-process task instead of spawn tasks (which I don't guess is not really representative of the majority of the use cases)

 - we lose the flexibility of `aruba`. For the scenarios in which the `When` step is "I run `bundle exec cucumber ...`" that' not really an issue, we have control there. But for example, the steps where `rake` is invoked, we don't have the possibility (yet) to use parameters etc.

 - I truly hope I did not break the tests meaning and purpose. I did my best to match as much as possible what was done before, but to be honest, I did not ensure that each of the 217 scenarios where failing before replacing Aruba.

 - using global state in the example features is still possible but may be tricky, as this will leak into the runner itself (so the `cucumber` of the command line, not the one invoked in the feature files)

 - I had to do a dirty hack to get [`fail_fast.feature`](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/cli/fail_fast.feature) passing without exiting the real cucumber. So if you run `bundle exec cucumber --fail-fast` and the scenario of this feature fails, the `--fail-fast` option will not be taken into account.

 - there's real risks of leaking data from the invoked cucumber to the real cucumber during execution (as there is no spawning). You can see that pretty easily as sometimes the colours in the output disappear...

 -  it leaves a bit of a mess in the `tmp` directory. Each scenario is executed in its own directory (eg: `tmp/noruba-rerun-scenarios-randomized-aa9122cb-1cf1-4347-aefa-d948b95a0a9b`). Those folders are destroyed unless the scenario fails. This should not be an issue most of the time but may become one when running tests locally and breaking loads of them.


noruba/features/lib/step_definitions/noruba_steps.rb
```diff
@@ -156,6 +66,22 @@ def make_arg_list(args)
...
+    '',
+    '  def initialize(config)',
+    '    @io = config.out_stream',
+    '  end',
+    '',
+    '  def attach(src, media_type)',
+    '    @io.puts src',
+    '  end',
+    'end'
+].join("\n"))
```
olleolleolle, line 82: Whitespace is wonky here. (Thanks for the hard work on this!)

```suggestion
  ].join("\n"))
```

vincent-psarga: indeed thanks for checking :) I haven't even tried yet to get Rubocop passing on this huge and ugly step definition file, but I guess I'll spend a few hours on cleanup if I ever get to ee all scenarios passing.
The good news so far is that I have the same failing tests on Windows and the various Ruby versions. So we may end up with a stable build on Windows  (which would be great)

luke-hill:
Whilst aruba might be a bit archaic e.t.c. - What is the situation we're trying to achieve by removing it. Does it make things work better on windows?

vincent-psarga:
The idea is to get the tests more reliable on Windows indeed (currently, event with a retry in the CI, I still have to re-run the build for windows to get it passing).
And a side benefit would be to het the tests running faster too.

aslakhellesoy:
Aruba is like Selenium. It enables interaction with an application through its user interface (but not below it).

Automating the majority of tests through the UI is slow and brittle. Failing tests provide very limited information about what caused the failure.

This is the main reason we're getting rid of it. Aruba seemed like a good idea at the time I wrote it. In more recent years I've come to realise the concept is flawed.

It's generally better to interact with the system through ports, and observe outcomes through adapters.

noruba/features/lib/step_definitions/command_line.rb
```diff
@@ -85,6 +89,10 @@ def make_arg_list(args)
 class RubyCommand < CommandLine
   def execute(file)
     capture_stdout { require file }
+  rescue RuntimeError => err
+    # no-op: this is raissed when Cucumber fails
```
olleolleolle, line 93:
```suggestion
    # no-op: this is raised when Cucumber fails
```

noruba/features/lib/step_definitions/command_line.rb
```diff
@@ -85,6 +89,10 @@ def make_arg_list(args)
...
+  rescue SystemExit => err
+    # No-op: well, we are ssupposed to exit the rake task
```
olleolleolle, line 95:
```suggestion
    # No-op: well, we are supposed to exit the rake task
```

vincent-psarga: Thanks, for the fixes :)

I'm stuck in lockdown with a keyboard on which the "s" key is half broken (it either does not work or does 2 "s" :D). At least the code breaks when this kind of thing happens, it's harder for the comments :D

(I'll merge your changes tomorrow morning, apparently I can't from the mobile :/ )

olleolleolle: Ouch!

Thanks for your work on this, it's a big job.

vincent-psarga:
I've updated the PR comment so it's a bit clearer what' done here :)

 Following on Aslak comment's below:

> Automating the majority of tests through the UI is slow and brittle. Failing tests provide very limited information about what caused the failure.

Well, it's faster and more stable. We don't get that much more feedback on the reason why the test failed that said. We're still interacting with cucumber at a pretty high level.

PR #1402, opened 2020-03-19
Fix code snippet on docstring

karamosky:
## Summary
When you have a step with a string argument and a Doc String argument, the generated snippet contains a duplicate argument name

## Details
This scenario currently fails

```gherkin
Scenario: Snippet for undefined step with combined arguments
    Given a file named "features/undefined_steps.feature" with:
      """
      Feature:
        Scenario: combined step
          Given a "thing" with:
          \"\"\"
          thing
          \"\"\"
      """
    When I run `cucumber features/undefined_steps.feature -s`
    Then the output should contain:
      """
      Given('a {string} with:') do |string, string2|
      end
      """
```

## Types of changes

## Checklist:

### DIFF
--- a/features/docs/defining_steps/snippets.feature
+++ b/features/docs/defining_steps/snippets.feature
@@ -55,3 +55,21 @@ Feature: Snippets
         pending # Write code here that turns the phrase above into concrete actions
       end
       """
+      
+  Scenario: Snippet for undefined step with combined arguments
+    Given a file named "features/undefined_steps.feature" with:
+      """
+      Feature:
+        Scenario: combined step
+          Given a file named "todos" with:
+          \"\"\"
+          - Buy milk
+          \"\"\"
+      """
+    When I run `cucumber features/undefined_steps.feature -s`
+    Then the output should contain:
+      """
+      Given('a file named {string} with:') do |string, string2|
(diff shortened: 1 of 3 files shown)


luke-hill:
Let me know when you're ready for this to be reviewed.

lib/cucumber/glue/snippet.rb
```diff
@@ -174,7 +174,8 @@ def result
 
         class DocString
           def append_block_parameter_to(array)
-            array << 'string'
+            parameter = format('string%<index>s', index: (array.size + 1 if array.include?('string')))
```
luke-hill: Here I get your logic, but instead of giving this a different name in different circumstances, lets just use a name which will be unique.

A good example would be `doc_string`

karamosky: I agree this is better :) I will make the changes

features/docs/defining_steps/snippets.feature
```diff
@@ -54,4 +54,4 @@ Feature: Snippets
         # table is a Cucumber::MultilineArgument::DataTable
         pending # Write code here that turns the phrase above into concrete actions
       end
-      """
+      """
```
luke-hill: newline char here

PR #1403, opened 2020-03-26
Avoid lazy init of mutex by moving it to constructor

olleolleolle:
## Summary

Move initialization of the `Mutex` to the constructor.

## Details

We need the lock for the primary concern of this class, write & read synchronized.

There's no benefit to waiting to initialize the Mutex.

## Motivation and Context

Avoid concurrency interaction trouble on non-MRI platforms.

## How Has This Been Tested?

Squinting.

## Types of changes

## Checklist:


olleolleolle:
cc @ioquatix

ioquatix:
Nice work!

PR #1409, opened 2020-04-21
Emit Meta message at beginning of run

aslakhellesoy:
See https://github.com/cucumber/cucumber/pull/976


aslakhellesoy:
Also see
* https://github.com/cucumber/cucumber-ruby-wire/pull/29
* https://github.com/cucumber/cucumber-ruby-core/pull/198

PR #1412, opened 2020-04-27
Enable logging non-string - fix #1410

### LINKED ISSUE #1410: Can't log objects
## Summary

Can't `log` objects

## Expected Behavior

This statement:

```ruby
log Date.parse('2020-02-03')
```

prints:

```
2020-02-03
```

## Current Behavior

I get an error:

```
Cucumber::LogTypeInvalid (Cucumber::LogTypeInvalid)
```

## Possible Solution

Allow any object and call `#to_s` on it

## Your Environment

* Version used: `4.0.0.rc.6`
* Operating System and version: `MacOS`

vincent-psarga:
## Summary

See #1410 

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/glue/proto_world.rb
+++ b/lib/cucumber/glue/proto_world.rb
@@ -107,8 +107,8 @@ module Cucumber
         attach(file, mime_type)
       end
 
-      def log(message)
-        attach(message.to_s.dup, 'text/x.cucumber.log+plain')
+      def log(*messages)
+        messages.each { |message| attach(message.to_s.dup, 'text/x.cucumber.log+plain') }
       end
 
       def attach(file, media_type)
--- a/lib/cucumber/glue/proto_world.rb
+++ b/lib/cucumber/glue/proto_world.rb
@@ -108,8 +108,7 @@ module Cucumber
       end
 
       def log(message)
-        raise Cucumber::LogTypeInvalid unless message.is_a?(String)
-        attach(message.dup, 'text/x.cucumber.log+plain')
+        attach(message.to_s.dup, 'text/x.cucumber.log+plain')
       end
 
       def attach(file, media_type)
--- a/spec/cucumber/glue/step_definition_spec.rb
+++ b/spec/cucumber/glue/step_definition_spec.rb
@@ -191,11 +191,12 @@ module Cucumber
           run_step 'Loud'
         end
 
-        it 'raises an exception if the message is not a String' do
+        it 'calls `to_s` if the message is not a String' do
+          expect(user_interface).to receive(:attach).with('["Not", 1, "string"]', 'text/x.cucumber.log+plain')
           dsl.Given(/Loud/) do
             log ['Not', 1, 'string']
           end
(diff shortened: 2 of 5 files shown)


lib/cucumber/glue/proto_world.rb
```diff
@@ -108,8 +108,7 @@ def embed(file, mime_type, _label = 'Screenshot')
       end
 
       def log(message)
-        raise Cucumber::LogTypeInvalid unless message.is_a?(String)
-        attach(message.dup, 'text/x.cucumber.log+plain')
+        attach(message.to_s.dup, 'text/x.cucumber.log+plain')
```
aslakhellesoy, line 111: Why the `dup`?

spec/cucumber/glue/step_definition_spec.rb
```diff
@@ -191,11 +191,12 @@ def step_match(text)
           run_step 'Loud'
         end
 
-        it 'raises an exception if the message is not a String' do
+        it 'calls `to_s` if the message is not a String' do
+          expect(user_interface).to receive(:attach).with('["Not", 1, "string"]', 'text/x.cucumber.log+plain')
           dsl.Given(/Loud/) do
             log ['Not', 1, 'string']
           end
-          expect { run_step 'Loud' }.to raise_exception(Cucumber::LogTypeInvalid)
```
aslakhellesoy, line 198: Is `Cucumber::LogTypeInvalid` still used? Should it be removed?

lib/cucumber/glue/proto_world.rb
vincent-psarga: Otherwise we log the object itself and it will display its value at display time, not at log time (there a test showing that and it fails without the `dup`)

spec/cucumber/glue/step_definition_spec.rb
vincent-psarga: indeed

lib/cucumber/glue/proto_world.rb
aslakhellesoy: I understand why we used to do `message.dup`, but I don't think we need `message.to_s.dup`. I think `message.to_s` is sufficient.

olleolleolle: (Me, a reader, _thought_ this was because the string would be used in a mutable context.)

vincent-psarga: I though the `to_s` would be enough too. In fact it's enough for non-string objects (obviously :D ) but when using a String we need the `dup`.

An example from the tests:

```ruby
        describe 'when modifying the printed variable after the call to puts' do
          define_feature <<-FEATURE
        Feature: Banana party

          Scenario: Monkey eats banana
            When puts is called twice for the same variable
          FEATURE

          define_steps do
            When(/^puts is called twice for the same variable$/) do
              foo = String.new('a')
              puts foo
              foo.upcase!
              puts foo
            end
          end

```

This test fails without the `dup`. I suspect that String.to_s do not return a duplicate but imply itself.

And @olleolleolle you're not wrong, it's just that the string might be mutated between the moment it I logged and the moment we actually display it.

Issue #1388, opened 2020-02-14
json-reporter: do not fail when embedding byte strings


vincent-psarga:
## Summary

While playing with the CCK, I found an issue for which I needed a money patch: https://github.com/cucumber/cucumber/pull/898/commits/d35ecc8b7c02e2690b8d50abbb0b8d6a2e04a9d9

Basically, if `File.file?(arg)` raises an exception, we throw it and do not attach the file. This should be included in a fix release of `cucumber-3`

## Current Behavior

~~~
~~~

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Operating System and version:
* Link to your project:

PR #1414, opened 2020-04-28
Prepare release 3.2.0

vincent-psarga:
Note: this should not be merged back to master, simply do a release with:
 - [x] Fix issue #1388 (which should help with Ruby CCK)
 - [x] Backport the `--out url` option so people who still use `cucumber-ruby 3.x` should be able to push report easily (#1406 and #1395)


vincent-psarga:
So there's no Webrick available for Ruby 2.2 apparently (https://github.com/ruby/webrick/commit/6259c3790ceb5773604cc9a2eb107e8fbaaad7e6 - it used to support 2.5+ and went down to 2.3 if I understand this correctly).

I'm puzzled on how to handle that:
 - do a release knowing that we have a dev_dependency failing for Ruby 2.2 (which EOL was 2 years ago) ?
 - remove the tests related to the `--out=url` option that rely on Webrick (knowing that this will be the last 3.x release on `cucumber-ruby` and work will now be done on master where we still have the tests working)

PR #1440, opened 2020-07-03
Support diff-lcs 1.4

mvz:

Cucumber-ruby does not work with the latest version of diff-lcs.

`Array#flatten` works recursively and (apparently) operates on anything that supports `#to_ary`. Diff-lcs 1.4 added `Diff::LCS::Change#to_ary`. Because of this `DiffMatrices#changes` raises an error. This pull request avoid flattening of Diff::LCS::Change objects by limiting flattening to one level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Improved
 
-* N/A
+* Update code to be compatible with `diff-lcs` versions 1.3 and 1.4
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


luke-hill:
All good in theory. Only thing I remember is the upper limit was added quite recently (I think by @vincent-psarga) So long as this doesn't conflict with that I'm happy with it.

mvz:
@luke-hill yes, I invited @vincent-psarga to review this pull request precisely because he was the one who introduced that limit.

vincent-psarga:
To be honest, I added the limit because the build was broken with 1.4 release and I needed a quick fix :) I'll try to have a look next week but if anyone feels it can (and should) be merged, don't hesitate :)

vincent-psarga:
That looks ok to me, thanks for finding out where the issue came from. Would you mind adding a small entry to the changelog to have a trace of that ?
And after the entry is added it'll be OK to be merged :)

mvz:
@vincent-psarga done. My brain always goes into an infinite loop when I try to pick a changelog category, so that choice may be a bit random (it's a Change, that makes an Improvement, by Fixing a problem 🤷).

vincent-psarga:
> @vincent-psarga done. My brain always goes into an infinite loop when I try to pick a changelog category, so that choice may be a bit random (it's a Change, that makes an Improvement, by Fixing a problem 🤷).

Oh I understand how "funny" it can be sometimes to get it in the right category. I think "Improved" is definitely right here :)

Issue #1445, opened 2020-07-15
NoMethodError (private method `puts' called for #<Cucumber::Formatter::Interceptor::Pipe:0x00007fc8a5bc0c08>)


haonv88:
**Describe the bug**
When using the JUnit Formatter for --format , I got error `NoMethodError (private method `puts' called for #<Cucumber::Formatter::Interceptor::Pipe:0x00007fc8a5bc0c08>)`

**To Reproduce**
1. Go to rails console
2. require 'cucumber/formatter/interceptor'
3. interceptederr = Cucumber::Formatter::Interceptor::Pipe.wrap(:stderr)
4. $stderr.puts "test" # raise error `NoMethodError (private method `puts' called for #<Cucumber::Formatter::Interceptor::Pipe:0x00007fc8a5bc0c08>)`

**Expected behavior**
Should not raise error. 

**Context & Motivation**

As I research, in cucumber-4.1.0/lib/cucumber/formatter/interceptor.rb:34
$stderr.puts("test") will invoke method
```
        def method_missing(method, *args, &blk)
          @pipe.send(method, *args, &blk) || super
        end
```
@pipe.send(method, *args, &blk) will log the error and return nil, then it keeps calling `super` and raise error

**Screenshots**
<img width="1295" alt="Screen Shot 2020-07-15 at 15 01 21" src="https://user-images.githubusercontent.com/7000603/87519638-18256580-c6ac-11ea-97f6-eed4c243cb69.png">

**Your Environment**
 - Versions used [4.1.0]
 - Operating System and version [MacOs 10.15.5 ]

vincent-psarga:
Thanks for the report.

I'm trying to investigate on what may cause this issue. Do you have a full stack-trace by any chance or is it the same than when you use the method to reproduce the issue ?

haonv88:
Hi @vincent-psarga I don't have full stack-trace. 
```
          command: |
            [[ -n $SKIP_EXECUTION ]] || (mkdir -p ~/cucumber && \
            ./bin/cucumber --format junit \
                                 --out ~/cucumber/junit.xml \
                                 $(circleci tests glob "features/**/*.feature" | circleci tests split --split-by=filesize))
```
When I run cucumber in CI. The above error (`NoMethodError private method`) will hide the real errors from ruby libraries, because `$stderr.puts` does not work anymore.

luke-hill:
[...]

PR #1449, opened 2020-07-21
Fix errors when running JUnit reporter

vincent-psarga:
See #1445 

The trouble came from `method_missing` in the interceptor which called `super` when the method called on the stream did not return a value.

This is a piece of code showing the same kind of issue:

```ruby
class FakePipe
  def no_return_value
    puts "No, I won't return anything"
  end

  def with_return_value
    puts "I let something here before returning a value"
    return 1
  end
end

class FakeInterceptor
  def initialize
    @pipe = FakePipe.new()
  end

  def method_missing(method, *args, &blk)
    @pipe.send(method, *args, &blk) || super
  end
end

interceptor = FakeInterceptor.new()
interceptor.with_return_value # calls  `with_return_value` on the fake stream and does not call super 
interceptor.no_return_value   # calls `no_return_value` on the fake stream, then `super` which fails.
```

I've check in the code and did not find other `method_missing` implementation which may show the same problem.

### DIFF
--- a/lib/cucumber/formatter/interceptor.rb
+++ b/lib/cucumber/formatter/interceptor.rb
@@ -5,6 +5,7 @@ module Cucumber
     module Interceptor
       class Pipe
         attr_reader :pipe
+
         def initialize(pipe)
           @pipe = pipe
           @buffer = StringIO.new
@@ -31,7 +32,8 @@ module Cucumber
         end
 
         def method_missing(method, *args, &blk)
-          @pipe.send(method, *args, &blk) || super
+          return @pipe.send(method, *args, &blk) if @pipe.respond_to?(method)
+          super
         end
 
         def respond_to_missing?(method, include_private = false)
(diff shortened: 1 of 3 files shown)


luke-hill:
Yeh so just for an explanation (in case you were interested). The return value of `puts` is `nil` but the return value of `p` is self (i.e. whatever you pass into `p`).

I think this is the reason often it's good in normal debugging to use `p` as it spits more out, but it modifies returns.

And then in your "or evaluation" the first bit is ran and returns a value, but the return value is `nil`, and then the or runs the second bit.

lib/cucumber/formatter/interceptor.rb
```diff
@@ -31,7 +32,7 @@ def unwrap!
         end
 
         def method_missing(method, *args, &blk)
-          @pipe.send(method, *args, &blk) || super
+          @pipe.respond_to?(method) ? @pipe.send(method, *args, &blk) : super
```
luke-hill, line 35: This is pretty solid. It's also the perfect companion. Often we should be using these 2 in tandem.

vincent-psarga: I think so :) and it's definitely safer this way

PR #1456, opened 2020-08-18
wip: removing the at_exit hook and explicitly closing ios inside formatters instead

cbliard:
when io is close at_exit when doing --publish, then the HttpIo is
pushing the report. This should not be pushed if an error occurs, like
the feature file does not exist. Hence this removal.

Co-authored-by: Aslak Hellesøy <aslak.hellesoy@smartbear.com>
Co-authored-by: Seb Rose <seb.rose@smartbear.com>

### DIFF
--- a/spec/cucumber/formatter/html_spec.rb
+++ b/spec/cucumber/formatter/html_spec.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'cucumber/formatter/spec_helper'
+require 'cucumber/formatter/html'
+require 'cucumber/cli/options'
+
+module Cucumber
+  module Formatter
+    describe HTML do
+      extend SpecHelperDsl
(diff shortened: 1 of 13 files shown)


spec/cucumber/formatter/html_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+            Scenario: Monkey eats banana
+              Given there are bananas
+        FEATURE
+
+        before(:each) do
+          run_defined_feature
+        end
+
+        it 'outputs html' do
+          expect(@out.string).to include '<!DOCTYPE html>'
```
luke-hill, line 32: If you want to be more rigorous, these could use the `start_with` and `end_with` matchers.

That is unless there is some other metadata (In which case ignore this comment).

aslakhellesoy:
Superseded by #1458

PR #1457, opened 2020-08-18
Add `file` attribute for CircleCI test splitting

bmack123:
Add `file` attribute for CircleCI test splitting

CircleCI uses the `junit` formatter to output results from cucumber. It can split up the files in parallel containers using the timing data as well as `filename` information found in the xml to balance out the test runs so that they roughly all finish at the same time. The current implementation of the junit formatter does not provide the filename as part of the xml so we are not able to take advantage of this feature 

This was originally implemented by Mike Lay at CircleCi, so he gets the credit for the implementation.  It builds the filename into the XML for each test case from the `@current_feature_data[:uri]` inside the junit formatter.

**Other Notes**

I have not submitted a PR for this repo before, if I formatted anything incorrectly thats my mistake and I'll work to fix them as fast as I can. I appear to be failing some build steps but it looks like master is failing those as well so I wasn't too worried about them being specific to the code.

### DIFF
--- a/features/docs/formatters/junit_formatter.feature
+++ b/features/docs/formatters/junit_formatter.feature
@@ -83,7 +83,7 @@ Feature: JUnit output formatter
       """
       <?xml version="1.0" encoding="UTF-8"?>
       <testsuite failures="1" errors="0" skipped="0" tests="2" time="0.05" name="One passing scenario, one failing scenario">
-      <testcase classname="One passing scenario, one failing scenario" name="Passing" time="0.05">
+      <testcase classname="One passing scenario, one failing scenario" name="Passing" time="0.05" file="features/one_passing_one_failing.feature">
         <system-out>
           <![CDATA[]]>
         </system-out>
@@ -91,7 +91,7 @@ Feature: JUnit output formatter
           <![CDATA[]]>
         </system-err>
       </testcase>
-      <testcase classname="One passing scenario, one failing scenario" name="Failing" time="0.05">
+      <testcase classname="One passing scenario, one failing scenario" name="Failing" time="0.05" file="features/one_passing_one_failing.feature">
         <failure message="failed Failing" type="failed">
           <![CDATA[Scenario: Failing
 
@@ -124,7 +124,7 @@ Feature: JUnit output formatter
       """
       <?xml version="1.0" encoding="UTF-8"?>
       <testsuite failures="1" errors="0" skipped="0" tests="2" time="0.05" name="Subdirectory - One passing scenario, one failing scenario">
-      <testcase classname="Subdirectory - One passing scenario, one failing scenario" name="Passing" time="0.05">
+      <testcase classname="Subdirectory - One passing scenario, one failing scenario" name="Passing" time="0.05" file="features/some_subdirectory/one_passing_one_failing.feature">
         <system-out>
           <![CDATA[]]>
         </system-out>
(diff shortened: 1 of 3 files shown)


luke-hill:
So in principle what you've done is fine. But it's worth pointing out this entire area of the codebase is in a large state of flux (We're moving over to using protobuf to pass messages which then can get consumed by custom formatters).

As for the failing builds. 2.3 shouldn't be on there, and the cck is being tweaked as we speak. 2.4 possibly needs a bit of TLC, I can check this later (But this PR is currently in draft, so I imagine you want to rework bits)?

The JUnit formatter (especially in the default formatter parts of the codebase), is quite old and I'm not sure exactly what we're planning on doing with this one (I know we're working on the JSON / HTML ones a lot currently), but yeh just so you know this area might be vanishing / disappearing to another sub-project.

@vincent-psarga I know you've done a load of work on the html formatter, are we planning on doing anything with the JUnit formatter?

bmack123:
Ultimately, I was just hoping to be able to support Circle's particular requirements. If they happen as part of a rework then thats fine by me. If there is a `junit` rework in progress  is there any way to include `file` as part of the `testcase` metadata?

mkly:
@luke-hill If this was changed to hide the `file` attribute behind either a command-line argument or an environment variable, do you think that might be a solution that could get merged? The benefit of that would be that it would not add any maintenance time for anyone unless they started making use of it explicitly.

I do understand the desire to not add complexity when you are in more of a feature-freeze maintenance mode position. We do see reasonable demand for this with CircleCI, so we would be motivated to make sure this kept working in the future. Any other suggestions that might make you feel more comfortable about a merge, definitely let us know!

aslakhellesoy:
I’m all for adding this even if the longer term goal is to replace it with a new implementation in the monorepo.

One question - do you know of an XML schema and/or DTD for JUnit XML that declares this new field? If not, this might break validation on platforms that use a schema/DTD that lacks it.

luke-hill:
@bmack123 I **don't** believe any work is being done to legacy / new JUnit formatters.

@mkly As mentioned, any work e.t.c. being done to improve formatters is generally good. Given the recent switch to keep all legacy formatters in place until cucumber6 is released, any work to the legacy / new JUnit formatters will be retained for a long time, so feel free to tackle it however you see fit.

Typically we would try use the top level Cucumber configuration, but using an Env var is fine too. Each to their own (And hiding it as default off also means we can introduce it reasonably easy). As aslak has mentioned as well, any introduction / improvement here is good because (Hope I'm not being too harsh), the JUnit formatter is the least formed of all the existing legacy formatters. It outputs very little info.

mkly:
JUnit XML is a bit of a "loose" standard as far as I'm aware, and there are other jUnit type formatters that do add `file` already. I agree with @aslakhellesoy about the potential issue with XML schema and/or DTD and we would probably break things on some people running validations if we did not hide it behind a flag(either env var or cli option).

mkly:
I went ahead and built this out behind a cli arg in #1471 

Feel like that should allow for people to add it they need it, but remain the same as it is today for others that are not making use of it. This will avoid any surprises with stuff like validation etc. Definitely let me know if there is anything I should change or adjust and thanks @bmack123 for sending this PR over in the first place! :tropical_drink:

bmack123:
closing this in favor of #1471

PR #1459, opened 2020-08-19
Update CI, remove support for unsupported rubies

aslakhellesoy:
This removes support for Ruby 2.3 and 2.4 as they are now unsupported ruby versions.

This does not mean Cucumber won't work on those rubies. It just means we no longer verify that it does. The reason for this is that some gems we depend on require ruby >= 2.5.


olleolleolle (approved):
Thanks!

PR #1471, opened 2020-09-04
Add optional file attribute to junit formatter

mkly:
Add optional file attribute to junit formatter

* Allow for adding `file=` attribute to junit
* Use cli arg `-f junit,fileattribute=true`

CircleCI has a test splitting feature that will split tests across multiple jobs based on the time spent. This feature requires a `file` attribute to be added. The cucumber-ruby test junit formatter does not include this attribute.

This adds a `-f junit,fileattribute=true` cli option that will add the `file` attribute to the `<testcase/>` elements in the junit formatter. By default, all junit formatter output will be unchanged unless this additional option is added.

**Additional context**
This is a continuation of Pull Request #1457
The `junit,fileattribute=true` was based on the conventions already in use, but I am more than happy to change that to something else if you feel there is a more appropriate way. This is my first submission to this project, so my apologies in advance if I missed anything and more than happy to make any adjustments or changes that you might feel will be easier to maintain etc. Thank you for your work on the project. I very much appreciate it!
Closes #1457

### DIFF
--- a/features/docs/formatters/junit_formatter.feature
+++ b/features/docs/formatters/junit_formatter.feature
@@ -114,6 +114,47 @@ Feature: JUnit output formatter
 
       """
 
+  Scenario: one feature, one passing scenario, one failing scenario, add file attribute
+    When I run `cucumber --format junit,fileattribute=true --out tmp/ features/one_passing_one_failing.feature`
+    Then it should fail with:
+      """
+
+      """
+    And the junit output file "tmp/TEST-features-one_passing_one_failing.xml" should contain:
+      """
+      <?xml version="1.0" encoding="UTF-8"?>
+      <testsuite failures="1" errors="0" skipped="0" tests="2" time="0.05" name="One passing scenario, one failing scenario">
+      <testcase classname="One passing scenario, one failing scenario" name="Passing" time="0.05" file="features/one_passing_one_failing.feature">
+        <system-out>
+          <![CDATA[]]>
+        </system-out>
+        <system-err>
+          <![CDATA[]]>
+        </system-err>
+      </testcase>
+      <testcase classname="One passing scenario, one failing scenario" name="Failing" time="0.05" file="features/one_passing_one_failing.feature">
+        <failure message="failed Failing" type="failed">
+          <![CDATA[Scenario: Failing
+
+      Given this step fails
+
+      Message:
+      ]]>
+          <![CDATA[ (RuntimeError)
+      ./features/step_definitions/steps.rb:4:in `/^this step fails$/'
+      features/one_passing_one_failing.feature:7:in `this step fails']]>
+        </failure>
+        <system-out>
+          <![CDATA[]]>
+        </system-out>
+        <system-err>
+          <![CDATA[]]>
+        </system-err>
+      </testcase>
+      </testsuite>
+
+      """
+
   Scenario: one feature in a subdirectory, one passing scenario, one failing scenario
     When I run `cucumber --format junit --out tmp/ features/some_subdirectory/one_passing_one_failing.feature --require features`
     Then it should fail with:
@@ -175,6 +216,38 @@ Feature: JUnit output formatter
         </system-err>
       </testcase>
       <testcase classname="Pending step" name="Undefined" time="0.05">
+        <skipped/>
+        <system-out>
+          <![CDATA[]]>
+        </system-out>
+        <system-err>
+          <![CDATA[]]>
+        </system-err>
+      </testcase>
+      </testsuite>
+
+      """
+
+  Scenario: pending and undefined steps add fileattribute
+    When I run `cucumber --format junit,fileattribute=true --out tmp/ features/pending.feature`
+    Then it should pass with:
+      """
+
+      """
+    And the junit output file "tmp/TEST-features-pending.xml" should contain:
+      """
+      <?xml version="1.0" encoding="UTF-8"?>
+      <testsuite failures="0" errors="0" skipped="2" tests="2" time="0.05" name="Pending step">
+      <testcase classname="Pending step" name="Pending" time="0.05" file="features/pending.feature">
+        <skipped/>
+        <system-out>
+          <![CDATA[]]>
+        </system-out>
+        <system-err>
+          <![CDATA[]]>
+        </system-err>
+      </testcase>
+      <testcase classname="Pending step" name="Undefined" time="0.05" file="features/pending.feature">
         <skipped/>
         <system-out>
           <![CDATA[]]>
--- a/lib/cucumber/formatter/junit.rb
+++ b/lib/cucumber/formatter/junit.rb
@@ -132,9 +132,11 @@ module Cucumber
         duration = ResultBuilder.new(result).test_case_duration
         @current_feature_data[:time] += duration
         classname = @current_feature_data[:feature].name
+        filename = @current_feature_data[:uri]
         name = scenario_designation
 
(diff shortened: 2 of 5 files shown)


vincent-psarga:
That looks nice thanks :)
I'm a bit surprised by the option look (with the comma inside), I've never really seen that kind of option before, but I guess it's still better than having a `--format=junit_with_filename` option or something like that.

Another thing I'm wondering (although I am certainly wrong here :D ), all test cases are pointing to the same file, shouldn't it contain the line number too ? (something like `file="features/one_passing_one_failing.feature:5"` for example). I don't know if Circle handle that kind of things, it really css a candid question here :)

lib/cucumber/formatter/junit.rb
```diff
@@ -153,8 +155,21 @@ def build_testcase(result, scenario_designation, output)
...
+      end
+
+      def get_testcase_arguments(classname, name, duration, filename)
+        arguments = { classname: classname, name: name, time: format('%<duration>.6f', duration: duration) }
+        arguments[:file] = filename if should_add_fileattribute?
+
+        arguments
+      end
+
+      def should_add_fileattribute?
```
luke-hill, line 169: lets remove the `should_` prefix here?

luke-hill (approved):
There was a few other naming things that seemed off, but they were consistent.

Using this kind of cli flag is also really good because by default nothing changes. So we can introduce in a point release.

lib/cucumber/formatter/junit.rb
mkly: Adjusted. Thank you.

mkly:
@vincent-psarga There actually appeared to be prior art in the code base for that convention with the commas. I too thought it was unusual, but like you, thought it was a good path to avoid the additional option like you described.
https://github.com/cucumber/cucumber-ruby/blob/58f5f5cc15d543ce6113d6b57b923bfe8b03e31c/spec/cucumber/cli/options_spec.rb#L116-L117
https://github.com/cucumber/cucumber-ruby/blob/58f5f5cc15d543ce6113d6b57b923bfe8b03e31c/features/docs/extending_cucumber/custom_formatter.feature#L54

@luke-hill 
> There was a few other naming things that seemed off

That is valid critique. Along with the adjustment to removing `should`, I've also changed `arguments` to `attributes` as they are eventually attributes of the `<testcase/>` element. I also went ahead and made `add_fileattribute?` more defensive by adding a check to make sure `@config.formats` is set even though I believe it would always need to be set for the junit formatter to run. I figured it was better safe than sorry there.

Thanks a bunch for looking at this so quickly. I appreciate it. And thank you for the thoughtful feedback!

mkly:
@vincent-psarga Apologies in that I forgot to address your second question!

> Another thing I'm wondering (although I am certainly wrong herhttps://github.com/mkly/troubleue :D ), all test cases are pointing to the same file, shouldn't it contain the line number too ?

I believe that adding the line number would end up causing the CircleCI test spitting by timing to not work. So for that use case it would be better to leave it off. I could see the general utility of it, though. Thank you again for your help in looking this over.

features/docs/formatters/junit_formatter.feature
```diff
@@ -187,6 +228,38 @@ Feature: JUnit output formatter
...
+  Scenario: pending and undefined steps add fileattribute
+    When I run `cucumber --format junit,fileattribute=true --out tmp/ features/pending.feature`
+    Then it should pass with:
+      """
+
+      """
+    And the junit output file "tmp/TEST-features-pending.xml" should contain:
+      """
+      <?xml version="1.0" encoding="UTF-8"?>
+      <testsuite failures="0" errors="0" skipped="2" tests="2" time="0.05" name="Pending step">
```
luke-hill, line 240: how brittle are these timestamps? I think Vincent had something that regex'd them. Could we use something like that here?

lib/cucumber/formatter/junit.rb
```diff
@@ -157,6 +160,21 @@ def build_testcase(result, scenario_designation, output)
...
+        attributes[:file] = filename if add_fileattribute?
+
+        attributes
+      end
+
+      def add_fileattribute?
+        return false if @config.formats.nil? || @config.formats.empty?
+
+        !!@config.formats.find do |format|
+          format[0] == 'junit' and format.dig(1, 'fileattribute') == 'true'
```
luke-hill, line 174: Ideally should use `&&` here. Can also do `format.first` if it's enumerable (Which it looks like).

lib/cucumber/formatter/junit.rb
```diff
@@ -157,6 +160,21 @@ def build_testcase(result, scenario_designation, output)
         @current_feature_data[:tests] += 1
       end
 
+      def get_testcase_attributes(classname, name, duration, filename)
+        attributes = { classname: classname, name: name, time: format('%<duration>.6f', duration: duration) }
```
luke-hill, line 164: Minor optimisation

```
{ class...... }.tap do |attributes|
  attributes[:file] = filename if add_fileattribute?
end
```

luke-hill (commented):
Couple of cleanups and a quick question

lib/cucumber/formatter/junit.rb
mkly: It is. The formats come in as an array of arrays. Adjusted. Thanks.

lib/cucumber/formatter/junit.rb
mkly: I like that method of using `.tap` much better. Thank you for the suggestion.

features/docs/formatters/junit_formatter.feature
mkly: These timestamps are indeed regex'd to `0.05`

https://github.com/cucumber/cucumber-ruby/blob/master/features/lib/support/junit.rb

And applied to any of those steps:
https://github.com/cucumber/cucumber-ruby/blob/master/features/lib/step_definitions/junit_steps.rb

Because of the above, it is my understanding that this would be stable as well.

luke-hill: Perfect, yes sorry I wasn't aware of that bit of the codebase. Perfectly fine :+1:

luke-hill (approved):
:tada:

mkly:
👋 Just stopping in to see if there is anything that could be pending for me to do here. Definitely, no rush as I realize everyone is busy. Only making sure things are not waiting for action on my part. Thanks again for your help.

luke-hill:
I'll merge this in as it seems fine. We have rollback options if required.

Before I do so, could you add a changelog entry.

chezsmithy:
> I'll merge this in as it seems fine. We have rollback options if required.
> 
> Before I do so, could you add a changelog entry.

@mkly looks like this just needs a changeling entry. I'm excited to see this merge so I can try some of the automated circle ci splitting features. Thanks for the nice change!

chezsmithy:
> I'll merge this in as it seems fine. We have rollback options if required.
> 
> 
> 
> Before I do so, could you add a changelog entry.

@mkly do you have a few minutes to please do the changelog?

bmack123:
@mkly is there anything we can do to get this over the line? I am happy to write a changelog if it will be helpful if I have access to the repo.

luke-hill:
Merged in #1489

chezsmithy:
Any chance we could get this pushed into a release?

luke-hill:
This was merged in PR#1489 and will be available in the next release of cucumber-ruby on rubygems. As for when that is. I've not got around to getting the new release process set up on my machine yet. But I think it'll be reasonably soon.

PR #1479, opened 2020-10-06
Fix CUCUMBER_PUBLISH_TOKEN by fixing curl option parser

aslakhellesoy:
The curl option parser logic was broken, so the `CUCUMBER_PUBLISH_TOKEN=abc` environment variable would result in the following HTTP header: `Authorization: Bearerabc` (without a space).

I've rewritten the parser using shellwords, and also removed the unconventional `-H "A:B" "A:B" "A:B"` syntax (specifying multiple headers with a single `-H`).


vincent-psarga:
I fix the `master` issues and then merge that.

PR #1480, opened 2020-10-07
Fix build

vincent-psarga:
Current build is failing, trying to get that fixed.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,7 @@ require:
   - rubocop-packaging
 
 AllCops:
+  NewCops: disable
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -154,3 +154,78 @@ Style/RegexpLiteral:
   EnforcedStyle: slashes
   Exclude:
     - 'spec/cucumber/project_initializer_spec.rb'
+
```
luke-hill, line 157: Use NewCops and save yourself time here.

See: https://github.com/cucumber/cucumber-rails/blob/master/.rubocop.yml#L10

olleolleolle:
Ah, green @vincent-psarga!

Issue #1473, opened 2020-09-08
--publish should print response body if http status >= 400


aslakhellesoy:
Currently only the HTTP status is displayed...

vincent-psarga:
Every response is now displayed (see #1483) - the report server should return banners when errors occur.

PR #1481, opened 2020-10-16
Include response body when --publish errors

jmks:
A small improvement re: #1473 

Simply includes the response's body in the error message.

Here's an example:

```
$ # before
$ CUCUMBER_PUBLISH_URL="http://localhost:4567/422 -X GET" be cucumber --publish
Traceback (most recent call last):
	2: from /home/jmks/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/cucumber-5.1.3/lib/cucumber/formatter/io.rb:40:in `block (3 levels) in new'
	1: from /home/jmks/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/cucumber-5.1.3/lib/cucumber/formatter/http_io.rb:75:in `close'
/home/jmks/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/cucumber-5.1.3/lib/cucumber/formatter/http_io.rb:128:in `send_content': request to http://localhost:4567/422 failed with status 422 (StandardError)

$ # after
$ CUCUMBER_PUBLISH_URL="http://localhost:4567/422 -X GET" be cucumber --publish
Traceback (most recent call last):
	2: from ~/src/cucumber-ruby/lib/cucumber/formatter/io.rb:40:in `block (3 levels) in new'
	1: from ~/src/cucumber-ruby/lib/cucumber/formatter/http_io.rb:75:in `close'
~/src/cucumber-ruby/lib/cucumber/formatter/http_io.rb:128:in `send_content': request to http://localhost:4567/422 failed with status 422: you did something wrong (StandardError)
```

from the "server":

```ruby
require "sinatra"

get "/422" do
  [422, "you did something wrong"]
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Changed
 
+* `--publish` request errors now include the response's body in the error message
+
 ### Removed
 
 ### Deprecated
--- a/lib/cucumber/formatter/http_io.rb
+++ b/lib/cucumber/formatter/http_io.rb
@@ -125,7 +125,7 @@ module Cucumber
         when Net::HTTPRedirection
           send_content(URI(response['Location']), method, headers, attempt - 1)
         else
-          raise StandardError, "request to #{uri} failed with status #{response.code}"
+          raise StandardError, "request to #{uri} failed with status #{response.code}: #{response.body}"
         end
       end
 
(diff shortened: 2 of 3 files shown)


luke-hill:
Seems legit. Probably worth putting a little changelog entry in as well if you can. Should be easy enough to approve and merge.

jmks:
Good point about the changelog. I've updated it.

luke-hill (approved):

Issue #1469, opened 2020-09-01
Running Scenario by line number from within Scenario doesn't work


BARK-PHOLLAND:
## Summary

I used to be able to run `cucumber feature_file_name.feature:XX` with `XX` being a line anywhere in the scenario (usually the line that failed last time, to easily re-run a failure) but now it only works if `XX` is the number of the "Scenario" heading line.

## Expected Behavior

`cucumber feature_file_name.feature:XX` should run the Scenario, from the beginning, that contains line `XX`.

## Current Behavior

The Scenario does not run, and the following output is produced.
~~~
0 scenarios
0 steps
0m0.000s
~~~

## Steps to Reproduce (for bugs)

1. Run any Scenario (with `cucumber feature_file_name.feature:XX`) from the command line, tagging the line number of any line within the scenario other than the "Scenario" heading

## Your Environment
ruby 2.7.1
MacOS 10.15.6
cucumber-rails 2.1
cucumber 4.1.0

vincent-psarga:
Some explanations from @brasmusson: https://cucumberbdd.slack.com/archives/C62GZFLLT/p1595951692019300

The long story. Cucumber-Ruby was the first cucumber implementations. Originally the execution of feature file was done by iterating the AST. This made the cucumber implementation fragile and error prone. But the whole AST was available when executing, which means that the hooks could access the AST. In Cucumber-Ruby 2.0 the concept of compiling feature files was introduced. However the compiled test cases includes links back to the AST, so the whole AST  was still available when executing, that is was still accessible from hooks. Then the concept of Pickles was created and the responsibility of compiling feature files became part of the responsibilities of the gherkin library (https://github.com/cucumber/cucumber/tree/master/gherkin#pickles).
The introduction of Pickles means IMHO that the central part of any cucumber implementation is a loop `execute(Pickle)`, and it also specifies that the data available when executing is only the Pickle. Since only the Pickle is available when executing the hooks consequently can access only data of the Pickle - not the data of the AST from which the Pickle was compiled.
The gherkin compiler and Pickle was introduced into Cucumber-Ruby i v4, and consequently hooks can no longer access the AST. Since Cucumber-Ruby users have been used to be able to access the AST from hooks, many of them now miss being able to do that.
[...]

PR #1497, opened 2021-01-13
Restore ability to specify a line number within a scenario.

botandrose:
Fixes regression described in https://github.com/cucumber/cucumber-ruby/issues/1469

**Regression**
I used to be able to run `cucumber feature_file_name.feature:XX` with XX being a line anywhere in the scenario (usually the line that failed last time, to easily re-run a failure) but now it only works if XX is the number of the "Scenario" heading line. Additionally, running `:1` used to run all scenarios, now it runs none.

**Solution**
Due to the internal changes that seem to make a "proper" fix invasive (see https://github.com/cucumber/cucumber-ruby/issues/1469#issuecomment-687193405), here is hacky solution in the CLI layer. When an file argument is found with a line number specified, we open the file, inspect the specified line, iterate upwards until we see `Scenario:`, and then mutate the supplied line number into that line number. If we happen to find `Scenario Outline:` first, just leave the number as-is. If neither is found, i.e. we're above the first scenario in e.g. `Background:` or `Feature:`, then all scenarios should be run, so remove the line number completely.

**Notes**
This is hardly the most elegant solution (or code), but I want to solicit discussion and feedback from the cucumber dev community before iterating much further on it. What do y'all think? Is there a better way to go about fixing this regression that doesn't involve rearchitecting cucumber's internals all over again?

### DIFF
--- a/lib/cucumber/cli/options.rb
+++ b/lib/cucumber/cli/options.rb
@@ -154,6 +154,7 @@ Specify SEED to reproduce the shuffling from a previous run.
         process_publish_options
 
         @args.map! { |a| "#{a}:#{@options[:lines]}" } if @options[:lines]
+        adjust_line_numbers
 
         extract_environment_variables
         @options[:paths] = @args.dup # whatver is left over
@@ -463,6 +464,31 @@ Specify SEED to reproduce the shuffling from a previous run.
           end
         end
       end
+
+      def adjust_line_numbers
+        @args.map! do |arg|
+          next arg unless arg =~ /:\d+$/
+          line_numbers = arg.split(':')
+          file = line_numbers.shift
+          lines = File.readlines(file)
+          line_numbers.map! do |line_number|
+            line_number.to_i.downto(1).find do |i|
+              line = lines[i - 1]
+              raise ScenarioOutlineFound if line =~ /^\s+Scenario Outline:/
+              line =~ /^\s+Scenario:/
+            end
+          rescue ScenarioOutlineFound
+            line_number
+          end
+          line_numbers.unshift(file)
+          line_numbers.compact.join(':')
+        rescue Errno::ENOENT
+          arg
+        end
+      end
+
+      class ScenarioOutlineFound < RuntimeError; end
+      private_constant :ScenarioOutlineFound
 
       def disable_profile_loading?
         @disable_profile_loading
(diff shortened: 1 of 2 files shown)


luke-hill (commented):
Before reviewing the code part of the test. Could we have a scenario with 3 steps, and you use the 2nd step (Just so we cover our bases with future regressions).

You can add this as an additional one or modify the existing one. It just will be a bit more robust.

botandrose:
Okay, test modified as requested! Rebased on master to resolve windows failure, too.

luke-hill:
Cheers. I'll have a look at the code when I next have some free time :) Unless someone else gets to it first. The tests though enable us to check this new logic still works.

lib/cucumber/cli/options.rb
```diff
@@ -464,6 +465,31 @@ def extract_environment_variables
...
+      def adjust_line_numbers
+        @args.map! do |arg|
+          next arg unless arg =~ /:\d+$/
+          line_numbers = arg.split(':')
+          file = line_numbers.shift
+          lines = File.readlines(file)
+          line_numbers.map! do |line_number|
+            line_number.to_i.downto(1).find do |i|
+              line = lines[i - 1]
+              raise ScenarioOutlineFound if line =~ /^\s+Scenario Outline:/
```
brasmusson, line 477: In Gherkin today also Scenarios can have an Examples Table ("Scenario Outline" only a synonym for "Scenario"), so it is not possible to decide if it is a simple Scenario (without Examples Table) of a Scenario with examples (previously known as "Scenario Outline" only by examine the keyword for the scenario line.

lib/cucumber/cli/options.rb
```diff
@@ -464,6 +465,31 @@ def extract_environment_variables
...
+              line =~ /^\s+Scenario:/
```
brasmusson, line 478: A) This does not handle feature files in other language than English.
B) This does not handle all alternatives possible in a feature file in English (in Gherkin today "Example" is an alternative to "Scenario")

brasmusson (changes requested):
It is unclear what the intended behavior is several corner cases
- when the line number is the line of the tag-line preceding the scenario line (the line of the scenario keyword)
- when the line number is the line of an (outline) step in a scenario with examples (previously known as a "Scenario Outline"
- etc
In the end I think this kind of change needs an rspec spec to specify the behavior comprehensively.

mattwynne:
Thanks for your contribution @botandrose.

I've got to say, I'd much rather we fix the shortcomings in the gherkin internals that have made this awkward, rather than plastering on a hack around the outside of it. Do you have any energy to help me pursue the other avenue to get this problem solved properly?

Issue #1468, opened 2020-09-01
Specifying --publish and no formatter does not default to using pretty formatter


aslakhellesoy:
This is because `--publish` registers as a formatter internally, causing cucumber to think there is no need to default to pretty

aslakhellesoy:
bump

cbliard:
Experienced by this customer https://app.intercom.com/a/apps/fy7zk1ud/inbox/inbox/all/conversations/160128100001667

>How to temporarily disable this feature, so that I can see results and failures in command line when building the cucumber tests ?

meliesan:
If I may add to previous issue experienced: (using cucumber/capybara ruby)
The undefined step definitions "code snippet" is also not provided in the report and unable to see it in the console either.

I have had to downgrade to Cucumber 5.1.3 but still seeing following issues, which requires me to still use the report; 

1. If I add "default: --publish-quiet" in cucumber.yml, I am able to see the code snippet for undefined steps in the console but not the details of the failures.
2. If I add "default: --publish" to cucumber.yml, I see the details of failures in the report only but no undefined steps code snippet in neither console nor the report.
3. If I omit any reference to the report, i.e. "--publish" or "--publish-quiet" in the cucumber.yml file, I see the code snippet in the console only but not the details of failures.

Currently I have to comment/uncomment the line in cucumber.yml file depending on if I want to see the code snippet in the console or the failures in the report.
Hope this helps !
Mel

meliesan:
I found the workaround to have both the url to the cucumber report and the details at runtime in the console.
in cucumber.yml, I added : 
`default: --publish --format pretty`
it's a little bit slow to kickoff the tests but it works.

now I can probably also use: (but have not tried it) :
`default: --publish-quiet --format pretty`

aslakhellesoy:
Fixed by #1500

PR #1500, opened 2021-01-20
Add pretty formatter with publish

aurelien-reeves:

https://github.com/cucumber/cucumber-ruby/issues/1468
Missing pretty formatter when using --publish

In cli/configuration, I have added the pretty formatter per default when only message(s) formatter(s) are specified.

### DIFF
--- a/lib/cucumber/cli/configuration.rb
+++ b/lib/cucumber/cli/configuration.rb
@@ -140,8 +140,14 @@ module Cucumber
       end
 
       def needs_default_formatter?
-        return true if @options[:formats].empty?
+        formatter_missing? || publish_only?
+      end
 
+      def formatter_missing?
+        @options[:formats].empty?
+      end
+
+      def publish_only?
         @options[:formats]
           .uniq
           .map { |formatter, _, stream| [formatter, stream] }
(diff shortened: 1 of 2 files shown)


lib/cucumber/cli/configuration.rb
```diff
@@ -126,13 +126,29 @@ def set_environment_variables
...
         end
         @options[:formats].uniq!
         @options.check_formatter_stream_conflicts
       end
+
+      def add_default_formatter
+        @options[:formats] << ['pretty', {}, @out_stream]
+      end
+
+      def needs_default_formatter?
```
luke-hill, line 142: This method is doing two things. It's checking the presence directly and checking whether the property of format is there? Could we maybe separate those two concerns and then use a boolean `&&`

i.e.

`format_missing? && format_incorrect_type?`

aurelien-reeves: I've made the requested change.
Does it look better?

aurelien-reeves:
Why is the windows build failing?
Is it a known issue?

I assume that it is not because of the changes I submit but maybe I am wrong?

luke-hill (approved):

PR #1503, opened 2021-01-26
Fix windows build

aurelien-reeves:
Install MSYS2 using Chocolatey allow a silent install.
Then, ridk can set-up msys2 properly and the ruby devkit.

The circle-ci windows build is broken

`ridk` may not be silent when installing `msys2`.
Installing `msys2` using `choco` allows to install it silently, then `ridk install 2 3` can properly set-up the ruby devkit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,14 @@ commands:
           name: "Add ansicon"
           command: |
             choco install ansicon
+      - run:
+          name: "Add MSYS2"
+          command: |
+            choco install msys2
+      - run:
+          name: "Set-up Ruby devkit"
+          command: |
+            ridk install 2 3
 
   build:
     description: "Build and run the tests"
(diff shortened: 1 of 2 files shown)


luke-hill:
Given this fixes the windows build. And I know from first hand experience that ridk is often the tricky bit (And we've had a lot of issues in the past here with windows), thankyou very much for looking into this and providing a fix!!!!

I'll approve. Could you write a little changelog message please. Because I know in the past we sometimes have missed "when" the windows build broke specifically. This would help us nail down that in Cucumber 5.x the windows build was good. It'll also probably encourage us to keep it this way :)

Thanks again

luke-hill (approved):

aurelien-reeves:
Sorry, I've been messy :'(

luke-hill:
Let me know when you want me to look at something.

aurelien-reeves:
Everything is fine now for that PR, it is alright now

PR #1504, opened 2021-01-27
Ignore CUCUMBER_PUBLISH_TOKEN during cucumber tests

aurelien-reeves:
Re-set it properly after each tests to be able to actually publish our
own reports.

This is an attempt to fix our own GHA

Unset CUCUMBER_PUBLISH_TOKEN before each cucumber tests to avoid interference within the test.
Then reset it properly after each tests to be able to publish our own reports.

### DIFF
--- a/features/lib/support/env.rb
+++ b/features/lib/support/env.rb
@@ -22,15 +22,12 @@ After do |scenario|
 end
 
 Around do |_, block|
-  original_publish_token = ENV['CUCUMBER_PUBLISH_TOKEN']
-  ENV.delete('CUCUMBER_PUBLISH_TOKEN')
-
+  original_publish_token = ENV.delete('CUCUMBER_PUBLISH_TOKEN')
   original_coloring = Cucumber::Term::ANSIColor.coloring?
 
   block.call
 
   Cucumber::Term::ANSIColor.coloring = original_coloring
-
   ENV['CUCUMBER_PUBLISH_TOKEN'] = original_publish_token
 end
 
--- a/features/lib/support/env.rb
+++ b/features/lib/support/env.rb
@@ -22,9 +22,16 @@ After do |scenario|
 end
 
 Around do |_, block|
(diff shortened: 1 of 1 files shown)


features/lib/support/env.rb
```diff
@@ -22,9 +22,16 @@
 end
 
 Around do |_, block|
+  original_publish_token = ENV['CUCUMBER_PUBLISH_TOKEN']
+  ENV.delete('CUCUMBER_PUBLISH_TOKEN')
```
olleolleolle, line 26: The value deleted from the hash is returned, so you can assign it:

```ruby
  original_publish_token = ENV.delete('CUCUMBER_PUBLISH_TOKEN')
```

olleolleolle (commented):
Just poking around

luke-hill (commented):
Seems fine (olle offered a slight optimisation).

One thing (Possibly one for @vincent-psarga), we mentioned a while back that `Around` hooks were likely to go. Is it potentially worth whilst we're here, popping these in the `Before` / `After`

I don't know if we're still looking to get rid of Around hooks. But I know we were around 3.x

features/lib/support/env.rb
aurelien-reeves: done ✅

aurelien-reeves:
> One thing (Possibly one for @vincent-psarga), we mentioned a while back that `Around` hooks were likely to go. Is it potentially worth whilst we're here, popping these in the `Before` / `After`
> 
> I don't know if we're still looking to get rid of Around hooks. But I know we were around 3.x

Ok, just let me know

romaingweb (approved):

romaingweb:
Would it make sense to do those changes (from `Around` to `Before` / `After`) later to not block this fix?

luke-hill:
Oki go nuts. Vincent can catch-up later if needs be.

PR #1514, opened 2021-03-29
Allow attaching path

### LINKED ISSUE #1506: Allow attaching a path
It's common that users pass a file path (usually for a png or jpg file) as the first argument to the `attach` method. This leads to attachments that cannot displayed. One of the reasons users attach paths is that tools like Capybara's `save_screenshot` method returns a path instead of image data. Many users don't realise they have to `IO.read` the path before passing it to `attach`.

I think the `attach` method should behave as follows:

If the argument is a string, and the file exists, read the file and attach that.
Use a file extension to media type lookup to determine the media type. If a media type was specified, throw an error if the specified media type is different from the inferred one.

aurelien-reeves:

Fixes #1506

We plan to allow attaching files by specifying their path
Cucumber will take care of opening and reading it.

### DIFF
--- a/lib/cucumber/glue/proto_world.rb
+++ b/lib/cucumber/glue/proto_world.rb
@@ -114,10 +114,10 @@ module Cucumber
       end
 
       # Attach a file to the output
-      # @param file [string|io] the file to attach. 
+      # @param file [string|io] the file to attach.
       #   It can be a string containing the file content itself,
       #   the file path, or an IO ready to be read.
-      # @param media_type [string] the media type. If file is a valid path, 
+      # @param media_type [string] the media type. If file is a valid path,
       #   media_type can be ommitted, it will then be inferred from the file name.
       def attach(file, media_type = nil)
         return super unless File.file?(file)
--- a/lib/cucumber/glue/proto_world.rb
+++ b/lib/cucumber/glue/proto_world.rb
@@ -115,7 +115,7 @@ module Cucumber
       def attach(file, media_type = nil)
         return super unless File.file?(file)
 
-        content = File.open(file, 'rb', &:read)
+        content = File.read(file, mode: 'rb')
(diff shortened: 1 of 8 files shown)


luke-hill:
Make sure to add a path to a file auto-detects (Maybe based off extension??)

You could have an outline with
```
foo.png
bar.txt
baz.md
bay.jpeg
```
e.t.c.

aurelien-reeves:
> 
> 
> Make sure to add a path to a file auto-detects (Maybe based off extension??)
> 
> You could have an outline with
> 
> ```
> foo.png
> bar.txt
> baz.md
> bay.jpeg
> ```
> 
> e.t.c.

Yeah, that is interesting indeed 👍 
At first I'll push a basic feature file with no auto-detection specified for a first iteration. Once it passes, we will be able to iterate on that auto-detection ideas

aslakhellesoy:
This looks great @aurelien-reeves! Nice and simple.

Reading your code I had an idea that we might be able to add at the same time: Make the media type optional when you're passing an existing path. We could use a [library](https://github.com/mime-types/ruby-mime-types) to look up a media type for a file extension.

lib/cucumber/glue/proto_world.rb
```diff
@@ -111,8 +112,13 @@ def log(*messages)
         messages.each { |message| attach(message.to_s.dup, 'text/x.cucumber.log+plain') }
       end
 
-      def attach(file, media_type)
-        super
+      def attach(file, media_type = nil)
+        return super unless File.file?(file)
+
+        content = File.open(file, 'rb', &:read)
```
olleolleolle, line 118: Would it be possible to use this here?

```suggestion
        content = File.read(file, mode: 'rb')
```

https://rubyapi.org/2.5/o/io#method-c-read

aurelien-reeves: I think so, yes 👍

PR #1517, opened 2021-03-31
Allow debugging with vscode

aurelien-reeves:
Enabling debugging with vscode

https://user-images.githubusercontent.com/24386780/113124870-4802a200-9216-11eb-84b2-b408e3c2a799.mp4

What do you think?


aurelien-reeves:
I've replaced dedicated `.vscode` configruation with documentation dedicated to VSCode users.

Setting-up debugging for ruby code in VSCode is still a bit tedious. That's why I suggest that kind of documentation, to help on-boarding potential new contributors.

What do you think?

vincent-psarga:
looks good to me :)
To be honest, I don't use VSCode as an IDE (I mainly use it like an Emacs with tabs :D I still run the code through the terminal), so if it works for you with VSCode I guess it's good :)

aurelien-reeves:
In order to be able to properly execute rspec from vscode, I've added its binstub.
Is it an issue?

PR #1519, opened 2021-04-06
Gherkin 18x

aslakhellesoy:
Automatically use gems on the local file system. Upgrade to the latest monorepo gems.


aurelien-reeves:
For developers, and on the CI, we never use the real depencencies, but local code, or the one in github - the one from the master branches I presume

That mean we never really test the actual gem with its real dependencies?

I would suggest to move the development dependencies in a `Gemfile.local`, as documented here: https://github.com/cucumber/cucumber-ruby/blob/master/CONTRIBUTING.md#installing-your-own-gems. We could complete the contributing doc and provide a `Gemfile.local.monorepo.sample` with the changes you have done in that PR to make the tests with local code easier

aslakhellesoy:
> That mean we never really test the gem with its actual dependencies?

I've just pushed some changes to make sure we also have jobs that run against released dependencies (currently only with ruby 2.5 - when this works we can do the same for all ruby version).

I think it's important to have CI jobs that do both:
* Unreleased dependencies will highlight problems early
* Released dependencies give us the confidence that we can release

WDYT?

aurelien-reeves:
> Unreleased dependencies will highlight problems early

I think checking with unreleased dependencies may be part of nightly builds instead of continuous ones

mattwynne:
OK I released `cucumber-ruby-core` 9.0.0 with updated dependencies, including cucumber-messages and gherkin.

I now see two failures bundling locally using `CUCUMBER_RELEASED_GEMS` for the `cucumber-wire` package:

```
Bundler could not find compatible versions for gem "cucumber-cucumber-expressions":
  In Gemfile:
    aruba (~> 1.0, >= 1.0.4) was resolved to 1.0.4, which depends on
      cucumber (>= 2.4, < 6.0) was resolved to 4.1.0, which depends on
        cucumber-cucumber-expressions (>= 10.1.0, ~> 10.1)

    cucumber-wire was resolved to 4.0.1, which depends on
      cucumber-cucumber-expressions (~> 12.1, >= 12.1.1)

Bundler could not find compatible versions for gem "cucumber-messages":
  In Gemfile:
    cucumber-wire was resolved to 4.0.1, which depends on
      cucumber-core (~> 8.0, >= 8.0.1) was resolved to 8.0.1, which depends on
        cucumber-messages (>= 13.0.1, ~> 13.0)

    cucumber-html-formatter was resolved to 4.3.0, which depends on
      cucumber-messages (>= 10.0.3, ~> 10.0)
```

For the first one, this feels like a circular problem: we're depending on something (aruba) that depends on something (cucumber) that dependes on something (cucumber-expressions) that we depend on too. We can't upgrade cucumber yet, because it depends on us package.

Not sure what to do about that one.

The other one, `cucumber-html-formatter` is in the Gemfile, but I'm not really sure why it's there. What does HTML formatting have to do with this project? It was added in [this commit](https://github.com/cucumber/cucumber-ruby-wire/commit/453756a2f79d4a416a7086c436350e132dfdc53c).

aurelien-reeves:
Out of date

PR #1523, opened 2021-04-13
Remove deprecated `embed` and `puts`

aurelien-reeves:
Removal of deprecated code.
Code deprecated in 4.0.
We are about to release 6.0.
The removed code can be easily replaced with `attach` and `log` without any other change in the code.


mattwynne:
Looks good to me!

Issue #1521, opened 2021-04-12
Cucumber marks scenario as failed even though it passed on rerun if next step following the flaky one executes "skip_this_scenario" method


eduardrudko:
**Describe the bug**
skip_this_scenario method makes previouse flaky step to label whole scenario thus whole run as failed instead of flaky

**To Reproduce**
1. You need to create two scenariouse like this:
```
Feature: One green and one flaky scenario

  @test
  Scenario: Green scenario
    When green step

  @test
  Scenario: Flaky scenario
    When flaky step
    Then skip scenario step
    When green step
    When green step
```
Step Definitions:

```
When(/^flaky step$/) do
  if Random.rand(10) > 5
    fail
  end
end

When(/^green step$/) do

end

When(/^skip scenario step$/) do
  skip_this_scenario
end
```

2. Run test as following: 
```
bundle exec cucumber \ 
          --retry 2 \
          --no-strict-flaky \
          --no-source \
          features/gherkin_scenarios/test_scenariou.feature
```

**Expected behavior**

The scenario is marked as flaky if flaky step failed and reran sucessfully

**Context & Motivation**

The above issue makes us think the scenario has failed and never passed but in reality it has failed and passed on rerun.

**Your Environment**
 - Cucumber 3.2.0
 - Reproduced on LINUX and Mac Big Sur. Quite less likely it has to do with the versions of those.
 - CircleCi, also reproduced locally in bare shell

**Additional context**

Run with skip_this_scenario method being executed and flaky step passed from first attempt

```
bundle exec cucumber \ 
          --retry 2 \
          --no-strict-flaky \
          --no-source \
          features/gherkin_scenarios/test_scenariou.feature

Using the default profile...
Feature: One green and one flaky scenario

  @test
  Scenario: Green scenario
    When green step

  @test
  Scenario: Flaky scenario
    When flaky step
    Then skip scenario step
    When green step
    When green step

2 scenarios (1 skipped, 1 passed)
5 steps (3 skipped, 2 passed)
0m6.540s
echo $?                                                                                                      
0 🟢 
```

Run without skip_this_scenario method being executed but flaky scenario failed and reran sucessfully.

```
bundle exec cucumber \
          --retry 2 \
          --no-strict-flaky \
          --no-source \
          features/gherkin_scenarios/test_scenariou.feature

Using the default profile...
Feature: One green and one flaky scenario

  @test
  Scenario: Green scenario
    When green step

  @test
[...]

PR #1525, opened 2021-04-16
add scenario for skip before flaky step

eduardrudko:

https://github.com/cucumber/cucumber-ruby/issues/1521

No solution yet

**Additional context**
For now scenario is failing since we have bug there, once the fix is on place it should pass

### DIFF
--- a/features/lib/step_definitions/scenarios_and_steps.rb
+++ b/features/lib/step_definitions/scenarios_and_steps.rb
@@ -110,13 +110,13 @@ Given('a step definition that looks like this:') do |content|
   write_file("features/step_definitions/steps#{SecureRandom.uuid}.rb", content)
 end
 
-Given('a scenario {string} that fails three times, passes and then skips scenario') do |full_name|
+Given('a scenario {string} that fails once, passes and then skips scenario') do |full_name|
   name = snake_case(full_name)
 
   create_feature("#{full_name} feature") do
     create_scenario(full_name) do
       """
-        Given it three times, then passes
+        Given it fails once, then passes
         And it skips scenario
       """
     end
@@ -125,21 +125,6 @@ Given('a scenario {string} that fails three times, passes and then skips scenari
   write_file(
     "features/step_definitions/#{name}_steps.rb",
     step_definition('/^it skips scenario/', 'skip_this_scenario')
-  )
-
-  write_file(
-    "features/step_definitions/#{name}_steps.rb",
-    [
-      step_definition('/^it skips scenario/', 'skip_this_scenario'),
-      step_definition(
-        '/^it three times, then passes/',
-        [
-          "$#{name} ||= 0",
-          "$#{name} += 1",
-          "expect($#{name}).to be > 3"
-        ]
-      )
-    ].join("\n")
   )
 
   write_file(
--- a/features/lib/step_definitions/scenarios_and_steps.rb
+++ b/features/lib/step_definitions/scenarios_and_steps.rb
@@ -109,3 +109,41 @@ end
 Given('a step definition that looks like this:') do |content|
   write_file("features/step_definitions/steps#{SecureRandom.uuid}.rb", content)
 end
+
+Given('a scenario {string} that fails three times, passes and then skips scenario') do |full_name|
+  name = snake_case(full_name)
+
+  create_feature("#{full_name} feature") do
+    create_scenario(full_name) do
+      """
+        Given it three times, then passes
+        And it skips scenario
+      """
+    end
+  end
+
+  write_file(
+    "features/step_definitions/#{name}_steps.rb",
+    step_definition('/^it skips scenario/', 'skip_this_scenario')
+  )
+
+  write_file(
+    "features/step_definitions/#{name}_steps.rb",
+    [
+      step_definition('/^it skips scenario/', 'skip_this_scenario'),
+      step_definition(
+        '/^it three times, then passes/',
+        [
+          "$#{name} ||= 0",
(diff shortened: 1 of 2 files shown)


eduardrudko:
cc @mattwynne

features/lib/step_definitions/scenarios_and_steps.rb
```diff
@@ -109,3 +109,41 @@ def snake_case(name)
...
+  create_feature("#{full_name} feature") do
+    create_scenario(full_name) do
+      """
+        Given it three times, then passes
+        And it skips scenario
+      """
+    end
+  end
+
+  write_file(
```
luke-hill, line 125: this looks like a duplicate block

features/lib/step_definitions/scenarios_and_steps.rb
```diff
@@ -109,3 +109,41 @@ def snake_case(name)
...
   write_file("features/step_definitions/steps#{SecureRandom.uuid}.rb", content)
 end
+
+Given('a scenario {string} that fails three times, passes and then skips scenario') do |full_name|
+  name = snake_case(full_name)
+
+  create_feature("#{full_name} feature") do
+    create_scenario(full_name) do
+      """
+        Given it three times, then passes
```
luke-hill, line 119: you can call this `it fails three times then passes`

features/lib/step_definitions/scenarios_and_steps.rb
```diff
@@ -109,3 +109,41 @@ def snake_case(name)
...
+        [
+          "$#{name} ||= 0",
+          "$#{name} += 1",
+          "expect($#{name}).to be > 3"
+        ]
+      )
+    ].join("\n")
+  )
+
+  write_file(
```
luke-hill, line 145: This isn't needed because of your memoization in the step

luke-hill (commented):
Couple of little tips / optimisations

eduardrudko:
@luke-hill thanks for your input, all are fixed

mattwynne:
> @luke-hill thanks for your input, all are fixed

Nice one @eduardrudko, you can use the "Resolve conversation" button to close those comment threads now.

features/lib/step_definitions/scenarios_and_steps.rb
```diff
@@ -109,3 +109,26 @@ def snake_case(name)
...
   write_file("features/step_definitions/steps#{SecureRandom.uuid}.rb", content)
 end
+
+Given('a scenario {string} that fails once, passes and then skips scenario') do |full_name|
+  name = snake_case(full_name)
+
+  create_feature("#{full_name} feature") do
+    create_scenario(full_name) do
+      """
+        Given it fails once, then passes
```
mattwynne, line 119: So it looks like we still need to generate a step def for this one, right?

We need a way to make it fail once, but then pass on the second attempt?

We can't store state in the World (the scenario context) as we get a fresh instance every time the scenario runs, so I'd be tempted to write a step def something like this:

```
step_definition('/^it fails once, then passes/', 'raise "eek" unless $flag;$flag = true')
```

Do you want to give that a try?

mattwynne: Oh wait I see from the commit history you already tried something like that, sorry. Maybe I'm reading the code wrong but it seems like there's no step def for this step at the moment?

eduardrudko: It is actually. Every scenario has a `Background` statement where we create 3 step definition, one of them is used here.
What i'm starting to realise is that `it fails once, then passes` step def has its own ${name} variable. And when `it fails once, then passes` executes second time it isn't flaky anymore. Still the tests has the same fail reason as the "valid" one. I didn't really want to introduce 3rd flaky scenario e.g. fails 3 times and passes but for now i have no other idea. Any thoughts guys?

eduardrudko: @mattwynne let me come again:
Yes, there is already built-in step for `it fails once, then passes`. Inside this step definition there is `$#{name}` initialization then the flakiness achieves by  `"$#{name} += 1"` and `"expect($#{name}).to be > 1"`. Since this step always runs as a `Background` refering to it second time gives true here without failing`"expect($#{name}).to be > 1"` and does not simulates flakiness. This is why in order to use "built-in" step `it fails once, then passe` it is required to reset #{name} again, to do that i need to pass the same #{name} to both steps like this:
```
  Background:
    Given a scenario "Fails-once" that fails once, then passes
```
and
```
    Scenario: Flaky scenario gives zero exit code in no-strict-flaky mode if flaky step fails, passes and skips scenario
      And a scenario "Fails-once" that fails once, passes and then skips scenario
```

As for me it confuses and thats why it stopes me to use pre-built flaky scenario.

eduardrudko: To simplify all things, i would suggest to go for another similar step `it three times, then passes`, unless we decide to leave this test be as part of the suite. cc @mattwynne

mattwynne:
Nooo I didn't mean to delete this branch or close the issue - damn!

mattwynne:
Sorry @eduardrudko - can you push again or open another PR. So sorry - I was cleaning up stale branches and I must have got carried away.

mattwynne:
I came here to say - would you like to schedule some time to pair on this? You can [find a slot in my diary here](https://calendly.com/mattwynne/55-mins) or ping me in https://cucumber.io/community#slack and we can find a time that works better for you if those options don't work.

PR #1528, opened 2021-04-27
Remove deprecation for the JSON formatter

aurelien-reeves:
As discussed at our last community meeting, the JSON formatter is no more deprecated, but entering maintenance mode.

### DIFF
--- a/lib/cucumber/cli/options.rb
+++ b/lib/cucumber/cli/options.rb
@@ -24,7 +24,7 @@ module Cucumber
                                                               "#{INDENT}the usage formatter, except that steps are not printed."],
         'junit'       => ['Cucumber::Formatter::Junit',       "Generates a report similar to Ant+JUnit. Use\n" \
                                                               "#{INDENT}junit,fileattribute=true to include a file attribute."],
-        'json'        => ['Cucumber::Formatter::Json',        '[DEPRECATED] Prints the feature as JSON'],
+        'json'        => ['Cucumber::Formatter::Json',        'Prints the feature as JSON'],
         'message'     => ['Cucumber::Formatter::Message',     'Outputs protobuf messages'],
         'html'        => ['Cucumber::Formatter::HTML',        'Outputs HTML report'],
         'summary'     => ['Cucumber::Formatter::Summary',     'Summary output of feature and scenarios']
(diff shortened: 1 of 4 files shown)


luke-hill:
Possibly not for this PR. But should we put in the readme/formatter output which ones use legacy ruby code and which ones use protobuf?

I know with the migration to JSON schema this might not help. But it would show users which ones are generic and maintained in 1 location

aurelien-reeves:
I though about adding something like that somewhere, like the `docs/cli.md` from cucumber-js: https://github.com/cucumber/cucumber-js/commit/21c03499eb4772a6efef7cb8965de244eebc3dbc?

But did not found where to put that info.

Maybe as ruby-doc in the json formatter itself at first?

lib/cucumber/cli/options.rb
```diff
@@ -24,7 +24,7 @@ class Options
                                                               "#{INDENT}the usage formatter, except that steps are not printed."],
         'junit'       => ['Cucumber::Formatter::Junit',       "Generates a report similar to Ant+JUnit. Use\n" \
                                                               "#{INDENT}junit,fileattribute=true to include a file attribute."],
-        'json'        => ['Cucumber::Formatter::Json',        '[DEPRECATED] Prints the feature as JSON'],
+        'json'        => ['Cucumber::Formatter::Json',        'Prints the feature as JSON'],
```
luke-hill, line 27: I meant for example on this line, could we say something like (Does not use messages)

Again, if you feel this needs a wider discussion, maybe at Cukenfest tomorrow I can do.

aslakhellesoy: I don't think the users will be any wiser if we say what you suggest @luke-hill. I propose:

```
Prints the feature as JSON. The JSON format is in maintenance mode. Please consider using the message formatter instead.
```

aslakhellesoy (changes requested):
LGTM, just change the `--help` text to what I suggested.

lib/cucumber/cli/options.rb
aurelien-reeves: I would go further with something like:
`Prints the feature as JSON. The JSON format is in maintenance mode. Please consider using the message formatter with the standalone json-formatter instead (https://github.com/cucumber/cucumber/tree/master/json-formatter).`

I think that someone with the need for formatting as json has not the same needs as someone which would benefit from the messages directly.

Also we could add later a dedicated webpage for the json-formatter somewhere

luke-hill: Go for it. Both those options are 10* better than mine

PR #1531, opened 2021-05-01
Fix skip scenario retry

eduardrudko:

https://github.com/cucumber/cucumber-ruby/issues/1521

No solution yet

**Additional context**
For now scenario is failing since we have bug there, once the fix is on place it should pass


eduardrudko:
@mattwynne i've scheduled a session for us on March 4.

mattwynne:
> @mattwynne i've scheduled a session for us on March 4.

Gah sorry again I had to reschedule. I hope you can make the new time!

PR #1536, opened 2021-05-10
Fix 'attach' when passing null byte in the data

### LINKED ISSUE #1529: `attach` fails when data is an array with a null byte
`File#file?` raise an error when there is a null byte in the data, leading to the data being lost instead of being properly encoded and attached.

```ruby
When('an array with {int} bytes is attached as plain/text') do |size|
  data = (0..size-1).map {|i| [i].pack('C') }.join
  attach(data, "text/plain") # path name contains null byte (ArgumentError)
end
```

possible fix:
```ruby
      def attach(file, media_type = nil)
        return super unless File.file?(file)

        # snip...

      rescue
        return super
      end
```

aurelien-reeves:

Fixes #1529 

Rescue any standard error in the attach method, then call super without any further process around it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,14 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 
+- `attach` can now handle null bytes in the data.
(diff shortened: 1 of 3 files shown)


aurelien-reeves:
Thanks @luke-hill 

I've just commit fixes for your suggestions

PR #1549, opened 2021-06-03
Update rule feature to test using tags

luke-hill:
Allow Rules to be used in cucumber-ruby

### cucumber-ruby
- Fix the rule test to use a tagged rule

### gherkin
- Fix GherkinQuery to build up the backward looking tag logic when in `Rule` as well as `Feature` (Previously just used feature).

**Additional context**
Needs to be merged AFTER the gherkin fix.


luke-hill:
This will fail CI currently. Once I've then merged in the fix to gherkin it should then pass.

aurelien-reeves:
Could we have a scenario which tries filtering on the rule's tag?

aurelien-reeves (approved):

luke-hill:
I think you can see that in the "passing" commit, it did actually pass. I just forgot a line of the tag (Which I copied in the latter commit).

luke-hill:
Note to all: This requires a new version of gherkin-ruby (Slated to be v19.0.4), which in turn will require other things updating first. It is a hard-blocker that those items are released **before** this PR can be rebased and merged.

aurelien-reeves (approved):

PR #1554, opened 2021-06-18
remove reference to protobuf in formatters help

davidjgoss:


olleolleolle:
This is a helpful change, and the message is now more truthful.

olleolleolle (approved):
Looks good to me, please merge the change!

luke-hill (approved):
I had to check the logs to check we have actually got rid of them. Although it's not actually out cut as a version yet I don't believe.

aslakhellesoy:
The `main` branch still uses protobuf messages since #1554 isn't merged yet, but this change is still valid.

PR #1562, opened 2021-07-21
Remove dependency to cucumber-wire

aurelien-reeves:
# Description

`cucumber-ruby-wire` has been design to be an external optional dependency. This is not the case yet and it is responsible for adding complexity in the dependency tree for cucumber-ruby.

This PR aims to remove the dependency to `cucumber-ruby-wire`. The wire protocol would still be available but requiring it explicitly.

## Type of change

- Refactoring (improvements to code design or tooling that don't change behaviour)
- Breaking change (will cause existing functionality to not
  work as expected)

## Note to other contributors

Tests may be passing but there are a lot of things to do before merging this PR. We will certainly have to deprecate it first in a new major version before removing it completely in another new one.

For now what remains to be done here is to add a hook for `cucumber-ruby-wire` to be able to install by itself.

# Checklist:

Your PR is ready for review once the following checklist is
complete. You can also add some checks if you want to.

- [x] Tests have been added for any changes to behaviour of the code
- [x] New and existing tests are passing locally and on CI
- [x] `bundle exec rubocop` reports no offenses
- [x] RDoc comments have been updated
- [x] CHANGELOG.md has been updated
- [x] cucumber-ruby-wire is ready and work with this branch
- [x] The wire protocol has been deprectated

PR #1564, opened 2021-07-27
Deprecate usage of built-in cucumber wire protocol

aurelien-reeves:
# Description

This is related to https://github.com/cucumber/cucumber-ruby/pull/1562

It is planned to remove the built-in support for cucumber-ruby-wire. It will still be available but as an optional plugin.

Before removing the support for it - which is done in #1562 - we need to deprecate it. A deprecation process should last for 2 major releases at least: the once which announce the deprecation, and the one which drops the deprecation.

So that deprecation should be part of an incoming cucumber-ruby 8.0.0, for #1562 being part of 9.0.0

## Type of change

  work as expected)

# Checklist:

### DIFF
--- a/spec/cucumber/runtime_spec.rb
+++ b/spec/cucumber/runtime_spec.rb
@@ -20,5 +20,26 @@ module Cucumber
         expect(subject.doc_string('Text')).to eq 'Text'
       end
     end
+
+    describe '#install_wire_plugin' do
+      it 'informs the user it is deprecated' do
+        stub_const('Cucumber::Deprecate::STRATEGY', Cucumber::Deprecate::ForUsers)
+        allow(STDERR).to receive(:puts)
+        allow_any_instance_of(Configuration).to receive(:all_files_to_load).and_return(['file.wire'])
+
+        begin
+          subject.run!
+        rescue NoMethodError
+          # this is actually expected
+        end
+
+        expect(STDERR).to have_received(:puts).with(
+          a_string_including([
+            'WARNING: # built-in usage of the wire protocol is deprecated and will be removed after version 9.0.0.',
+            'See https://github.com/cucumber/cucumber-ruby-wire#migration-from-built-in-to-plugin for more info.'
+          ].join(' '))
+        )
+      end
+    end
   end
 end
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -3,6 +3,7 @@
 require 'fileutils'
 require 'cucumber/configuration'
 require 'cucumber/create_meta'
+require 'cucumber/deprecate'
 require 'cucumber/load_path'
 require 'cucumber/formatter/duration'
 require 'cucumber/file_specs'
@@ -261,7 +262,14 @@ module Cucumber
     end
 
     def install_wire_plugin
-      Cucumber::Wire::Plugin.new(@configuration, @support_code.registry).install if @configuration.all_files_to_load.any? { |f| f =~ /\.wire$/ }
+      return unless @configuration.all_files_to_load.any? { |f| f =~ /\.wire$/ }
+
+      Cucumber::Wire::Plugin.new(@configuration, @support_code.registry).install
+      Cucumber.deprecate(
+        'See https://github.com/cucumber/cucumber-ruby-wire#migration-from-built-in-to-plugin for more info',
+        ' built-in usage of the wire protocol',
+        '9.0.0'
+      )
     end
 
     def log


aurelien-reeves:
Once https://github.com/cucumber/cucumber-ruby-wire/pull/46 is merged, this one will be ready too I guess

mattwynne:
I think we can release the deprecation notice in a minor release, so 7.1 rather than 8.0. See the advice [here](https://semver.org/#how-should-i-handle-deprecating-functionality) from the semver docs:

> Before you completely remove the functionality in a new major release there should be at least one **minor** release that contains the deprecation so that users can smoothly transition to the new API.

(emphasis mine)

UPGRADING.md
```diff
@@ -0,0 +1,68 @@
...
+# features/support/wire.rb
+require 'cucumber/wire'
+```
+
+The wire protocol will be installed, and no deprecation message will be shown anymore.
+
+## Usage of AfterConfiguration hook
+
+The AfterConfiguration hook has a new parameter: `registry`. It is the instance of
+`Cucumber::Glue::RegistryAndMore` that has been instanciated for the `runtime`'s
```
luke-hill, line 51: instantiated*

spec/cucumber/runtime_spec.rb
```diff
@@ -20,5 +20,26 @@ module Cucumber
...
     end
+
+    describe '#install_wire_plugin' do
+      it 'informs the user it is deprecated' do
+        stub_const('Cucumber::Deprecate::STRATEGY', Cucumber::Deprecate::ForUsers)
+        allow(STDERR).to receive(:puts)
+        allow_any_instance_of(Configuration).to receive(:all_files_to_load).and_return(['file.wire'])
+
+        begin
+          subject.run!
```
luke-hill, line 31: There's probably a better way of doing this, but if we're deleting it soon it's not worth the hassle

CHANGELOG.md
```diff
@@ -18,10 +18,27 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
...
+
 ### Removed
 
 ### Security fixes
 
+### Deprecated
+
+- The built-in Wire protocol
+
+  The Wire protocol would still be officially supported, but as an optional plugin
```
luke-hill, line 34: The wire protocol is still officially supported, .....

luke-hill (approved):
Minor language stuff

spec/cucumber/runtime_spec.rb
aurelien-reeves: Yes, this won't last for long

I'm curious: what do you have in mind?

aurelien-reeves:
> 
> 
> I think we can release the deprecation notice in a minor release, so 7.1 rather than 8.0. See the advice [here](https://semver.org/#how-should-i-handle-deprecating-functionality) from the semver docs:
> 
> > Before you completely remove the functionality in a new major release there should be at least one **minor** release that contains the deprecation so that users can smoothly transition to the new API.
> 
> (emphasis mine)

The change of the "AfterConfiguration" hook: is it a breaking change? If so, that mean we should release it as 8.0.0?

mattwynne:
> The change of the "AfterConfiguration" hook: is it a breaking change? If so, that mean we should release it as 8.0.0?

OK I understand now. Yes, if we're changing the `AfterConfiguration` hook's parameter signature that's a breaking change so we'd need a major release.

I have some concerns about exposing that whole `Cucumber::Glue::RegistryAndMore` object. The reason it has that silly name is to make it clear that it [has too many responsibilties](https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/glue/registry_and_more.rb#L42). Many of the methods on that class are really internal Cucumber stuff, so I suggest we wrap it with a Facade to limit the surface area we're exposing to users.

aurelien-reeves:
> 
> 
> > The change of the "AfterConfiguration" hook: is it a breaking change? If so, that mean we should release it as 8.0.0?
> 
> OK I understand now. Yes, if we're changing the `AfterConfiguration` hook's parameter signature that's a breaking change so we'd need a major release.
> 
> I have some concerns about exposing that whole `Cucumber::Glue::RegistryAndMore` object. The reason it has that silly name is to make it clear that it [has too many responsibilties](https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/glue/registry_and_more.rb#L42). Many of the methods on that class are really internal Cucumber stuff, so I suggest we wrap it with a Facade to limit the surface area we're exposing to users.

Done
Deprecation planned for v8.0
Removal for v9.0

I've also added a new RegistryWrapper which exposes only what was needed by the wire plugin: `create_expression` and `current_world`

luke-hill (approved):
Seems good.

luke-hill:
On a side-note, probably not for here. We also should look into creating an alias for this for v8 - BeforeAll, and removing the name, because it's such a confusing and non-standard name.

aurelien-reeves:
> 
> 
> On a side-note, probably not for here. We also should look into creating an alias for this for v8 - BeforeAll, and removing the name, because it's such a confusing and non-standard name.

You mean, renaming the `BeforeAll` hook? (deprecated in v8, removed in v9?)

luke-hill:
Renaming it "to" `BeforeAll` yes.

Duplicate them in v8, in v9 remove AfterConfiguration

aurelien-reeves:
Ok got it, thanks

mattwynne:
> I've also added a new RegistryWrapper which exposes only what was needed by the wire plugin: `create_expression` and `current_world`

Great. Even those two methods look like very different responsibilities, and I'm very curious why we need to be passing around a reference to the `current_world` but I think that's a rabbit hole we don't want to go down right now. Limiting this API surface seem good enough for now to me.

aurelien-reeves:
> > I've also added a new RegistryWrapper which exposes only what was needed by the wire plugin: `create_expression` and `current_world`
> 
> Great. Even those two methods look like very different responsibilities, and I'm very curious why we need to be passing around a reference to the `current_world` but I think that's a rabbit hole we don't want to go down right now. Limiting this API surface seem good enough for now to me.

This is because `wire` instantiate a `Cucumber::StepMatch` object: https://github.com/cucumber/cucumber-ruby-wire/blob/c3b537064a73db67c8ad9e9dab905e4cfdf700bf/lib/cucumber/wire/protocol/requests.rb#L37

That object requires the `current_world` which it retrieves from the `registry`: https://github.com/cucumber/cucumber-ruby/blob/3b0385233284815e720aec4abc030fa4f96d8423/lib/cucumber/step_match.rb#L18

mattwynne:
> > > I've also added a new RegistryWrapper which exposes only what was needed by the wire plugin: `create_expression` and `current_world`
> > 
> > 
> > Great. Even those two methods look like very different responsibilities, and I'm very curious why we need to be passing around a reference to the `current_world` but I think that's a rabbit hole we don't want to go down right now. Limiting this API surface seem good enough for now to me.
> 
> This is because `wire` instantiate a `Cucumber::StepMatch` object: [cucumber/cucumber-ruby-wire/blob/main/lib/cucumber/wire/protocol/requests.rb#L37](https://github.com/cucumber/cucumber-ruby-wire/blob/main/lib/cucumber/wire/protocol/requests.rb#L37)
> 
> That object requires the `current_world` which it retrieves from the `registry`: [./lib/cucumber/step_match.rb#L18](https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/step_match.rb#L18)

Aye, I see that.

It just feels wrong for me for the `StepMatch` to be carrying around a (mutable) reference to the current world. It would make more sense to have whoever calls `args` pass in the `current_world` instance at the time it calls that method. But as I say, this is a rabbit hole I don't think we want to go down right now.

PR #1565, opened 2021-08-04
Move CI to GitHub actions

aurelien-reeves:
# Description

Moving the CI from Circle CI to Gihub actions.

Regular (ubuntu) builds: ✔
Windows builds: ✔
JRuby build: ✔
🆕 MacOS builds 🆕: ✔

The previous GitHub workflow was designed to push to cucumber reports, but it did not worked anymore since personal tokens have been deprecated on reports. It has been removed, we may consider recreating a dedicated workflow for that purpose if we want to, with a new collection token.

If we want to activate back the CCK tests, we could created a dedicated workflow too

I noticed that we are using a pretty old version of Rubocop, and updating it reveals a lot of offenses. I suggest to try upgrading it as part of another PR to avoid blocking this one.

## Type of change

# Checklist:


aurelien-reeves:
It seems that the build does not work on windows with ruby 2.5 for unknown reason.
Still, we have ruby 2.6, 2.7 and 3.0 which are running no Windows successfully. With circle-ci, we had only one build, and I can't even figure out which version of ruby it is.

olleolleolle (approved):
I am for going forward with this as it is now.

Rakefile
```diff
@@ -1,17 +1,26 @@
 require 'rubygems'
 require 'bundler'
+require 'rbconfig'
+# rubocop:disable Style/DoubleNegation
+is_windows = !!(RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/)
```
luke-hill, line 5: I'm aware this was copied from elsewhere, but a better one would be to use the .match? method

luke-hill (commented):
Should probably optimise this method whilst we're here.

Rakefile
aurelien-reeves: I think I'll take a look on this in another PR
That's why I've opened the one to update rubocop. With a more recent version of rubocop, we won't need this anymore

aurelien-reeves: Actually, I've just activated rubocop also on windows build, and it is working fine

luke-hill (approved):

PR #1566, opened 2021-08-04
Upgrade rubocop

aurelien-reeves:
# Description

Rubocop is fixed to a pretty old version.
The purpose of this PR is to try upgrading it step by step.

## Type of change

## Note to other contributors

One of the first offense to be fixed with a new version is done in `lib/cucumber/glue/dsl.rb`. Before we just had the `extend(Cucumber::Glue::Dsl)` instruction to extend the global Object class. This triggered an offense. I've fixed it by explicitly replacing it with the following:

```ruby
class Object
  # TODO: can we avoid adding methods to the global namespace (Kernel)
  prepend(Cucumber::Glue::Dsl)
end
```

That way the extension of the global Object class is made explicit. Tests seems to be passing. Does that looks good to advanced ruby developers?

# Checklist:


lib/cucumber/glue/dsl.rb
```diff
@@ -136,5 +136,7 @@ def register_rb_step_definition(regexp, symbol = nil, options = {}, &proc)
   end
 end
 
-# TODO: can we avoid adding methods to the global namespace (Kernel)
-extend(Cucumber::Glue::Dsl)
+class Object
```
luke-hill, line 139: You can do this, but part of the config file is where you specify overrides,

So you could ignore this rule on this file, and stipulate why

aurelien-reeves: It seems to work like that. Test are passing well. Is there some cases where it could fail?

luke-hill: The only difference for this would come in a situation whereby someone is using BOTH a user-defined version of these method / lVar names and the cucumber ones. We shouldn't really look to cover that situation.

But what I'm saying here, is instead of us just tricking rubocop, lets leave it as is, but document it in rubocop.yml by saying "This allows users to have nice DSL defined methods like ParameterType and Before in their support directory, instead of needing to instantiate new objects."

examples/i18n/ru/lib/calculator.rb
```diff
@@ -11,9 +11,11 @@ def result
     @stack.last
   end
 
+  # rubocop:disable Lint/BinaryOperatorWithIdenticalOperands
```
luke-hill, line 14: Instead of doing in-line stuff can we use the config file and comment why we're doing this.

- If we know why and we won't fix it, document why
- If we're not sure, put it into the todo file so we can tackle it later

aurelien-reeves: When I use in-line rubocop comment, that apply to a single line in the code. Is this possible with the rubocop config file?

Some comments may be missing.

luke-hill: I realise I missed this.

The reason we use the yml file (And potentially allow the exclude across the whole file), is it documents the behaviours in one place, so I know where to check.

Using the magic comments makes tracking down and debugging / optimising code a real pain point

aurelien-reeves:
Rubocop has been bumped to the latest version, and offenses have been fixed.

To review this PR, I think it would be easier to look at the commits rather than the diff.

@luke-hill What do you think? I removed all inline magic comment I had added before except for one.

lib/cucumber/cli/main.rb
```diff
@@ -14,7 +14,12 @@ def execute(args)
         end
       end
 
-      def initialize(args, _ = nil, out = STDOUT, err = STDERR, kernel = Kernel)
+      # The second argument is there for retrocompatibility
```
luke-hill, line 17: Not sure on how things work here, but given it's an unused argument I assume we plan on removing this in cucumber v8 ruby?

lib/cucumber/configuration.rb
```diff
@@ -189,7 +189,7 @@ def support_to_load
     def all_files_to_load
       files = require_dirs.map do |path|
         path = path.tr('\\', '/') # In case we're on windows. Globs don't work with backslashes.
-        path = path.gsub(/\/$/, '') # Strip trailing slash. # rubocop:disable Style/RegexpLiteral
+        path = path.gsub(%r{/$}, '') # Strip trailing slash.
```
luke-hill, line 192: We are allowed to use `/` inside the standard `//` regex matchers if we escape it. i.e. `/\//`

lib/cucumber/events/step_activated.rb
```diff
@@ -20,6 +20,8 @@ class StepActivated < Core::Event.new(:test_step, :step_match)
       def initialize(test_step, step_match)
         @test_step = test_step
         @step_match = step_match
+
+        super
```
luke-hill, line 24: Whilst this is a rubocop requirement, it is also an unsafe auto-correct. So I just wanted to check we have coverage for this?

lib/cucumber/events/step_definition_registered.rb
```diff
@@ -14,6 +14,8 @@ class StepDefinitionRegistered < Core::Event.new(:step_definition)
       # _@private
       def initialize(step_definition)
         @step_definition = step_definition
+
+        super
```
luke-hill, line 18: Same here

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -83,17 +83,17 @@ module ANSIColor
         next if method_name =~ /.*_param/
         code = <<-COLOR
           def #{method_name}(string=nil, &proc)
```
luke-hill, line 85: This is a really crappy way of doing this metaprogramming, we should be using define method, but maybe one for another time.

it also fixes the security eval offense (But can defer), maybe add this to the rubocop todo file

lib/cucumber/glue/proto_world.rb
```diff
@@ -38,9 +38,6 @@ def step(name, raw_multiline_arg = nil)
       #     And I am logged in as "Susan"
       #   }
       # @param [String] steps_text The Gherkin snippet to run
-      def steps(steps_text)
```
luke-hill, line 41: This is a biggie, removing this (I know we said we wanted to)

If we do want to, this should be a breaking change
Also we should probably move the RDoc above

lib/cucumber/glue/step_definition.rb
```diff
@@ -122,9 +122,8 @@ def ==(other)
 
       # @api private
       def arguments_from(step_name)
-        args = @expression.match(step_name)
+        @expression.match(step_name)
         # @registry.invoked_step_definition(regexp_source, location) if args
```
luke-hill, line 126: Can we delete this line?

lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -161,7 +161,7 @@ def hashes
       def symbolic_hashes
         @symbolic_hashes ||=
           hashes.map do |string_hash|
-            Hash[string_hash.map { |a, b| [symbolize_key(a), b] }]
+            string_hash.transform_keys { |a| symbolize_key(a) }
```
luke-hill, line 164: Are we confident that `string_hash` is always a hash?

lib/cucumber/multiline_argument/data_table.rb
```diff
@@ -284,18 +284,22 @@ def map_headers(mappings = {}, &block)
       #     end
       #   end
       #
+      # rubocop:disable Style/OptionalBooleanParameter # the optional boolean parameter is kept for retrocompatibility
       def map_column!(column_name, strict = true, &conversion_proc)
         # TODO: Remove this method for 2.0
```
luke-hill, line 289: Haha, Always love methods like this, Definitely should dive into removing this maybe>? In v8>?

lib/cucumber/runtime.rb
```diff
@@ -36,6 +36,7 @@ class FileNotFoundException < FileException
 
   class FeatureFolderNotFoundException < RuntimeError
     def initialize(path)
+      super
```
luke-hill, line 39: Could this affect the path assign (Usually safe supers go at the bottom)

spec/cucumber/cli/configuration_spec.rb
```diff
@@ -5,11 +5,11 @@
 
 module Cucumber
   module Cli
-    describe Configuration do
-      module ExposesOptions
-        attr_reader :options
-      end
+    module ExposesOptions
```
luke-hill, line 8: This now leaks into all potential other scenarios. Ideally we would write this as

```
let(:exposes_options) do
  Module.new do
    attr_reader :options
  end
end
```

spec/cucumber/formatter/http_io_spec.rb
```diff
@@ -6,15 +6,15 @@
 require 'spec_helper'
 require 'cucumber/formatter/io'
 
-RSpec.shared_context 'an HTTP server accepting file requests' do
-  module WEBrick
-    module HTTPServlet
-      class ProcHandler < AbstractServlet
-        alias do_PUT do_GET # Webrick #mount_proc only works with GET,HEAD,POST,OPTIONS by default
-      end
+module WEBrick
```
luke-hill, line 9: Same here but with Class.new (avoids rspec leakage)

spec/cucumber/formatter/junit_spec.rb
```diff
@@ -2,42 +2,41 @@
 
 require 'spec_helper'
```
luke-hill, line 3: This is a crazily complex file, so I've not reviewed it.

spec/support/fake_objects.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'cucumber/core/filter'
+
+# The following fake objects was previously declared within `describe` scope.
+# Declaring into scope did not isolate them.
+#
+# Moving those into a dedicated support file explitely state that those are available
```
luke-hill, line 8: `# Moving these into a dedicated support file explicitly states that these are available globally as soon as they are required once.`

spec/support/fake_objects.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'cucumber/core/filter'
```
luke-hill, line 3: Traditionally in spec support files you would use contexts and then allow extra contexts with let blocks described.

However given this is a large piece of work you've done, maybe I can build ontop of this as my first contribution back in a while?

luke-hill (commented):
So yeh reviewing this was hell! Please don't give me a PR with 158 things again to review 😓 

We should use the todo file to partially fix things up and then just click them off in the todo file.

lib/cucumber/cli/main.rb
aurelien-reeves: I am pro removing it indeed.
And as it is unused, maybe we could consider it was already deprecated, thus removing it in v8

lib/cucumber/configuration.rb
aurelien-reeves: Yes. But that is responsible for a rubocop offense.
Per default, Style/RegexpLiteral enforces using slashes, but do not allow inner slashes thus requires using %r with inner slashes. Cf. https://docs.rubocop.org/rubocop/1.0/cops_style.html#styleregexpliteral

I've let the default rubocop configuration here, but fix the ofenses without having magic comments.

If we prefer avoiding the use of %r, we may consider configuring the cop by allowing inner slashes.

lib/cucumber/events/step_activated.rb
aurelien-reeves: I took a look at the source code of the parent class: https://github.com/cucumber/cucumber-ruby-core/blob/main/lib/cucumber/core/event.rb

It seems to be safe here.

Don't know if we have coverage for this. But as this is a center piece of cucumber-ruby, I thing we may have notice issues.

lib/cucumber/events/step_definition_registered.rb
aurelien-reeves: Same response.

But yeah, trying to cover that would be helpful

lib/cucumber/formatter/ansicolor.rb
aurelien-reeves: Yeah, I just fix offenses. But that whole file ... I think it need a proper refactoring.
But for another PR?

spec/cucumber/cli/configuration_spec.rb
aurelien-reeves: Actually, it was already leaking into all other scenarios.
So the fix of the offenses make it actually explicit.

So, I guess we could also thing of refactoring that as part of another PR

spec/cucumber/formatter/http_io_spec.rb
aurelien-reeves: Same: it was actually already leaking to everything. The change has just made it explicit

spec/cucumber/formatter/junit_spec.rb
aurelien-reeves: This is actually the same kind of changes as with http_io_spec and configuration_spec. Nothing else

spec/support/fake_objects.rb
aurelien-reeves: Yes, I agree. I was pretty focus in trying to fix offenses without disabling cop.

That allow to show us more explicitly where good refactoring topics are.

I suggest to refactor as part of other PRs

lib/cucumber/glue/proto_world.rb
aurelien-reeves: I do not understand.
This is part of some kind of public API?
If we are just calling "super", why would there be issues?

lib/cucumber/glue/step_definition.rb
aurelien-reeves: The one that is commented?
If so, yes, I guess so :)

lib/cucumber/multiline_argument/data_table.rb
aurelien-reeves: Would the code do something else as it would have before?

lib/cucumber/multiline_argument/data_table.rb
aurelien-reeves: If any objection, agree

lib/cucumber/runtime.rb
aurelien-reeves: I'll check

aurelien-reeves:
> 
> 
> So yeh reviewing this was hell! Please don't give me a PR with 158 things again to review 😓
> 
> We should use the todo file to partially fix things up and then just click them off in the todo file.

Yeah, sorry for that 😓
I hate big PR myself.

aurelien-reeves:
So, to summarize:

What we could do as part of dedicated PR (I would open dedicated issues to make sure we do not forget those):

- removing the unused `_=nil` from `cli::main#initialize` parameter list in v8
- allowing inner slashes in RegexpLiteral rubocop configuration
- `formatter/ansicolor.rb` need to be refactored
- removing `DataTable#map_column!` in v8 which was planned to be removed in v2
- refactor fake modules and classes in `configuration_spec.rb`, `http_io_spec.rb`, `junit_spec.rb` and `fake_objects.rb` to prevent defining those fakes globally

To merge the current PR:

- make sure that calling `super` in the events `initialize` methods is safe as part of the actual PR (`lib/cucumber/events/step_activated.rb` and `lib/cucumber/events/step_definition_registered.rb`)
- can we safely delete some code from `proto_world.rb`? I am still confused with this. It seems to be useless. What would make it useful? `Cucumber::Glue#steps` and `Cucumber::Glue#inspect` are just calling `super` without any other code.

Did I missed something? Does that looks like a plan?

lib/cucumber/formatter/ansicolor.rb
luke-hill: Happy for that.

lib/cucumber/multiline_argument/data_table.rb
luke-hill: Yes. Previously this co-erced the Hash. This is technically a breaking change (if the previous input "could" be a non-hash)

spec/cucumber/cli/configuration_spec.rb
luke-hill: Oki I'm happy to tackle it elsewhere, but lets try tackle this sooner rather than later as it's a reasonably large covering fix.

luke-hill:
Yep sounds great. I've attached a couple of additional comments where needed.

spec/cucumber/cli/configuration_spec.rb
aurelien-reeves: Agree

lib/cucumber/multiline_argument/data_table.rb
aurelien-reeves: I've checked.
`hashes` will always be an array of hashes.
It uses the private method `build_hashes` which ends with `.map(&:to_hash)`

luke-hill (approved):

aurelien-reeves:
>   make sure that calling `super` in the events `initialize` methods is safe as part of the actual PR (`lib/cucumber/events/step_activated.rb` and `lib/cucumber/events/step_definition_registered.rb`)

Actually, [`Cucumber::Core::Event`](https://github.com/cucumber/cucumber-ruby-core/blob/main/lib/cucumber/core/event.rb) generates sub-classes with initializers dedicated to generate attribute readers.

So in derived classes as StepDefinitionRegistered or StepActivated, writing an initializer to initialize the attribute readers is doing what already does the base class.

Calling `super`is doing it twice.

Removing the initializer is removing duplicated code.

And this is actually covered with tests: when not properly initializing attribute readers for those two classes, some tests broke.

aurelien-reeves:
>    can we safely delete some code from `proto_world.rb`? I am still confused with this. It seems to be useless. What would make it useful? `Cucumber::Glue#steps` and `Cucumber::Glue#inspect` are just calling `super` without any other code.

I have restored the deleted methods. There purpose is actually documentation.

aurelien-reeves:
@luke-hill I guess I answer all your questions. Can we merge?

PR #1567, opened 2021-08-13
Add a release workflow

mattwynne:
# Description

Adds an automated release workflow.

## Type of change

## Note to other contributors

If your change may impact future contributors, explain it here, and remember to update README.md and CONTRIBUTING.md accordingly.

# Checklist:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,49 @@
+name: Release
+
+on:
+  push:
+    branches: [release/*]
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    environment: Release
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v2
+      - name: Read version to release from the changelog
+        id: next-release
+        uses: cucumber-actions/changelog-action@v1.3
+        with:
+          args: latest
+      - name: Set version environment variable
+        run: echo "version"=${{ steps.next-release.outputs.result }} > $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


mattwynne:
Reviewers, I'd particularly like feedback on what needs to be better documented so that this will make sense to someone who is new to it.

luke-hill (approved):
Again seems legit but would want to see it in action first. The one you showed me before didn't make sense.

olleolleolle (approved):
I find these added notes helpful!

Perhaps where we mention the CHANGELOG, we can note that the CHANGELOG.md _is_ used by automated tools, so that its formatting is of importance.

mattwynne:
> Again seems legit but would want to see it in action first. The one you showed me before didn't make sense.

Does this one make any more sense? What questions have you got?

mattwynne:
@mvz has given some excellent feedback [in Slack](https://cucumberbdd.slack.com/archives/C62GZFLLT/p1629182697016900?thread_ts=1628886639.015600&cid=C62GZFLLT). I think we should not merge this until the issue about tagging the Release commit has been resolved, as explained in [this comment](https://github.com/cucumber/common/issues/1688#issuecomment-899864133).

I think the fix is to create a new `create-release` action which we can test properly, as there are a few different strategies for determining the right commit to tag. It's too risky to only test these workflows when we're actually trying to do a release. This will remove duplication in the `release.yaml` workflows in the various repos, and ensure we handle it consistently.

PR #1569, opened 2021-08-30
Add BeforeAll and AfterAll hooks

aurelien-reeves:
# Description

Adding BeforeAll and AfterAll hooks. They are executed before all scenarios are executed and
after all scenarios have been executed.

Like the javascript ones, they do not have access to the `World` and they cannot mutate the test environment. Their purpose is to set-up and/or clean-up external things like databases or browsers.

I choose to keep the AfterConfiguration hook as-is. We could consider its purpose not totally the same: AfterConfiguration has access to the configuration, and can even update it. BeforeAll cannot. This is still opened to discussion :p

## Type of change

- This change requires an update of [cucumber.io/docs](https://cucumber.io/docs)

## Update required of cucumber.io/docs

If the [Cucumber documentation](https://cucumber.io/docs/) requires an update,
submit a PR to [cucumber/docs](https://github.com/cucumber/docs/) and
reference it here.

e.g. "Ref: cucumber/docs/pull/#99"

# Checklist:


aurelien-reeves:
So, I choose not to share the `World` with `BeforeAll` and `AfterAll` hooks. It was easier to implement, `cucumber-js` also does that, and it prevents the users trying to share things using the `World` that would actually be lost because - if I am not wrong - the World is newly instanciated for each scenario.

But still, BeforeAll and AfterAll could be used for globally settin-up and cleaning-up things like database, browsers, and things related to the subject under test.

What do you think?

aurelien-reeves:
Question: should we prevent invoking those hooks in dry-run?

luke-hill:
> Should these hooks have access to World

No they should not, it doesn't make sense and doesn't cause any issues

> I have kept AfterConfiguration as well as BeforeAll

That's fine, but in my eyes why don't we just move the logic from AfterConfiguration to BeforeAll - maybe over 2 majors. The concept of anything with the word `After` in it being ran first is just plain wrong imo.

> Should we prevent running hooks in dry run

Yes because dry-run should concern itself with things that a feature/scenario may or may not do (Not a suite). But I am not overly concerned with this

aurelien-reeves:
> 
> 
> > Should these hooks have access to World
> 
> No they should not, it doesn't make sense and doesn't cause any issues

👍

> 
> > I have kept AfterConfiguration as well as BeforeAll
> 
> That's fine, but in my eyes why don't we just move the logic from AfterConfiguration to BeforeAll - maybe over 2 majors. The concept of anything with the word `After` in it being ran first is just plain wrong imo.

So I would suggest to deprecate AfterConfiguration in favor of InstallPlugin rather than BeforeAll. InstallPlugin hook has the configuration object, BeforeAll has not.

> 
> > Should we prevent running hooks in dry run
> 
> Yes because dry-run should concern itself with things that a feature/scenario may or may not do (Not a suite). But I am not overly concerned with this

👍
I'll see what I can do

aurelien-reeves:
As part of this PR, I've added a short document to summarize all the hooks and their purpose.

Could you please review it? It is there: features/docs/writing_support_code/hooks/README.md

features/docs/cli/dry_run.feature
```diff
@@ -84,3 +84,34 @@ Feature: Dry Run
...
+
+  Scenario: With BeforeAll and AfterAll hooks
+    Given a file named "features/test.feature" with:
+      """
+      Feature:
+        Scenario:
+          Given this step fails
+      """
+    And the standard step definitions
+    And a file named "features/step_definitions/support.rb" with:
```
luke-hill, line 96: Can we also add (For my benefit), `Given "this step fails" { raise "Step fails" }` just to cover my OCD of having 2 diff issues

luke-hill:
2 second look looks good for docs, I'll have plenty of review comments but it'll be minor polishes. I'll get to reviewing it in the next day or so. (if you can hold out till then that'll be good).

features/docs/cli/dry_run.feature
aurelien-reeves: This is actually part of "the standard step definitions"
I've copy/paste that from a scenario before that one actually

As this is not the purpose of this scenario, we could replace that with "this step pass" if you want ?

aurelien-reeves:
> 
> 
> 2 second look looks good for docs, I'll have plenty of review comments but it'll be minor polishes. I'll get to reviewing it in the next day or so. (if you can hold out till then that'll be good).

No worries, no rush, take your time :)

features/docs/cli/dry_run.feature
luke-hill: Yes please do. That way we're not testing two things but one

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+[`AfterConfiguration`](#afterconfiguration) and [`InstallPlugin`](#installplugin)
+hooks are dedicated to plugins and are meant to extend Cucumber. For example,
+[`AfterConfiguration`](#afterconfiguration) allows you to dynamically change the
+configuration before the execution of the tests, and [`InstallPlugin`](#installplugin)
+allows to have some code that would have deeper impact on the execution.
+
+### AfterConfiguration
+
+**Note:** this is a legacy hook. You may consider using [`InstallPlugin`](#installplugin) instead.
```
luke-hill, line 35: Shall we put on here about it being deprecated, or are we doing that elsewhere>?

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
...
+```ruby
+AfterConfiguration do |configuration|
+  # configuration is an instance of Cucumber::Configuration defined in
+  # lib/cucumber/configuration.rb.
+end
+```
+
+### InstallPlugin
+
+In addition of the configuration, `IntallPlugin` also has access to some of Cucumber
```
luke-hill, line 46: In addition to the configuration, `InstallPlugin` ...

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
+# Cucumber Hooks
+
+Cucumber proposes several hooks to let you specify some code to be executed at
+different stage of the test execution, like before or after the execution of a
```
luke-hill, line 4: different stages of test execution,

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
...
+  - Per scenario:
+    - [Around](#around)
+    - [Before](#before-and-after)
+      - Per step:
+        - [AfterStep](#afterstep)
+    - [After](#before-and-after)
+- [AfterAll](#beforeall-and-afterall)
+
+You can define as many hooks as you want. If you have several hooks of the same
+types - for example, several `BeforeAll` hooks - they will be all executed once.
```
luke-hill, line 23: `<new para>`

Multiple hooks of the same type are executed in the order that they were defined. If you wish to control this order, use manual requires in `env.rb` (This file is loaded first), or migrate them all to one `hooks.rb` file.

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+You can see an example in the [Cucumber Wire plugin](https://github.com/cucumber/cucumber-ruby-wire).
+
+## BeforeAll and AfterAll
+
+`BeforeAll` is executed once before the execution of the first scenario. `AfterAll`
+is executed once after the execution of the last scenario.
+
+They have no parameter. Their purpose is to set-up and/or clean-up your environment
```
luke-hill, line 67: These two types of hooks have no parameters. Their purpose...

features/docs/writing_support_code/hooks/README.md
```diff
@@ -0,0 +1,123 @@
...
+Around do |scenario, block|
+  SomeDatabase::begin_transaction do # this is just for illustration
+    block.call
+  end
+end
+```
+
+## Before and After
+
+`Before` is executed before each scenario. `After` is executed after each scenario.
```
luke-hill, line 98: Lets use the `test_case` naming syntax here.

features/docs/writing_support_code/hooks/before_all_hook.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Scenario: A single BeforeAll hook
+
+    A BeforeAll hook will be invoked a single time before all the scenarios are
+    executed.
+
+    Given a file named "features/f.feature" with:
+      """
+      Feature: BeforeAll hook
+        Scenario: #1
```
luke-hill, line 17: Question. Is `#` in the scenario name permitted / going to cause problems.

features/docs/writing_support_code/hooks/before_all_hook.feature
```diff
@@ -0,0 +1,49 @@
...
+      """
+    And a file named "features/step_definitions/steps.rb" with:
+      """
+      hookCalled = 0
+
+      BeforeAll do
+        hookCalled += 1
+      end
+
+      Then /^the BeforeAll hook have been called$/ do
```
luke-hill, line 31: `has been called`

features/docs/writing_support_code/hooks/README.md
aurelien-reeves: I would not deprecate it yet, just discourage it. At least for the moment.

But maybe I over-think this 🙇‍♂️

features/docs/writing_support_code/hooks/before_all_hook.feature
aurelien-reeves: I'll check that 😅

aurelien-reeves: I review the scenario, and made some tests.
the `#` character seems to be properly supported in scenario names

features/docs/writing_support_code/hooks/README.md
aurelien-reeves: I'll create a dedicated PR for deprecating it for the next version 👌

features/docs/writing_support_code/hooks/README.md
```diff
@@ -43,7 +47,7 @@ end
 
 ### InstallPlugin
 
-In addition of the configuration, `IntallPlugin` also has access to some of Cucumber
+In addition to the configuration, `IntallPlugin` also has access to some of Cucumber
```
luke-hill, line 50: InstallPlugin

luke-hill (approved):

mattwynne:
> Should these hooks have access to World

There is a use-case for these hooks being able to set some kind of read-only initial state/context so that, for example, you could access the database connection you created in the `BeforeAll` hook from step defs. There are obviously other ways of doing this in Ruby, but if we can see a way to do this and avoid people having to use class props / global variables, that would be nice.

For example, we could pass a reference to a mutable `test_run_context` hash into this hook, which we then surface as a read-only hash in the `World` hook.

aurelien-reeves:
> There is a use-case for these hooks being able to set some kind of read-only initial state/context so that, for example, you could access the database connection you created in the `BeforeAll` hook from step defs. There are obviously other ways of doing this in Ruby, but if we can see a way to do this and avoid people having to use class props / global variables, that would be nice.
> 
> For example, we could pass a reference to a mutable `test_run_context` hash into this hook, which we then surface as a read-only hash in the `World` hook.
> 
> Of course, we can do that as a second iteration.

As you said, I suggest to do that as part of a second iteration, or even just waiting for user feedback and requests

PR #1571, opened 2021-09-27
jRuby build is failing with jRuby 9.3

aurelien-reeves:
# Description

    bundle exec rspec

Is resulting into a failing tests only with jruby 9.3: `spec/cucumber/glue/step_definition_spec.rb:92`

When executing only that test, or spec file, or even all the spec in `spec/cucumber/glue`, the test is properly passing.

I am running `rspec --bisect` at the moment. Still waiting for it to finish.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

Please add an entry to the relevant section of CHANGELOG.md as part of this pull request.

# Checklist:

Your PR is ready for review once the following checklist is
complete. You can also add some checks if you want to.

- [ ] Tests have been added for any changes to behaviour of the code
- [ ] New and existing tests are passing locally and on CI
- [ ] `bundle exec rubocop` reports no offenses
- [ ] RDoc comments have been updated
- [ ] CHANGELOG.md has been updated

PR #1572, opened 2021-09-27
Inform about non-support of JRuby 9.3

aurelien-reeves:
# Description

Refs. https://github.com/cucumber/cucumber-ruby/pull/1571

Due to issue with JRuby 9.3, I suggest to not support it until we have a solution

That would prevent delaying a potential release of cucumber-ruby


olleolleolle (approved):
Looks good!

PR #1578, opened 2021-10-12
Bring the new release process

aurelien-reeves:


aurelien-reeves:
> 
> 
> In order to avoid duplicating the release process - can you just make it look like this?
> 
> https://raw.githubusercontent.com/cucumber/cucumber-expressions/main/RELEASING.md

It is not entirely complete or right for ruby gems, so I would prefer to work on that as part of another task
Would you agree?

PR #1589, opened 2021-11-25
Refactor formatter/ANSIColor

### LINKED ISSUE #1583: Refactor formatter/ansicolor.rb
As part of #1566 we noticed that it could be nice to refactor `formatter/ansicolor.rb`

aurelien-reeves:
# Description

Fixes #1583 

## Type of change

# Checklist:

### DIFF
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -7,7 +7,7 @@ Cucumber::Term::ANSIColor.coloring = false if !$stdout.tty? && !ENV.key?('AUTOTE
 
 module Cucumber
   module Formatter
-    # This modules allows to format cucumber related outputs using ANSI escape sequences.
+    # This module allows to format cucumber related outputs using ANSI escape sequences.
     #
     # For example, it provides a `passed` method which returns the string with
     # the ANSI escape sequence to format it green per default.
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -85,12 +85,12 @@ module Cucumber
       ALIASES.each_key do |method_name|
         next if method_name =~ /.*_param/
 
-        define_method(method_name) do |string = nil, &proc|
-          apply_styles(ALIASES[method_name], string, &proc)
+        define_method(method_name) do |text = nil, &proc|
+          apply_styles(ALIASES[method_name], text, &proc)
         end
 
-        define_method("#{method_name}_param") do |string = nil, &proc|
-          apply_styles(ALIASES["#{method_name}_param"], string, &proc) + apply_styles(ALIASES[method_name])
+        define_method("#{method_name}_param") do |text = nil, &proc|
+          apply_styles(ALIASES["#{method_name}_param"], text, &proc) + apply_styles(ALIASES[method_name])
         end
       end
 
@@ -112,8 +112,8 @@ module Cucumber
 
       private
 
-      def apply_styles(styles, string = nil, &proc)
-        styles.split(',').reverse.reduce(string) do |result, method_name|
+      def apply_styles(styles, text = nil, &proc)
+        styles.split(',').reverse.reduce(text) do |result, method_name|
           send(method_name, result, &proc)
         end
       end
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -61,16 +61,14 @@ module Cucumber
         'tag'       => 'cyan'
       )
 
-      def set_custom_colors(colors)
+      def customize_colors(colors)
         colors.split(':').each do |pair|
           a = pair.split('=')
           ALIASES[a[0]] = a[1]
         end
       end
 
-      if ENV['CUCUMBER_COLORS'] # Example: export CUCUMBER_COLORS="passed=red:failed=yellow"
-        set_custom_colors(ENV['CUCUMBER_COLORS'])
-      end
+      customize_colors(ENV['CUCUMBER_COLORS']) if ENV['CUCUMBER_COLORS'] # Example: export CUCUMBER_COLORS="passed=red:failed=yellow"
 
       # Eval to define the color-named methods required by Term::ANSIColor.
       #
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -70,7 +70,7 @@ module Cucumber
 
       customize_colors(ENV['CUCUMBER_COLORS']) if ENV['CUCUMBER_COLORS'] # Example: export CUCUMBER_COLORS="passed=red:failed=yellow"
 
-      # Eval to define the color-named methods required by Term::ANSIColor.
+      # Define the color-named methods required by Term::ANSIColor.
       #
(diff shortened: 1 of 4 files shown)


aurelien-reeves:
Do we still need the `cukes` method in `lib/cucumber/formatter/ansicolor.rb`?

It looks like there were part of old warning mesages which are not displayed anymore: https://github.com/cucumber/cucumber-ruby/commit/05252afbc915c43114446e4975fd2c4983e5d83c

And they are unused at the moment

aurelien-reeves:
> Yes :D ❤️

Is this an answer to the question "do we remove those 'cukes' methods"?

luke-hill:
Sorry I meant this is the better way to do it (Define method vs eval).

aurelien-reeves:
I think we can let those `cukes` method for now

aurelien-reeves:
I know you already approved it
I pushed some documentation updates in the meantime. I just want to be sure it is right

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -29,26 +56,18 @@ module Formatter
...
-    #   ruby -e "require 'rubygems'; require 'term/ansicolor'; puts Cucumber::Term::ANSIColor.attributes"
+    #   ruby -e "require 'rubygems'; require 'cucumber/term/ansicolor'; puts Cucumber::Term::ANSIColor.attributes"
     #
     module ANSIColor
       include Cucumber::Term::ANSIColor
 
+      # :stopdoc:
       ALIASES = Hash.new do |h, k|
-        "#{h[Regexp.last_match(1)]},bold" if k.to_s =~ /(.*)_param/
+        next unless k.to_s =~ /(.*)_param/
```
luke-hill, line 68: We should probably use the match predicate here as it's more optimal 

See: https://github.com/JuanitoFatas/fast-ruby#stringmatch-vs-stringmatch-vs-stringstart_withstringend_with-code-start-code-end

JuanitoFatas repo is considered the holy grail for all ruby optimisations.

NB: This will **force us** to remain as 2.4+ (We are currently 2.5+), and not go back to 2.3+

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -82,54 +108,16 @@ module ANSIColor
       ALIASES.each_key do |method_name|
         next if method_name =~ /.*_param/
```
luke-hill, line 109: Here as well for optimisation

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -82,54 +108,16 @@ module ANSIColor
...
 
-        code = <<-COLOR
-          def #{method_name}(string=nil, &proc)
-            #{"#{ALIASES[method_name].split(',').join('(')}(string, &proc#{')' * ALIASES[method_name].split(',').length}"}
```
luke-hill, line 87: Is the method signature copied anywhere or have we removed this?

lib/cucumber/term/ansicolor.rb
```diff
@@ -2,9 +2,31 @@
...
+    #       $stdout.puts grey("Feature path:") + blue(bold(config.feature_dirs))
+    #     end
+    #   end
+    #
+    # To see what colours and effects are available, just run this in your shell:
+    #
+    #   ruby -e "require 'rubygems'; require 'cucumber/term/ansicolor'; puts Cucumber::Term::ANSIColor.attributes"
+    #
     module ANSIColor
+      module_function
```
luke-hill, line 28: We don't call this without an explicit mixin, and even if we did, we also use class << self below (Using both options is odd and quite messy).

spec/cucumber/formatter/ansicolor_spec.rb
```diff
@@ -29,6 +29,16 @@ module Formatter
 
         expect(passed('foo')).to eq 'foo'
       end
+
+      it 'works with custom colors' do
+        apply_custom_colors('passed=red,bold')
+        expect(passed('foo')).to eq "\e[31m\e[1mfoo\e[0m\e[0m"
+        apply_custom_colors('passed=green')
```
luke-hill, line 36: what is this line doing?

luke-hill (commented):
Quite a bit of this looks new and I didn't quite understand. So I've only reviewed part of it.

spec/cucumber/formatter/ansicolor_spec.rb
aurelien-reeves: It reset the colors.
Without this, other tests executed after that one may have the custom color kept to 'passed=red,bold'

lib/cucumber/formatter/ansicolor.rb
aurelien-reeves: Sorry, I don't get your question :(

spec/cucumber/formatter/ansicolor_spec.rb
luke-hill: Place that inside an after hook inside the relevant scope. Maybe inside a nicely named method like `reset_colours_to_default` or something

lib/cucumber/formatter/ansicolor.rb
luke-hill: I couldn't find this method signature, i.e. the split(',').join('(') e.t.c.

aurelien-reeves: I have rewritten that as part.

I had a hard time figuring out what was actually doing `.split.join * .split.lengh`.

I hope the new implementation is easier to understand!?

aurelien-reeves:
> Quite a bit of this looks new and I didn't quite understand. So I've only reviewed part of it.

It is supposed to be easier to understand than it was before actually 😓

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -7,17 +7,44 @@
 
 module Cucumber
   module Formatter
-    # Defines aliases for coloured output. You don't invoke any methods from this
-    # module directly, but you can change the output colours by defining
-    # a <tt>CUCUMBER_COLORS</tt> variable in your shell, very much like how you can
-    # tweak the familiar POSIX command <tt>ls</tt> with
-    # <a href="http://mipsisrisc.com/rambling/2008/06/27/lscolorsls_colors-now-with-linux-support/">$LSCOLORS/$LS_COLORS</a>
+    # This modules allows to format cucumber related outputs using ANSI escape sequences.
```
mattwynne, line 10: s/modules/module

```suggestion
    # This module allows to format cucumber related outputs using ANSI escape sequences.
```

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -7,17 +7,44 @@
...
+    #   class MyFormatter
+    #     extend Cucumber::Term::ANSIColor
+    #
+    #     def on_test_step_finished(event)
+    #       $stdout.puts undefined(event.test_step) if event.result.undefined?
+    #       $stdout.puts passed(event.test_step) if event.result.passed?
+    #     end
+    #   end
+    #
+    # This modules also allows the user to customize the format of cucumber outputs
```
mattwynne, line 30:
```suggestion
    # This module also allows the user to customize the format of cucumber outputs
```

mattwynne (approved):

PR #1603, opened 2022-01-04
Fix 1595: make sure to extend the world with a valid module

### LINKED ISSUE #1595: Namespaced worlds become nil after first scenario runs
**Describe the bug**
When namespaced Worlds are used, after the first scenario runs all namespaced worlds change from being a valid `Object` to being `nil`. This has no impact on pure methods, but stateful methods (e.g. `attr_accessor`) break in unclear ways. This is currently not caught by the worlds test suite (`features/docs/writing_support_code/world.feature`), as it does not test stateful methods, nor multi-scenario features.

**To Reproduce**

1. Create a step definition file containing the following definitions:

```rb
module ModuleOne
  attr_accessor :state
  def pure
    42
  end
end

World(module_one: ModuleOne)

Then /^module one is not nil$/ do
  expect(module_one).not_to be nil
end

Then /^state can be set$/ do
  expect { module_one.state = 42 }.not_to raise_error
end

Then /^pure method can be called$/ do
  expect { module_one.pure }.not_to raise_error
end
```
 
2. Create a feature file containing the following scenarios:

```cucumber
Feature: Multi-scenario namespaced worlds which keep state state                
  Scenario: 1. All checks pass
    Then module one is not nil 
    Then state can be set 
    Then pure method can be called
  
  Scenario: 2. Namespaced world becomes nil
    Then module one is not nil

  Scenario: 3. Stateful methods do not work
    Then state can be set 

  Scenario: 4. Pure methods still work
    Then pure method can be called
```

3. Run the tests, and observe the results for the following scenarios:
    - `1. All checks pass` passes
    - `2. Namespaced world becomes nil` fails with  `expected not #<NilClass:8> => nil   got #<NilClass:8> => nil`
    - `3. Stateful methods do not work` fails with `expected no Exception, got #<FrozenError: can't modify frozen NilClass: nil>`
    - `4. Pure methods still work` passes

**Expected behavior**
It is expected that all scenarios pass, regardless of execution order, as well as that namespaced worlds can maintain state per-scenario, via standard Ruby patterns such as `attr_accessor`.

**Context & Motivation**
Looking at using namespaced worlds to provide an up-front declaration of my test suites state, rather than relying on instance variables.

**Screenshots**
N/A

**Your Environment**
- ruby-cucumber, v7.1.0 & main (a7e68556)
- ruby, v3.0.0
- Ubuntu 20.04 LTS, running kernel 5.4.0

**Additional context**
This issue does not appear to be a problem with non-namespaced worlds, i.e. both scenarios would run if `World(ModuleOne)` was used, and the `module_one.` namespace dropped from the step definitions.

The behaviour can be resolved if all namespaced modules are explicitly undefined after each scenario (i.e. via an `After` hook). This was traced back to `lib/cucumber/glue/proto_world.rb` in `add_namespaced_modules!`, lines 189 - 193:

```rb
                inner_world = if self.class.respond_to?(namespace)                                                                                                                                                 
                                instance_variable_get(variable_name)
                              else
                                Object.new
                              end
```

It can also be resolved if the conditional is replaced with just `Object.new`.

I believe that after the first scenario executes, the instance variable `"@__#{namespace}_world"` is set to nil alongside all other global state, but is not reinitialized to a new valid Object. The modules methods are then defined on `nil`, which causes stateful methods to fail since `nil` is frozen.

aurelien-reeves:
# Description

Fixes #1595

Make sure to extend the World with a valid Object rather than extending it with something we think is valid but may not be.

## Type of change

# Checklist:

### DIFF
--- a/features/docs/writing_support_code/world.feature
+++ b/features/docs/writing_support_code/world.feature
@@ -95,6 +95,45 @@ Feature: World
     When I run `cucumber features/f.feature`
     Then it should pass
 
+  Scenario: A world is existing over multiple scenarios
+    Given a file named "features/support/helpers.rb" with:
+      """
+      module Helpers
+        def helper_method
+          42
+        end
+      end
+
+      World(my_namespace: Helpers)
+      """
+    And a file named "features/step_definitions/step.rb" with:
+      """
+      Then /^my_namespace is not nil$/ do
+        expect(my_namespace).not_to be_nil
+      end
+
+      Then /^the helper method is called$/ do
+        expect(my_namespace.helper_method).to eql(42)
+      end
+      """
+    And a file named "features/f.feature" with:
+      """
+      Feature: Calling a method
+        Scenario: The namespace exists
+          Then my_namespace is not nil
+        Scenario: I call a method from a namespace
+          Then the helper method is called
+        Scenario: The namespace still exists
+          Then my_namespace is not nil
+        Scenario: I still call a method from a namespace
+          Then the helper method is called
+      """
+    When I run `cucumber features/f.feature`
+    Then it should pass with:
+      """
+      4 scenarios (4 passed)
+      """
+
   Scenario: A world is extended using multiple modules with different namespaces
     Given a file named "features/support/helpers.rb" with:
       """
--- a/features/docs/writing_support_code/world.feature
+++ b/features/docs/writing_support_code/world.feature
@@ -121,17 +121,15 @@ Feature: World
       Feature: Calling a method
         Scenario: The namespace exists
           Then my_namespace is not nil
-        Scenario: I call a method from a namespace
-          Then the helper method is called
+          And the helper method is called
         Scenario: The namespace still exists
           Then my_namespace is not nil
-        Scenario: I still call a method from a namespace
-          Then the helper method is called
+          And the helper method is called
       """
     When I run `cucumber features/f.feature`
     Then it should pass with:
       """
-      4 scenarios (4 passed)
+      2 scenarios (2 passed)
       """
 
   Scenario: A world is extended using multiple modules with different namespaces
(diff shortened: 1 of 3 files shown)


features/docs/writing_support_code/world.feature
```diff
@@ -95,6 +95,45 @@ Feature: World
...
+        expect(my_namespace.helper_method).to eql(42)
+      end
+      """
+    And a file named "features/f.feature" with:
+      """
+      Feature: Calling a method
+        Scenario: The namespace exists
+          Then my_namespace is not nil
+        Scenario: I call a method from a namespace
+          Then the helper method is called
```
luke-hill, line 125: you can combine 125 onto the scenario above for optimisation

features/docs/writing_support_code/world.feature
```diff
@@ -95,6 +95,45 @@ Feature: World
...
+        Scenario: The namespace still exists
+          Then my_namespace is not nil
+        Scenario: I still call a method from a namespace
+          Then the helper method is called
```
luke-hill, line 129: This can be combined to the scenario above for optimisation

luke-hill (commented):
Had a load of other optimisations but I'm going to assume that `namespaced_world_modules` on line140 can be an empty array so the entire line142 invocation will be an empty array iterated (i.e. nothing)

features/docs/writing_support_code/world.feature
aurelien-reeves: Actually, the issue is reproduced when there are multiple scenarios.
If all steps were in a single scenarios, the original issue would not be triggered

features/docs/writing_support_code/world.feature
aurelien-reeves: Same as above

luke-hill (approved):

Issue #1604, opened 2022-01-05
Message "TestRunFinished" always report "success: false" whatever the result of the execution is


aurelien-reeves:
**Describe the bug**
The "TestRunFinished" message does not report "success: true" if the run was actually ok.
It is actually not implemented at all. See the `message_builder` formatter: it does not set the success for the TestRunFinished message: https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/formatter/message_builder.rb#L229

**To Reproduce**
Run `cucumber --format message`. Whatever your tests are failing or passing, `{"testRunFinished":{"success":false` is always reported.

**Expected behavior**
`{"testRunFinished":{"success":` depends on the execution and may be `true` if no failure is reported. It should match the result of the cucumber command.

PR #1606, opened 2022-01-07
Fix TestRunFinished.success message to return true when the run has passed

aurelien-reeves:
# Description

Fixes. #1604

## Type of change

# Checklist:


aurelien-reeves:
Sorry for the messed history. I guess it will be clean once merged

features/docs/formatters/message.feature
```diff
@@ -43,3 +43,28 @@ Feature: Message output formatter
...
+    Given a file named "features/steps.rb" with:
+      """
+      Given('a passed step') {}
+      Given('a failed step') { skip_this_scenario }
+      """
+    When I run `cucumber features/my_feature.feature --format message`
+    Then output should be valid NDJSON
+    And the output should contain:
+    """
+    {"testRunFinished":{"success":true
```
luke-hill, line 69: Qu) Is the state only true/false? so pending is classed as true? Could we maybe just make the failed step actually be a passing step just to make me feel better

features/docs/formatters/message.feature
```diff
@@ -43,3 +43,28 @@ Feature: Message output formatter
...
+    """
```
luke-hill, line 70: newline

lib/cucumber/runtime.rb
```diff
@@ -79,7 +79,7 @@ def run!
 
       receiver = Test::Runner.new(@configuration.event_bus)
       compile features, receiver, filters, @configuration.event_bus
-      @configuration.notify :test_run_finished
+      @configuration.notify :test_run_finished, !failure?
```
luke-hill, line 82: To avoid setting public in private can we just move this up to above private?

luke-hill (commented):
Minor queries

features/docs/formatters/message.feature
aurelien-reeves: Success is true or false, yes.
I took the same way the CLI use to exit with 0 or not.
Pending step are passing or failing depending the strict mode.

I can change the test as requested, yes 👌

features/docs/formatters/message.feature
aurelien-reeves: Where do you want a newline? Wouldn't you an indent for the docstring intead?

lib/cucumber/runtime.rb
aurelien-reeves: Don't get that comment, sorry 🤔

luke-hill: This method you're using (I think it's a method), is a hidden public/private method (As in it's actually public, but nestled way down deep inside the private methods, then toggled back to public).

Ideally as we're using it, lets fix it up and just make it public in the normal way.

features/docs/formatters/message.feature
luke-hill: In the viewing of the PR it says there isn't a newline (At the end of the entire file). Just a common convention

lib/cucumber/runtime.rb
aurelien-reeves: Got it, thanks 👌

luke-hill (approved):

PR #1608, opened 2022-01-17
Release 8.0.0.rc.1

aurelien-reeves:
# Description

Preparation for releasing v8.0.0 RC.1


aurelien-reeves:
The ruby 2.5 jobs are failing due to Nokogiri now requiring Ruby 2.6.
Indeed, [ruby 2.5 is EOL since 2021-03-31](https://www.ruby-lang.org/en/downloads/branches/)

Shall we:
- require ruby >= 2.6
- remove the ruby 2.5 job but keeping the requirement to 2.5 in the gemspec (so, 2.5 would be unofficially supported)
- ping nokogiri - which is a dev dependency - to a version which requires >= 2.5
- something else?

luke-hill:
Can I ask why we're hard-pinning everything to super up to date versions? Like I don't mind it, but is it necessary? We could just let people use nokogiri 1.12?

aurelien-reeves:
I would say that I would recommend to always use latest versions for dependencies (related to nokogiri), and discourage the use of not maintained ones (related to ruby 2.5).

The option to ping nokogiri to 1.12 in order to still support ruby 2.5 would not be the one I would go for 😅

But as I asked here publicly and requested review from you @luke-hill , @aslakhellesoy and @mattwynne , I will comply with the majority :D

mattwynne:
If Ruby 2.5 is EOL, let's stop supporting it officially. People can always pin to an older Cucumber if they're stuck on an older Ruby for some reason.

aurelien-reeves:
So, here's my proposal.
Any though @luke-hill @aslakhellesoy @mattwynne ?

luke-hill:
Will look at this, just bit busy for a day or so. If it's ok I'll review weds/thurs

aurelien-reeves:
I guess it can wait for a couple of days 🙂

cucumber.gemspec
```diff
@@ -18,37 +18,37 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency 'webrick', '~> 1.7', '>= 1.7.0'
 
   # For maintainer scripts
-  s.add_development_dependency 'octokit', '~> 4.21', '>= 4.21.0'
+  s.add_development_dependency 'octokit', '~> 4.22', '>= 4.22.0'
 
   # Needed for examples (rake examples)
   s.add_development_dependency 'rack-test', '~> 1.1', '>= 1.1.0'
   s.add_development_dependency 'sinatra', '~> 2.1', '>= 2.1.0'
   # Capybara dropped support for Ruby 2.3 in 3.16.0 - keep 3.15.0 while we're still supporting ruby 2.3
```
luke-hill, line 50: we can delete this comment

lib/cucumber/cli/rerun_file.rb
```diff
@@ -10,7 +10,7 @@ def self.can_read?(path)
       end
 
       def self.real_path(path)
-        path[1..-1] # remove leading @
+        path[1..] # remove leading @
```
luke-hill, line 13: Can we amend this comment above this, and be a bit more verbose (If I'm correct), saying.

`# This removes the leading @ from path`

It looks like maybe we need to fix something? Not sure.... Can we check

luke-hill (approved):
Approving notwithstanding some comment changes

lib/cucumber/cli/rerun_file.rb
aurelien-reeves: I've just removed the comment for now

PR #1611, opened 2022-01-29
Add Truffleruby support

gogainda:


lib/cucumber/formatter/backtrace_filter.rb
```diff
@@ -22,6 +22,7 @@ module Formatter
     @backtrace_filters << RbConfig::CONFIG['rubylibdir'] if RbConfig::CONFIG['rubylibdir']
 
     @backtrace_filters << 'org/jruby/' if ::Cucumber::JRUBY
+    @backtrace_filters << '<internal:core>' if RUBY_ENGINE == "truffleruby"
```
gogainda, line 25: @eregon do you think it's safe change? It does the job (all internal errors are filtered out and tests are passing), but later on I noticed that  some TypeError may dissapear as well, for example :
```
<internal:core> core/marshal.rb:51:in `__marshal__': singleton class can't be dumped (TypeError)
	from <internal:core> core/marshal.rb:1006:in `serialize'
	from <internal:core> core/marshal.rb:1309:in `dump'
	from -e:1:in `<main>'
```

aslakhellesoy: Could you please fix the rubocop error (use single quotes), and add truffleruby to the [GitHub Actions build matrix](https://github.com/cucumber/cucumber-ruby/blob/main/.github/workflows/cucumber-ruby.yml)?

gogainda: done

luke-hill:
@gogainda it seems as though the existing CI fails some tests. I haven't got much time yet to check, but can you see if these are ones that should be passing or if this is a truffleruby nuance.

lib/cucumber/formatter/backtrace_filter.rb
eregon: It should be any path starting with `<internal:`, that's also used by CRuby:
```
$ ruby -ve 'p method(:then).source_location'
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
["<internal:kernel>", 120]
```

Since it's converted to a Regexp later, then `'^<internal:'` is best.

aslakhellesoy:
> checking, tests have passed but the process failed with:
> 
> ```
> Error: Process completed with exit code 137.
> ```

Yes, looks like it: https://github.com/cucumber/cucumber-ruby/runs/4999431897?check_suite_focus=true#step:3:459

I think we need to get to the bottom of what's causing that because we merge this. We need to be confident that all tests pass on truffleruby.

gogainda:
Will close it for a while until I've found a root cause

gogainda:
@aslakhellesoy I've added additional workflow for rubocop job.

Since output for every version is the same , we want to run it only once and leave ruby matrix only for testing

gogainda:
created another PR for it https://github.com/cucumber/cucumber-ruby/pull/1612

PR #1621, opened 2022-04-05
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

> Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

https://www.legitsecurity.com/blog/github-privilege-escalation-vulnerability

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   rubocop:
     runs-on: ubuntu-latest


mpkorstanje:
Heya, thanks for all the links. I think I understand what you are trying to convey but I don't understand why you've only changed one file.

Are you looking to contribute or is this a bug report with example fix?

naveensrinivasan:
> Heya, thanks for all the links. I think I understand what you are trying to convey but I don't understand why you've only changed one file.
> 
> Are you looking to contribute or is this a bug report with example fix?

This is a fix for just one file, which I was able to do. You can consider this as an example fix.

aurelien-reeves:
Thanks @naveensrinivasan to point that permissions thing to us.

As we have a lot of repositories and more workflows in the cucumber organization, I guess we may consider configuring that at the organization level, and set more permissive permissions when this would be required. Do you think that would be a good thing? @mpkorstanje @mattwynne @aslakhellesoy what would be your opinion on that?

naveensrinivasan:
That is a good idea. But how do you override those permissions? That has to be done in the code. 

Also keeping this in the code will guarantee that even if those org permissions are changed this will be a fallback. 

This will also have a way of auditing changes. 

I'll let you decide. 👍

aurelien-reeves:
I close that one as we actually need to do such things at a higher level.

Still, thanks for pointing that out to us

PR #1624, opened 2022-04-13
Warn user when using space-separated string with cucumber_opts

### LINKED ISSUE #1614: Rake task fails with composite tag value option
**Describe the bug**

When cucumber rake task `cucumber_opts` field is a string and it contains composite tag option like `--tags 'not @skip'` cucumber rake task fails with `No such file or directory - @skip'. You can use `cucumber --init` to get started.` error.

**To Reproduce**
1. Create new cucumber rake task like this
   ```ruby
   desc 'Run features'
   Cucumber::Rake::Task.new(:features) do |t|
     t.cucumber_opts = "--tag 'not @skip'"
   end
   task :cucumber => :features
   ```
   Where `cucumber_opts` task property is set to a string value ends by composite tag option, `` for example
1. Run the rake task `bundle exec rake cucumber`
1. See error: ``No such file or directory - @skip'. You can use `cucumber --init` ``

**Expected behavior**

The rake task is running scenarios meets chosen tags criteria.
In the above example all scenarios except having `@skip` tag will run.

**Additional context**

Looks like the next split is a root cause of the issue: https://github.com/cucumber/cucumber-ruby/blob/3ebc762c3b9717e8593cd28f55ad356e287ebef6/lib/cucumber/rake/task.rb#L116

aurelien-reeves:
# Description

Fixes #1614 

## Type of change

### DIFF
--- a/lib/cucumber/rake/task.rb
+++ b/lib/cucumber/rake/task.rb
@@ -36,7 +36,7 @@ module Cucumber
         attr_reader :args
 
         def initialize(libs, cucumber_opts, feature_files)
-          raise 'libs must be an Array when running in-process' unless Array == libs.class
+          raise 'libs must be an Array when running in-process' unless libs.instance_of? Array
 
           libs.reverse_each { |lib| $LOAD_PATH.unshift(lib) }
           @args = (
@@ -113,7 +113,7 @@ module Cucumber
       attr_reader :cucumber_opts
 
       def cucumber_opts=(opts) # :nodoc:
-        unless String == opts.class
+        unless opts.instance_of? String
           @cucumber_opts = opts
           return
         end
--- a/lib/cucumber/rake/task.rb
+++ b/lib/cucumber/rake/task.rb
@@ -113,7 +113,15 @@ module Cucumber
       attr_reader :cucumber_opts
 
       def cucumber_opts=(opts) # :nodoc:
-        @cucumber_opts = String == opts.class ? opts.split(' ') : opts
+        unless String == opts.class
+          @cucumber_opts = opts
+          return
+        end
+
+        @cucumber_opts = opts.split(' ')
+        return if @cucumber_opts.length <= 1
+
+        $stderr.puts 'WARNING: consider using an array rather than a space-delimited string with cucumber_opts to avoid undesired behavior.'
       end
 
       # Whether or not to fork a new ruby interpreter. Defaults to true. You may gain
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -135,6 +135,9 @@ Style/AccessModifierDeclarations:
 Style/ClassAndModuleChildren:
   Enabled: false
 
+Style/ClassEqualityComparison:
+  Enabled: true
+
 Style/Documentation:
   Enabled: false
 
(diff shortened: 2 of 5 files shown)


valerianb (approved):

lib/cucumber/rake/task.rb
```diff
@@ -113,7 +113,15 @@ def run
       attr_reader :cucumber_opts
 
       def cucumber_opts=(opts) # :nodoc:
-        @cucumber_opts = String == opts.class ? opts.split(' ') : opts
+        unless String == opts.class
+          @cucumber_opts = opts
```
luke-hill, line 117: can just be return @cucumber_opts = opts

lib/cucumber/rake/task.rb
```diff
@@ -113,7 +113,15 @@ def run
       attr_reader :cucumber_opts
 
       def cucumber_opts=(opts) # :nodoc:
-        @cucumber_opts = String == opts.class ? opts.split(' ') : opts
+        unless String == opts.class
```
luke-hill, line 116: This should be opts.class == String (In rubocop it's referred to as yodacondition)

lib/cucumber/rake/task.rb
```diff
@@ -113,7 +113,15 @@ def run
...
+          @cucumber_opts = opts
+          return
+        end
+
+        @cucumber_opts = opts.split(' ')
+        return if @cucumber_opts.length <= 1
+
+        $stderr.puts 'WARNING: consider using an array rather than a space-delimited string with cucumber_opts to avoid undesired behavior.'
```
luke-hill, line 124: Generic question. For all of our warnings we emit, do we use stderr or do we use some custom logger (I can't remember). i.e. for the deprecation stuff is this also the way we do it?

luke-hill (commented):
Generic tidy-ups nothing breaking though

lib/cucumber/rake/task.rb
aurelien-reeves: I did that first but it triggered a rubocop offense: Lint/ReturnInVoidContext

lib/cucumber/rake/task.rb
aurelien-reeves: Yes, we use $stderr with all warnings (there are not that many actually :p)

lib/cucumber/rake/task.rb
aurelien-reeves: Will change it
I actually copy/paste that from somewhere else in the code. Will take a look if there are a lot like that or not, maybe change all at once

aurelien-reeves: Done

lib/cucumber/rake/task.rb
aurelien-reeves: With the deprecation helper, we default to $stderr, but there is the possibility to override it with a stream

lib/cucumber/rake/task.rb
luke-hill: Yeh I think it's because you have a writer method here. Rubocop actually ignores this issue and admits there's a problem. So I'm happy with what you've done for now. Definitely a bit ugly but if the ruby style guide can't agree, there's no reason for us to go off.

aurelien-reeves:
Thanks @luke-hill and @valerianb

PR #1629, opened 2022-04-24
Use `required_rubygems_version` instead of `rubygems_version`

mishina2228:
# Description

This is the same change made in https://github.com/cucumber/cucumber-rails/pull/533.

`rubygems_version` is used, probably a mistake for `required_rubygems_version`.
Because `rubygems_version` should not be set.
refs: https://github.com/rubygems/rubygems/blob/2faada63ae65d8559bcba9e083f3c35ba6c18fdc/lib/rubygems/specification.rb#L536-L541

## Type of change

# Checklist:


aurelien-reeves (approved):
Thanks 👍

PR #1636, opened 2022-05-18
Add support for named hooks

aurelien-reeves:
# Description

Add support for named hooks

This should pass once https://github.com/cucumber/common/pull/1984 is released

- [x] add some unit tests
- [x] update documentation
- [x] add changelog entry
- [ ] ~~optionally add it to reporters?~~


aurelien-reeves:
I wanted to update at least the json and consoles reporters, but it is really a difficult at the moment.

For now only message-based formatters properly reports hook's names. We could still update other formatters later if we have some requests to do so

luke-hill (approved):

PR #1641, opened 2022-05-21
Fix NoMethodError that is raised when the environment variable CUCUMBER_COLORS is defined

s2k:
# Description

This fixes a NoMethodError exception that is raised, when the environment variable `CUCUMBER_COLORS` is set, by allowing `apply_custom_colors` to be called right after its definition in `Cucumber::Formatter::ANSIColor`.

Defining the method as a class method of the module allows it to be called in the module itself.

## Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 
-* When the environment variable CUCUMBER_COLORS is set, its value is used instead of raising a NoMethodError.
+* Cucumber may raise NoMethodError when CUCUMBER_COLORS environment was set ([PR#1641](https://github.com/cucumber/cucumber-ruby/pull/1641/) [s2k](https://github.com/s2k))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 
+* When the environment variable CUCUMBER_COLORS is set, its value is used instead of raising a NoMethodError.
+
 ### Changed
 
 ### Removed
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -94,6 +94,17 @@ module Cucumber
       end
       apply_custom_colors(ENV['CUCUMBER_COLORS']) if ENV['CUCUMBER_COLORS']
 
+      # Provide an instance method of the same name, in case users are using it.
+
+      # Apply the custom color scheme
+      #
+      # example:
+      #
+      #   apply_custom_colors('passed=white')
+      def apply_custom_colors(colors)
+        ANSIColor.apply_custom_colors(colors)
+      end
+
       # Define the color-named methods required by Term::ANSIColor.
       #
       # Examples:
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -94,17 +94,6 @@ module Cucumber
       end
       apply_custom_colors(ENV['CUCUMBER_COLORS']) if ENV['CUCUMBER_COLORS']
 
-      # Provide an instance method of the same name, in case users are using it.
-
-      # Apply the custom color scheme
-      #
-      # example:
-      #
-      #   apply_custom_colors('passed=white')
-      def apply_custom_colors(colors)
-        ANSIColor.apply_custom_colors(colors)
-      end
-
       # Define the color-named methods required by Term::ANSIColor.
       #
       # Examples:
--- a/lib/cucumber/formatter/ansicolor.rb
+++ b/lib/cucumber/formatter/ansicolor.rb
@@ -86,7 +86,7 @@ module Cucumber
       # example:
       #
       #   apply_custom_colors('passed=white')
-      def apply_custom_colors(colors)
+      def self.apply_custom_colors(colors)
         colors.split(':').each do |pair|
           a = pair.split('=')
           ALIASES[a[0]] = a[1]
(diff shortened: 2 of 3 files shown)


s2k:
As a question for the reviewers: I wonder if there should be additional tests for this. While the existing tests had to be adapted to reflect the code change, I'm not (entirely) sure whether this is sufficient (given that the bug slipped through with the tests). Even though `CUCUMBER_COLORS='passed=red,bold' cucumber --help` would have revealed the bug, I'm not sure this would be a good fit for another scenario in a feature file.

aurelien-reeves (changes requested):
Thanks for your PR @s2k :)

Would it be possible to keep an instance methods - which would just call the class one - in order to prevent anything to break for users who would have hacked their cucumber?

And thus we could make sure to test the two methods actually

Also there is a call to apply_custom_colors right after it has been defined. I am surprised you did not had to change that call actually, but I may have missed something with ruby here 🤔

s2k:
Thanks for the feedback.

I had to change the definition of `apply_custom_colors` to be a class method, in order to be able to call it right after the definition. That's because the context is the module.

Here's an example:

```ruby
module Example
  def for_instances(arg)
    p arg
  end

  def self.for_class(arg)
    p arg
  end

  begin
    for_class('Calling the class method is OK')
  rescue NoMethodError => e
    p e
  end
  
  begin
    for_instances('The instance method is not available in the class')
  rescue NoMethodError => e
    puts "That didn't work"
    puts "Message:"
    puts "*" * 40
    puts e.message
    puts "*" * 40
  end
end
```

Running it yields:

```text
"Calling the class method is OK"
That didn't work
Message:
****************************************
undefined method `for_instances' for Example:Module

    for_instances('The instance method is not available in the class')
    ^^^^^^^^^^^^^
Did you mean?  for_class
****************************************
```

In other words: `self` inside a class (or module) definition is the class/module _itself_. To be able to call a method within the class/module definition it needs to be a class method (similar to `attr_reader` etc.).

I hesitate to provide an instance method, because the method is not meant to be called on an instance (but provided it anyway 🙂), so we can discuss it.

aurelien-reeves:
Thanks for the update.

@luke-hill what is your opinion on this?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 
+* When the environment variable CUCUMBER_COLORS is set, its value is used instead of raising a NoMethodError.
```
aurelien-reeves, line 19:
```suggestion
* Cucumber may raise NoMethodError when CUCUMBER_COLORS environment was set ([PR#1641](https://github.com/cucumber/cucumber-ruby/pull/1641/) [s2k](https://github.com/s2k))
```

lib/cucumber/formatter/ansicolor.rb
```diff
@@ -86,14 +86,25 @@ module ANSIColor
...
 
+      # Provide an instance method of the same name, in case users are using it.
+
+      # Apply the custom color scheme
+      #
+      # example:
+      #
+      #   apply_custom_colors('passed=white')
+      def apply_custom_colors(colors)
+        ANSIColor.apply_custom_colors(colors)
```
luke-hill, line 105: It's not strictly speaking good practice to invoke things direct on modules. Especially if they're not module functions.

luke-hill (changes requested):
If we have a mixture of things we need (or you want to permit both the class method and the instance method variant, your better bet is simply manipulating yourself from instance to class by doing `self.class.foo`

aurelien-reeves (approved):

spec/cucumber/formatter/ansicolor_spec.rb
```diff
@@ -36,7 +36,7 @@ module Formatter
 
       context 'with custom color scheme' do
         before do
-          apply_custom_colors('passed=red,bold')
+          ANSIColor.apply_custom_colors('passed=red,bold')
```
luke-hill, line 39: whilst I don't agree with it, can't we just do self.apply..... here as we've got included module in our block.

Ideally this should all be ran in an isolated non-leaking block, but that might be a lot of work for a tiny bugfix PR

aurelien-reeves:
> No worries, I'll just need to revert [1560c6e](https://github.com/cucumber/cucumber-ruby/commit/1560c6eb55ba94e1ed8a882f18c6f43c6d6a776b), right?

Yes, in addition to the last comments from @luke-hill

spec/cucumber/formatter/ansicolor_spec.rb
s2k: I tried that and it didn't work: Replacing the explicit `ANSIColor` with `self` (or `self.class` for that matter) still causes a `NoMethodError`. For the sake of completeness, I also tried to `extend ANSIColor` in the tests, but that didn't help either.

IIRC, methods defined in a `Module` become instance methods when the module i included, but module methods do not become class methods, when including the module.

In fact, I'd much rather solve this with a block, as @luke-hill suggests. (And leave this PR as is.)

aurelien-reeves: Right now, I would suggest to merge it as-is as it is working fine and it is actually fixing an issue.

We know that we have room for improvements related to ANSIColors. We can do that as part of another PR

@luke-hill what do you think?

luke-hill: The problem here is when you include the module those become included as instance methods on the class, but what we're doing here is running inside an rspec context - which is explicitly frowned upon. Ideally we should run in an isolated context. 

Also calling Module.method_name isn't a good practice as this doesn't work in reality, only when they're as either module functions or modified on the eigenclass of the module.

For now if we want to merge AS-IS and it fixes something go for it.

s2k: I'd like to note that I don't understand why exactly calling `self.class.apply_custom_colors` doesn't work in the RSpec code, since `Cucumber::Formatter::ANSIColor` is in the list of ancestors at the point the `before` is running.

s2k:
OK, as far as I can see we could now merge this PR, no?

diabolo:
This needs releasing. The bug with CUCUMBER_COLORS is in the current release 8.0.0 7 months after it has been fixed in  cucumber:main. Can we get a point release out to address this?

PR #1648, opened 2022-06-09
Simplify Gemfile so dependabot can parse it

### LINKED ISSUE #1637: Automatic dependency update PRs
### 🤔 What's the problem you've observed?

Renovate only looks at the `Gemfile` so most of our dependencies (which are in the `.gemspec`) are ignored.

### ✨ Do you have a proposal for making it better?

Take a look at https://github.com/marvin-bitterlich/renovate-gemspec-minimal-example and see if it works for us.

### 📚 Any additional context?

This is still an [open issue on renovate](https://github.com/renovatebot/renovate/issues/10616).

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/developer_experience.md)*

mattwynne:
# Description

I've been working on trying to resolve https://github.com/cucumber/cucumber-ruby/issues/1637 and trying a different tack using Dependabot rather than Renovate.

In order to get Dependabot to parse our dependencies, we need to simplify the Gemfile to plain text.

This seems to work, [based on a test I did in my fork](https://github.com/mattwynne/cucumber-ruby/network/updates/389434839). It's even created a couple of PRs (https://github.com/mattwynne/cucumber-ruby/pull/2, https://github.com/mattwynne/cucumber-ruby/pull/3).

The downside is that we lose this fancy ability to configure our development environments. To me this feels like a good trade-off, as we can still un-comment these lines locally if we need to work on a particular library in conjunction with this one. I went for having them commented out by default as I feel like that's the lowest barrier to entry for a new developer checking out the code.

Fixed #1637 

## Type of change

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,8 @@ gem 'rubocop', '~> 1.22', '= 1.30.1'
 gem 'rubocop-packaging', '~> 0.3', '= 0.5.1'
 
 # To hack on Cucumber together with any of these libraries, uncomment the line below:
-#gem 'cucumber-core', path: '../cucumber-ruby-core'
-#gem 'cucumber-cucumber-expressions', path: '../cucumber-expressions/ruby'
-#gem 'cucumber-gherkin', path: '../common/gherkin/ruby'
-#gem 'cucumber-html-formatter', path: '../html-formatter/ruby'
-#gem 'cucumber-messages', path: '../common/messages/ruby'
+# gem 'cucumber-core', path: '../cucumber-ruby-core'
+# gem 'cucumber-cucumber-expressions', path: '../cucumber-expressions/ruby'
+# gem 'cucumber-gherkin', path: '../common/gherkin/ruby'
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
To me, this is fine. 👍 Keeps the tools running.

aurelien-reeves:
> The downside is that we lose this fancy ability to configure our development environments.

Actually, it is better to rely on local gemfile to do this as explained in our contributing guide: https://github.com/cucumber/cucumber-ruby/blob/main/CONTRIBUTING.md#using-a-local-gemfile

aurelien-reeves:
Beside that, great news that we have a working fix to finally have a bot which automatically update our dependencies \o/

mattwynne:
> Actually, it is better to rely on local gemfile to do this

Ah, perfect!

PR #1686, opened 2022-12-23
Ran ./script/update-gemspec

mattwynne:
I'd love to be able to rely on dependabot but it doesn't seem to be working properly

To do:
- [x] drop Ruby 2.6 support (so Capybara will install)
- [ ] release CCK with new messages dependency
- [x] re-release Gherkin with new messages dependency


mattwynne:
We can't drop 2.6 until v9

PR #1695, opened 2023-01-06
Replace mime-types gem with mini_mime

orien:
## Description

[mini_mime](https://github.com/discourse/mini_mime) is a minimal mime type library that's more performant and uses less memory than the `mime-types` gem.

It has replaced `mime-types` in several projects:

* `rails` →  `actionmailer` → `mail` gem: https://github.com/mikel/mail/pull/1059
* `capybara` https://github.com/teamcapybara/capybara/pull/1884
* `httparty` https://github.com/jnunemaker/httparty/pull/769

Using it as a dependency of `cucumber`, downstream projects would be able to reuse the same dependency that is already available in most Rails apps, instead of pulling in an extra `mime-types` dependency.

### Type of change

- Refactoring (of dependencies) (improvements to code design or tooling that don't change behaviour)

## Checklist

### DIFF
--- a/cucumber.gemspec
+++ b/cucumber.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'cucumber-html-formatter', '>= 19.1', '< 21.0'
   s.add_dependency 'cucumber-messages', '>= 18', '< 22'
   s.add_dependency 'diff-lcs', '~> 1.5', '>= 1.5.0'
-  s.add_dependency 'mime-types', '~> 3.4', '>= 3.4.1'
+  s.add_dependency 'mini_mime', '~> 1.0'
   s.add_dependency 'multi_test', '~> 1.1', '>= 1.1.0'
   s.add_dependency 'sys-uname', '~> 1.2', '>= 1.2.2'
 
--- a/lib/cucumber/glue/proto_world.rb
+++ b/lib/cucumber/glue/proto_world.rb
@@ -3,7 +3,7 @@
 require 'cucumber/gherkin/formatter/ansi_escapes'
 require 'cucumber/core/test/data_table'
 require 'cucumber/deprecate'
-require 'mime/types'
+require 'mini_mime'
 
 module Cucumber
   module Glue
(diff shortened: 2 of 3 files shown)


luke-hill:
I will get to this. Sorry for delays. Just starting a new job and still configuring all my emails / IDE's e.t.c.

lib/cucumber/glue/proto_world.rb
```diff
@@ -92,7 +92,7 @@ def attach(file, media_type = nil)
         return super unless File.file?(file)
 
         content = File.read(file, mode: 'rb')
-        media_type = MIME::Types.type_for(file).first if media_type.nil?
+        media_type = MiniMime.lookup_by_filename(file)&.content_type if media_type.nil?
```
luke-hill, line 95: If you look above on line 82 you'll see a very common media type that we log is a custom one we've invented. I'd be surprised if this was one of the cached ones in mini-mime (The docs there state that this new gem is a lot slower for uncached ones).

orien: FYI: the `MiniMime.lookup_by_filename` call is only made if a `media_type` value _isn't_ provided to the `attach` method. It's there to fill in the gap if the type isn't known at ahead of time.

The example on line 82 will not result in a call to `MiniMime.lookup_by_filename`.

orien: I believe this is the use case:

https://github.com/cucumber/cucumber-ruby/blob/cd5f138416bd3307c406c0720ba04ba75cd6c123/features/docs/writing_support_code/attachments.feature#L33-L35

cucumber.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'cucumber-html-formatter', '>= 19.1', '< 21.0'
   s.add_dependency 'cucumber-messages', '>= 18', '< 22'
   s.add_dependency 'diff-lcs', '~> 1.5', '>= 1.5.0'
-  s.add_dependency 'mime-types', '~> 3.4', '>= 3.4.1'
+  s.add_dependency 'mini_mime', '~> 1'
```
luke-hill, line 30: can we lock this to 1.0 then it's rtm. Sorry about the confusion with the caching thing

orien: Awesome! Thanks.

I've fixed this with a rebase: [`e021f50`](https://github.com/cucumber/cucumber-ruby/compare/b635fa51fb635d122b70a680acbd355b035bac4d..e021f50052a1fc1721585b74a8991df507f8ac7e)

luke-hill (approved):

Issue #1702, opened 2023-04-03
Add source reference on `Envelope.parameterType`


jkronegg:
### 🤔 What's the problem you're trying to solve?

When implementing a plugin, I can determine the source reference of a Step Definition (using handler on `Envelope.stepDefinition`), but I cannot determine the source reference for a `ParameterType`.

The cause is `CachingGlue.emitParameterTypeDefined` which creates an `Envelope` with a `ParameterType` which miss the source reference information.

### ✨ What's your proposed solution?

A simple solution would be to:

- add a `SourceReference` in the `io.cucumber.messages.types.ParameterType`. Hint: requires changing the cucumber messages schema (https://github.com/cucumber/messages/tree/main/jsonschema) which generates classes for Java and other languages, see https://github.com/cucumber/messages/issues/145. A snapshot version of the library should be made.
- fill the `ParameterType` source reference when the `Envelope.parameterType` "event" is sent to the event bus, see https://github.com/cucumber/cucumber-jvm/issues/2719 for the corresponding Java feature.

### ⛏ Have you considered any alternatives or workarounds?

N/A

luke-hill:
Note to self. This is needed to fix `cck_spec > parameter-type > generates valid message structure spec`

luke-hill:
This is implemented and working in #1738

PR #1738, opened 2023-09-11
Make cucumber CCK compliant

luke-hill:
# Description

This PR aims to do 3 things
- [ ] 1) make cucumber ruby v9.0.2+ ~CCKv12~ CCKv13 (Thanks to bugfixes), compliant. We have let a **lot** slip by the wayside.
  - [x] `#attach` now has a 3rd optional parameter `filename` which can provide a filename when renaming things
  - [x] The `TestStepResult` message now has an additional 4th property of `exception` however in Ruby this is a partial duplication. This is definitely backwards compatible as we are not changing the existing error message property
  - [x] The `ParameterType` message now has an additional 6th property of `source_reference` which is a location object returning a URI (File location), and line (Line number), of where the parameter types **transformer** is located
    - [x] We need to amend messages to expose the transformer property (And tidy up some slightly inconsistent properties), on the `ParameterType` object during creation
- [x] 2) Clear up a couple of tiny "nuances" that hide lots of issues when running local vs remote. The primary one is CCK tests should always be runnable. This way we don't let things slip again
- [x] 3) Remove a couple of CI fixes for travis. We don't use travis

## Type of change

- New features

Changelog entry needs doing after everything has been merged.

This fixes a bunch of issues of interdependency hell. Everything "should" now be permissible with the latest versions of internals.

# Checklist:


botandrose:
@luke-hill looks like you've got your work cut out for you! If you want to delegate some of that out to me, I enjoy refactoring and etc, and I'd be happy to throw some hours at it if it results in a sooner v13!

luke-hill:
If you want to tackle refactoring e.t.c. feel free. But the timeline needs to be this way because the v12 cucumber core is a chunky amount of refactoring and then v13 will have a new breaking change for a boolean param.

The issue is the cucumber ruby repo is tracking an old cck version, and I've never been involved with the cck physically. So I need someone to coach me on how it works and how to fix the problem (We're currently 2 majors behind the cck spec and the cck is likely to be updated soon as well).

luke-hill:
@botandrose heads up this is likely to take a LOT longer than initially anticipated.

Consequently I might do some more refactoring / rubocop fixes as this will be several weeks when I get time to dive into this.

As such, if you want to also tackle tech debt / refactoring go for it.

lib/cucumber/glue/registry_and_more.rb
```diff
@@ -207,7 +207,17 @@ def parameter_type_envelope(parameter_type)
             name: parameter_type.name,
             regular_expressions: parameter_type.regexps.map(&:to_s),
             prefer_for_regular_expression_match: parameter_type.prefer_for_regexp_match?,
-            use_for_snippets: parameter_type.use_for_snippets?
+            use_for_snippets: parameter_type.use_for_snippets?,
+            source_reference: Cucumber::Messages::SourceReference.from_h(
+              # TODO: This needs fixing before we release the next version - this is simply put in to trick the
```
luke-hill, line 212: @jkronegg - If you want to mob on this or provide extra context / solutions before I'm next back feel free.

jkronegg: @luke-hill thanks for asking, but I'm not a ruby developer. What do you expect from me?

luke-hill: Ah sorry, nevermind then. I'll look into this later on. I thought you might have known what to do as you fixed the other ones I believe.

luke-hill: ping @botandrose if you want something technical to get your teeth into, this is something which I have half-fixed, but incorrectly. I haven't had much chance to fix this and I'm busy fixing a few other things in CCK / rails, so feel free to tackle if you want.

lib/cucumber/formatter/message_builder.rb
```diff
@@ -42,11 +42,12 @@ def output_message
         raise 'To be implemented'
       end
 
-      def attach(src, media_type)
+      def attach(src, media_type, filename = 'jeff')
```
luke-hill, line 45: Note to self, remove this when we're happy that it's passing through all the consumers

lib/cucumber/glue/registry_and_more.rb
botandrose: @luke-hill Wow, this is some meta meta stuff!

I took a peek and managed to get something working, but it ain't pretty:

```ruby
uri, line = *parameter_type.instance_variable_get(:@transformer).source_location
...
source_reference: Cucumber::Messages::SourceReference.new(
  uri: uri,
  location: Cucumber::Messages::Location.new(
    line: line
  )
)
```

If this is the right direction, we should probably push that first line into `Cucumber::Expressions::ParameterType#source_location`, or at least expose a `#transformer` reader.

luke-hill: Atm I'm doing some refactoring for a couple of days. I have some more commits on this branch. So if you'd like keep digging on your own, and then I'll push up an update here thurs/fri.

For clarity, this is to enable us to be more conformant to the new CCK (Which we're tracking heavily behind). If you check the person above they made changes in Java for this same thing here. See https://github.com/cucumber/cucumber-ruby/issues/1702 initially for the original issue and https://github.com/cucumber/cucumber-jvm/pull/2739/files for the code solution.

I'm not expecting this to be a big fix. But if you're interested to keep digging feel free. I'll be able to assist more next week.

lib/cucumber/glue/proto_world.rb
```diff
@@ -86,14 +86,16 @@ def log(*messages)
       # @param file [string|io] the file to attach.
       #   It can be a string containing the file content itself, the file path, or an IO ready to be read.
       # @param media_type [string] the media type.
-      # If file is a valid path, media_type can be omitted, it will then be inferred from the file name.
-      def attach(file, media_type = nil)
+      #   If file is a valid path, media_type can be omitted, it will then be inferred from the file name.
+      # @param filename [string] the name of the file you wish to specify.
+      #   BEHAVIOUR HERE TBD
```
luke-hill, line 91: Note to self, fix this before merging

lib/cucumber/glue/registry_and_more.rb
luke-hill: @botandrose I've fixed this.

luke-hill:
Note to self. CCKv13 is now primarily passing. CDATA issue is known about (Missing step def is now correctly flagged in error).

luke-hill:
### Note for watchers

cdata fix didn't make it into 13.0.2 of the CCK, but some more inconsistencies (Not failing ones), did.

13.0.2+ of the CCK is needed to make this PR green. ETA on getting all remaining reconciliation fixes is not yet known

lib/cucumber/formatter/message_builder.rb
```diff
@@ -192,10 +193,13 @@ def on_test_step_finished(event)
 
         result_message = result.to_message
         if result.failed? || result.pending?
+          message_element = result.failed? ? result.exception : result
```
luke-hill, line 196: @ehuelsmann / @davidjgoss - This is where I'm tripping up now and cannot explain it.

In the CCK from here (https://github.com/cucumber/compatibility-kit/commit/a3546f798a0caa09715e82027608e93b83f6d0a4#diff-7df09050f374e2ce8fc33ec3c05df9b40844714d7d622da7b1a56a73fd7f10bd) I can see for over 2 years we've never had a message property in a `PENDING` message. 


However..... We've had little coverage of PENDING messages - just that one scenario.

From the above you can see previously we were always treating pending steps here as requiring 3 properties (Correct on old schema - `status`, `duration` and `message`). This PR adds the fourth one for backwards compatibility.

However it seems as though something went awry in the CCK reconciliation updates I've been doing. But I cannot spot what. It seems as though the CCK **never** had this `message` property set (Last 2 years).

I probably need to explain it better. But not 100% sure where I've gone wrong / what's wrong

luke-hill:
ping @botandrose - Made some decent headway in last 2-3 weeks. Now looking back at this PR as something that can be fixed up and completed soon

davidjgoss (approved):

PR #1748, opened 2024-01-05
Port CCK logic

luke-hill:
# Description

Porting the CCK logic over to the gem proper. Will be requiring a major bump at CCK end (Will be a minor here)

Items to complete
- [x] Cucumber Ruby
  - [x] Changelog entry
  - [x] Refactor of existing code to use new CCK links
  - [x] In "first" iteration we need to ensure we're using the gem's feature files, but our step def, support code and miscellaneous files (Subsequently we should move to only have step/support code and use the remotely found files)
- [x] CCK
  - [x] Check we're not publishing any bespoke ruby code in the gem (Should only have the single namespace `CCK::Examples` with a few minor links to determine files
  - [x] Release Major CCK (v15?), which will cause this PR to start going green
## Type of change

  work as expected)
- If your change may impact future contributors, explain it here, and remember to update README.md and CONTRIBUTING.md accordingly.

# Checklist:


luke-hill:
Predictably this has raised a few issues @davidjgoss - Not sure when I'll get round to them. I think one requires a bug/hotfix in the CCK and maybe 1 of the features (attachments), has some oddities that don't play nicely when moving from 1 repo to another.

PR #1758, opened 2024-07-23
Opt-in for MFA requirement

tagliala:
Make the gem more secure by requiring that all privileged operations by any of the owners require OTP.

Ref: https://guides.rubygems.org/mfa-requirement-opt-in/

### DIFF
--- a/cucumber.gemspec
+++ b/cucumber.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
     'changelog_uri' => 'https://github.com/cucumber/cucumber-ruby/blob/main/CHANGELOG.md',
     'documentation_uri' => 'https://www.rubydoc.info/github/cucumber/cucumber-ruby/',
     'mailing_list_uri' => 'https://groups.google.com/forum/#!forum/cukes',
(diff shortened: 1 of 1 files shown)


olleolleolle:
The most important bit for maintainers will then be to check that the API key used **has been allowed to stand in for MFA**.

https://guides.rubygems.org/api-key-scopes/

> ENABLE MFA ON SPECIFIC API KEYS
> 
> If your account has MFA enabled on the UI only or UI and gem signin [authentication level](https://guides.rubygems.org/setting-up-multifactor-authentication/#authentication-levels), you have the option to enable MFA on a specific API key. This will require an OTP code for gem push, yank, owner --add/--remove commands.
> 
> You **can toggle this option when** creating or **editing an API key** on the UI. New API key with MFA enabled

luke-hill:
Because we've not updated rubocop here for a while. See https://github.com/cucumber/cucumber-rails/blob/main/.rubocop.yml#L18 for a brief comment ditty on why this isn't the case. This won't be something we have capacity to change soon

tagliala:
Ok thanks, got it, the current release process doesn't play well with MFA

PR #1761, opened 2024-09-01
Fix "Hash.new with kwargs" error in Ruby 3.4

amatsuda:
# Description

Calling Hash.new with keyword arguments has been removed in Ruby 3.4, and it's warned to "use Hash.new({ key: value }) instead" in Ruby 3.3.
Please see https://bugs.ruby-lang.org/issues/19236 for more details about this change.

So here's a patch that fixes the warning and error below:

Ruby 3.3
```
warning: Calling Hash.new with keyword arguments is deprecated and will be removed in Ruby 3.4; use Hash.new({ key: value }) instead
```

Ruby 3.4
```
unknown keywords: :strict, :proc (ArgumentError)
```

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ### Fixed
 - Fixed an issue where a change to one example in compatibility testing wasn't fully adhered to ([luke-hill](https://github.com/luke-hill))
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Thank you @amatsuda! Oh, wait #1757 does a different tack - perhaps this is a duplicate effort.

amatsuda:
Oh, indeed. I'm sorry for the noise!

PR #1768, opened 2024-11-19
Add funding_uri to gemspec

andrew:
I noticed that the project already has a FUNDING.yml. This PR adds the `funding_uri` to your gemspec to help increase visibility using the `bundle fund` command in Bundler.


andrew:
I can also open a similar PR for https://github.com/cucumber/cucumber-ruby-core if this is accepted.

luke-hill:
We actually have multiple sources for funding, but given this is an improvement it is easy to accept/merge

Thankyou!

andrew:
`funding_uri` only accepts one source unfortuntely, happy to add a different one if you have a preferred source?

PR #1806, opened 2025-12-11
v11/CCK v22 conformance

luke-hill:
# Description

Currently cucumber is v20 of the CCK conformant.

First step is to make it v22 conformant. Which requires adding in a bunch of new scenarios to the CCK testing, and then validating which items are still passing and what changes need to be made

## Type of change

  work as expected)

# Checklist:
### Current passing state 75/93

Items to check / fix (First pass)
- [x] Reversed order being executable from the CLI
- [ ] When a global hook fails. It seems to kill the entire process and not capture any messages
- [x] New assertion for consistent test run id does not seem to be always present#
- [x] `Attachment` is missing `timestamp`
- [ ] `TestStepResult` is missing some keys
- [x] The retry scenario is failing for an unknown reason

### DIFF
--- a/lib/cucumber/formatter/message_builder.rb
+++ b/lib/cucumber/formatter/message_builder.rb
@@ -26,6 +26,12 @@ module Cucumber
         @repository = Cucumber::Repository.new
         @query = Cucumber::Query.new(@repository)
 
+        @test_run_started_id = config.id_generator.new_id
+        @test_case_by_step_id = {}
+        @pickle_id_step_by_test_step_id = {}
+
+        # Ensure all handlers for events occur after all ivars are instantiated
+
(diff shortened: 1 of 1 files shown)


luke-hill:
Posting in here so there's a visible update outside of code

```
# CCK v22 conformance status update - Jan 2026
  # OVERALL: 93 examples, 15 failures, 78 passed
  # SANITIZED: 69 examples, 0 failures, 69 passed
```

luke-hill:
ping @brasmusson after all merges, re-analysis and tidy up. I found one new issue (Which will be auto-fixed once we are testing against CCK v29 - we're ahead of the game). I have now re-enabled a bunch of the previously failing CCK scenarios and we now have only 10 left to fix.

PR #1830, opened 2026-03-13
Print threads backtrace on SIGINFO/SIGPWR

sobrinho:
# Description

Debugging stuck builds is always tricky and this will add a thread backtrace dump to STDERR when the process is signaled with SIGINFO (macOS) or SIGPWR (linux).

_On macOS you can use CTRL+T to signal the running process. Linux you can use `kill -PWR pid`._

I have a PR open against RSpec as well [here](https://github.com/rspec/rspec/pull/314).

This is a common pattern used by [Sidekiq](https://github.com/sidekiq/sidekiq/blob/c2941506ddf5c8b691d0c6c318f454a76b0abf78/lib/sidekiq/cli.rb#L216-L225) and [Puma](https://github.com/puma/puma/blob/a1b5b5e7e1b8d34b7d24964f668733299be930a2/lib/puma/launcher.rb#L478-L489) that I'm aware of but possibly other projects as well.

## Type of change

# Checklist:


spec/cucumber/cli/main_spec.rb
```diff
@@ -65,6 +65,28 @@ def do_execute
             subject.execute!
           end
         end
+
+        thread_dump_signal = Signal.list['INFO'] || Signal.list['PWR']
```
luke-hill: One thing I'm mindful of since ruby 3.4/4.0 is auto-requiring of stdlib gems.

Is this something we need to add to a gemfile somewhere or add a require in for?

spec/cucumber/cli/main_spec.rb
```diff
@@ -65,6 +65,28 @@ def do_execute
...
+
+        context 'when interrupted with thread dump signal', skip: thread_dump_signal.nil? do
+          let(:runtime) { double('runtime').as_null_object }
+
+          it 'dumps the thread backtrace to the error stream' do
```
luke-hill: This assertion is checking 2 items. Can we split this out so we don't increase rubocop overhead

lib/cucumber/cli/main.rb
```diff
@@ -90,6 +91,26 @@ def trap_interrupt
...
+        signal = Signal.list['INFO'] || Signal.list['PWR']
+
+        return if signal.nil?
+
+        trap(signal) do
+          Thread.list.each do |thread|
+            prefix = "Thread TID-#{(thread.object_id ^ Process.pid).to_s(36)} #{thread.name || '<no name>'}"
+
+            if thread.backtrace&.any?
+              thread.backtrace.each do |line|
```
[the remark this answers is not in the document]
luke-hill: Actually looking at your implementation here. Backtrace will always be an array.

If it's 0 length, each will still functionally work. Is there a reason we want to artifically insert an additional line in?

spec/cucumber/cli/main_spec.rb
```diff
@@ -62,6 +62,28 @@ def do_execute
...
+        allow(runtime).to receive(:failure?).and_return(false)
+
+        expect(kernel).to receive(:exit).with(0)
+
+        subject.execute!(runtime)
+
+        tid = (Thread.current.object_id ^ Process.pid).to_s(36)
+
+        expect(stderr.string).to match(/Thread TID-#{tid} <no name> #{__FILE__}:#{__LINE__ - 11}:in (?:`kill'|'Process\.kill')/)
+      end
```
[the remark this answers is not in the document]
luke-hill: We have some logic for backtrace filtering. Try lean on that to avoid rewriting things we'll need to refactor

luke-hill:
@sobrinho if you're able to get this fixed up reviews wise and CI wise in the next few days I can commit to merge this and deploying this into v11. WDYT?

spec/cucumber/cli/main_spec.rb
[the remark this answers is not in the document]
sobrinho: Regex is fine, add the line var.

sobrinho:
@luke-hill fixed!

spec/cucumber/cli/main_spec.rb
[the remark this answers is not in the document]
luke-hill: Can we lean on the conditionals as previously mentioned. Whether it's a RUBY_VERSION >= check or a direct method

See `invoke_in_world`

Reason: When we remove ruby 3.4 as the lowest version it's easy to scan the codebase and remove all refs

spec/cucumber/cli/main_spec.rb
```diff
@@ -62,6 +62,31 @@ def do_execute
         subject.execute!
       end
     end
+
+    thread_dump_signal = Signal.list['INFO'] || Signal.list['PWR']
```
luke-hill, line 70: Can this be placed as a let

luke-hill (commented):
Just a couple of things to raise. Plus I ran the pipeline now

spec/cucumber/cli/main_spec.rb
sobrinho: Can't be a let, it's used on `skip: thread_dump_signal` to avoid running the scenario in environments that doesn't have either signal.

spec/cucumber/cli/main_spec.rb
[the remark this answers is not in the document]
sobrinho: Done.

spec/cucumber/cli/main_spec.rb
sobrinho: This works fine without requires in 4.0.1:

```
➜  cucumber-ruby git:(main) ✗ asdf set ruby 4.0.1 
➜  cucumber-ruby git:(main) ✗ ruby -v
ruby 4.0.1 (2026-01-13 revision e04267a14b) +PRISM [arm64-darwin25]
➜  cucumber-ruby git:(main) ✗ irb
irb(main):001> Signal
=> Signal
```

spec/cucumber/cli/main_spec.rb
sobrinho: You mean we have 2 `expect` here?

sobrinho:
I can't answer some of the comments anywhere since the code has changed.

About backtrace filtering, we don't want to filter anything, we want the raw backtrace for debugging purposes as the same of Sidekiq/Puma/Unicorn/RSpec. A hidden backtrace line could be the cause, for instance.

luke-hill (approved):

sobrinho:
@luke-hill looks like master is broken on CI as well. Would you check?

luke-hill:
@sobrinho this is a false negative. `rake` was broken for 13.4 - Quickly patched as 13.4.1 one day later

sobrinho:
Okay, let me check why JRuby and TruffleRuby broke and make a fix.

sobrinho:
I'm not sure why JRuby failed on CI, it is working just fine locally for both RSpec and Cucumber itself:

Screenshot 2026-04-17 at 14 41 23

Screenshot 2026-04-17 at 14 41 44

Running `bundle exec rake` also works just fine:

Screenshot 2026-04-17 at 14 48 35

There are broken tests (not related to this PR) but the process is finishing with 1 instead of 158 as the CI.

Should we care?

sobrinho:
For TruffleRuby, I fixed the spec, its backtrace is a bit different from MRI/JRuby (no original location).

luke-hill:
The CCK tests should not be failing any more. If they are - it implies your branch is not fully up to date (v11 released recently), has fixed up the CCK portion of the suite so it will either pass or not run

sobrinho:
@luke-hill updated the branch with HEAD. Would you fire the actions again?

sobrinho:
Interesting, JRuby is yet OOM'ing but runs fine locally.

We might need to raise the memory on GH Actions, if possible.

sobrinho:
Okay, wrong hypothesis, JRuby is failing to trap SIGPWR on Linux, looking for a solution.

sobrinho:
@luke-hill one more try? :)

sobrinho:
Now it is cucumber, working on it 🫠

sobrinho:
Never mind, the action was from the previous commit. Would you run the latest commit?

sobrinho:
Alright, last fix for Rubocop and we are good now! :)

sobrinho:
@luke-hill would you fire the actions again?

sobrinho:
@luke-hill looks good now!

luke-hill:
Been busy with some other training stuff recently. I will get to this soon I promise.

luke-hill:
Thanks for the contribution @sobrinho 

I'll tidy up the changelog so it's in unreleased. I'm not sure what the release cadence will look like in the coming month or so as there's lots of refactors I'm working on simultaneously.

PR #1848, opened 2026-04-07
V11: Deletions

luke-hill:
# Description

This will include every PR merged in that is of category v11 deletions
Each PR should have their own changelog (But some may have them done after the fact)
This should be merged in near the end of the v11 lifecycle so that we don't have a muddy setup in prep for a v12 release

## Type of change

  work as expected)
- This change requires an update of [cucumber.io/docs](https://cucumber.io/docs)
- If your change may impact future contributors, explain it here, and remember to update README.md and CONTRIBUTING.md accordingly.

Each PR should have their own changelog entry

# Checklist:


luke-hill:
I've missed a bunch of logic in the features. Apparently over 40 are failing. So this will likely need a PR to be merged into this to fix these up.

features/docs/cli/retry_failing_tests.feature
```diff
@@ -38,11 +38,11 @@ Feature: Retry failing tests
 
   Scenario: Retry twice, so Fails-twice starts to pass too
     When I run `cucumber -q --retry 2 --format summary`
-    Then it should pass with:
+    Then it should fail with:
```
luke-hill: @brasmusson shouldn't this be `pass`

brasmusson: @luke-hill Since the `Flaky` result (always) says `Ok?` is `false` it `fails` (but the scenario is reported as `flaky` instead of `failed` since it passed on the third try).

luke-hill: Note to self.

Discussed this.

Code-wise this is correct. Viability wise this is wrong. We need to amend `Flaky.ok?` to actually return `true` always, because flaky results are indeed "good"

PR #1857, opened 2026-04-12
Handle failures in BeforeAll hooks and AfterAll hooks

brasmusson:
# Description

* All BeforeAll hooks and AfterAll hooks shall be executed regardless of they pass or fail.
* The TestCases shall only be executed if all BeforeAll hooks passes.
* BeforeAll hooks shall be executed after the TestRunStarted event has fired.
* AfterAll hooks shall be executed before the TestRunFinished event is fired.
* Print the global hook failures shall be printed by the console formatters.
* Aslo global hook failures shall affect the exit code.

With this change Cucumber-Ruby conforms fully with CCK v22

## Type of change

# Checklist:


luke-hill (commented):
Once v11 is out I'll get to this. Might take me a week or two

lib/cucumber/formatter/message_builder.rb
```diff
@@ -193,6 +202,7 @@ def on_test_run_started(*)
 
       def on_test_case_started(event)
         @current_test_case_started_id = test_case_started_id(event.test_case)
+        @current_test_run_hook_started_id = nil
```
luke-hill: This shouldn't be needed

lib/cucumber/formatter/message_builder.rb
```diff
@@ -61,13 +61,22 @@ def initialize(config)
...
 
       def attach(src, media_type, filename)
-        attachment_data = {
-          test_step_id: @current_test_step_id,
-          test_case_started_id: @current_test_case_started_id,
-          media_type: media_type,
-          file_name: filename,
-          timestamp: time_to_timestamp(Time.now)
-        }
+        attachment_data = if @current_test_run_hook_started_id.nil?
```
luke-hill: To avoid crazy indents can we do something like this

```suggestion
        attachment_data =
          if @current_test_run_hook_started_id.nil?
            {
              .....
            }
```

lib/cucumber/glue/registry_and_more.rb
```diff
@@ -147,15 +147,25 @@ def install_plugin(configuration, registry)
       end
 
       def before_all
+        set_up_world_for_global_hooks
+        all_succeded = true
```
luke-hill: Typo *succeeded (In other spots too)

lib/cucumber/glue/registry_and_more.rb
```diff
@@ -186,9 +196,10 @@ def invoke_run_hook(hook, pseudo_method)
         begin
           hook.invoke(pseudo_method, [])
           @configuration.notify(:test_run_hook_finished, hook, Core::Test::Result::Passed.new(timer.duration))
+          true
         rescue StandardError => e
           @configuration.notify(:test_run_hook_finished, hook, Core::Test::Result::Failed.new(timer.duration, e))
-          raise
+          false
```
luke-hill, line 204: So this is where the magic happens/happened. Thanks for this spot!

lib/cucumber/glue/registry_and_more.rb
```diff
@@ -147,15 +147,25 @@ def install_plugin(configuration, registry)
       end
 
       def before_all
+        set_up_world_for_global_hooks
```
luke-hill, line 150: Can you explain the goal here of this bit (Both the setup and the clearer on line 156 for this method)

lib/cucumber/runtime.rb
```diff
@@ -72,15 +72,13 @@ def run!
 
       load_step_definitions
       fire_install_plugin_hook
-      fire_before_all_hook unless dry_run?
```
luke-hill, line 75: Is this meant to be deleted?

luke-hill (commented):
Few minor questions. Great stuff though

lib/cucumber/runtime.rb
brasmusson: Yes. The ndjson files of the cck dev kit specifies that the expected message sequence is that the TestRunHookStarted/Finished messages for the before all hook come after the TestRunStarted message. Therefore I have moved the execution of the before all hooks to after the TestRunStarted event has been sent. The TestRunStarted event is broadcast by one of the filters (BroadcastTestRunStartedEvent) close to the end of the filter chain, so I have therefore created the FilreBeforeAllHooks filter which is placed directly after the BroadcastTestRunStartedEvent filter. Therefore before all hooks shall not be fired from here.

lib/cucumber/glue/registry_and_more.rb
brasmusson: Calling `attach` from hooks and step definition works since the world object handles the attach and creates an event/message for it. To be able to call `attach` from before all and after all hooks, I create a world object here so something exist to create the event/message for the `attach` calls in before all hooks.

lib/cucumber/formatter/message_builder.rb
brasmusson: In the attach method we need to know if the attachment comes from an before_all/after_all hook or from a test case step of hook. Since we store both `current_test_case_started_id` and `current_test_run_hook_started_id`  can be set when test step or hooks are executed and when after all hooks are executed. In this pull request the solution is to use the condition `if @current_test_run_hook_started_id.nil?` to decide whether it is a test step or hook, or a before all hook or after all hook that is executing. Using that condition (in the attach method) requires that we ensure that `@current_test_run_hook_started_id` is `nil` when a test step or hook is executed. That is the purpose of this line.

lib/cucumber/formatter/message_builder.rb
brasmusson: fixed

lib/cucumber/glue/registry_and_more.rb
brasmusson: fixed

brasmusson:
LGTM

PR #1865, opened 2026-05-11
Pass all messages through the event bus

brasmusson:
# Description
Make the MessageBuilder a mandatory formatter and let it pass the created message to the event bus. Users of the message stream then only have to register for on_envelope on the event bus to get the full message stream.

Previously only the messages produced by the gherkin library have been passed though the event bus. This change is a step to completely replace the event other than messages with messages.

After this change event users, mainly formatters, can be updated to use messages without having to inherit from MessageBuilder - which is a kind of hack - but to use the final design of registering for the message stream in the event bus.

When most or all event users have be updated to use message. Event producers can be change to produce messages instead. Then MessageBuilder need to be updated to not convert those messages from events. Finally MessageBuilder can be completely removed.

## Type of change

# Checklist:


luke-hill:
> When most or all event users have be updated to use message. Event producers can be change to produce messages instead. Then MessageBuilder need to be updated to not convert those messages from events. Finally MessageBuilder can be completely removed.

I spoke with Rien about this on thursday. This is something I think is something we should do, but it sounds hard. 

Step 1 I'm going to produce a mapping doc probably as an issue detailing which events map 1->1, which events don't map, and which messages don't have a mapping event source

lib/cucumber/glue/registry_and_more.rb
```diff
@@ -99,6 +99,7 @@ def register_rb_step_definition(string_or_regexp, proc_or_sym, options)
         step_definition = StepDefinition.new(@configuration.id_generator.new_id, self, string_or_regexp, proc_or_sym, options)
         @step_definitions << step_definition
         @configuration.notify :step_definition_registered, step_definition
+        @configuration.notify :envelope, step_definition.to_envelope
```
luke-hill, line 102: Can we remove this out into a separate PR targeting the v11 migration branch

luke-hill (commented):
Reviewed a couple of files (message/rerun), get that work.

The rest I will likely need some help with / pair review

lib/cucumber/runtime.rb
```diff
@@ -191,12 +191,16 @@ def set_encoding
     def report
       return @report if @report
 
-      reports = [summary_report] + formatters
+      reports = [message_builder, summary_report] + formatters
       reports << fail_fast_report if @configuration.fail_fast?
       reports << publish_banner_printer unless @configuration.publish_quiet?
       @report ||= Formatter::Fanout.new(reports)
```
brasmusson, line 191: I start to think that what we see here is technical debt from the time before we started to use event i CucumberRuby. Then the formatters had a lot of methods `beforeScenario`, `afterScenario`, `beforeStep` etc. The Fanout was utility to "iterate" over all the formatters and call these methods, so the "runtime" called `beforeScenario` on the Fanout which resulted in that `beforeScenario` was called on each formatter.

Now when we are using events what is important is that the right formatters are created. When they are created they register for the events they are interested in on the event bus (they get the Configuration object which has the event bus passed to the initialize method). Then the get event from the event bus and do their job.

So the remaining purpose of the code here (which is call from line 77 `self.visitor = report`) is to create the right formatter. Summary_report we always want to create as the runtime uses it to determine the exit code. Now we always want to create the message builder so that is can listen to event and create message, send the to the event bus so that message users can get all messages from the event bus. Then we shall create the formatters specified by the cli options etc.

PR #1876, opened 2026-05-22
bugfix: Handle media type json a bit friendlier

### LINKED ISSUE #1787: HTML cannot be rendered if a 'text/json' attachment is present
### 👓 What did you see?

Given an `object` which represents some JSON object, `attach(object, "text/json")` will correctly add an attachment envelope to the output message stream, and will be included in the HTML body. However, attempting to open this report (at least in Chrome) results in only a white screen being rendered. 

If for example, `object = { "foo": 42 }`, then within the message stream we have (prettified):

```json
{
  "attachment": {
    "body": {"foo":42},
    "contentEncoding": "IDENTITY",
    "mediaType": "text/json",
    "testCaseStartedId": "a92d9794-86d9-4c14-a913-1f8ddb736289",
    "testStepId": "5996f98b-fe36-48e3-94be-b3b2433344c2"
  }
}
```

### ✅ What did you expect to see?

The page should continue to be loadable, and either display the attached `text/json` in a plaintext form or in any other way.

### 📦 Which tool/library version are you using?

cucumber-html-formatter v19.2.0

### 🔬 How could we reproduce it?

1. Create `features/test.feature` with:
```feature
# features/test.feature
Feature: Reproduction of problem
  Scenario: Example scenario
    When nothing happens
```
2. Create `features/step_definitions/all.rb` with:
```ruby
When('nothing happens') { }
```
3. Create `features/support/env.rb` with:
```ruby
After do |test|
  context = {
    "foo" => 42,
  }

  attach(context, "text/json")
end
```
4. Run `cucumber -f html -o cucumber.html`
5. Load the `cucumber.html` into a browser of choice, and check that:
    1. only a white screen is displayed, and
    2. the `attachment` message is present in the message list, and
    3. if the `attachment` message is removed, the page is rendered correctly

### 📚 Any additional context?

_No response_

luke-hill:
# Description

Partially fixes #1787 

## Type of change

# Checklist:


luke-hill:
Hi @mitchgrout can you try this branch/PR and see if it fixes your problem

luke-hill:
It's worth pointing out this actually changes the behaviour of attach. Which previously attached `application/json` bodies as encoded, now it attaches them unencoded. Which according to the CCK is the desired behaviour

lib/cucumber/formatter/message_builder.rb
```diff
@@ -67,6 +68,9 @@ def attach(src, media_type, filename)
         if media_type&.start_with?('text/')
           attachment_data[:content_encoding] = Cucumber::Messages::AttachmentContentEncoding::IDENTITY
           attachment_data[:body] = src
+        elsif media_type.end_with?('json')
+          attachment_data[:content_encoding] = Cucumber::Messages::AttachmentContentEncoding::IDENTITY
+          attachment_data[:body] = src.is_a?(Hash) ? src.to_json : src
```
mpkorstanje, line 82: Looking at the surrounding code, I don't think this implementation is correct.

The attachment encoding is currently derived from the `media_type` but should be derived from the type of `src`. If `src` is a `string` or equivalent type that can be presented properly as a json string then `IDENTITY` can be used. Otherwise `BASE64` should be used.

mpkorstanje: In pseudo code:

```
if src instance of string-like:
          attachment_data[:content_encoding] = IDENTITY
          attachment_data[:body] = src
else if src instance of byte-array-or-stream-like
          attachment_data[:content_encoding] = BASE64
          attachment_data[:body] = Base64.strict_encode64(body)
else 
          throw "unsupported src type, must be string-like or byte-stream like".
```

luke-hill: I looked into the CCK for this. And for the CCK it wants to present JSON bodies as un-encoded.

cf: https://github.com/cucumber/compatibility-kit/blob/main/devkit/samples/attachments/attachments.ndjson#L46

mpkorstanje: The determination of encoding isn't attached to the media type but rather the type of the source.

This is matters because creating the the `BASE64` encoding is rather expensive and difficult to read. So we only use it when the source can't be represented as a regular string.

You can see this in the step definitions. For all `IDENTITY` encodings we use either:

```
the string {string} is attached as {string}
the following string is attached as {string}:
```

https://github.com/cucumber/compatibility-kit/blob/b96ee7298b719b52cc8d9d0be3aa95b7dac20109/devkit/samples/attachments/attachments.ts#L5
https://github.com/cucumber/compatibility-kit/blob/b96ee7298b719b52cc8d9d0be3aa95b7dac20109/devkit/samples/attachments/attachments.ts#L22

While for the `BASE64` we have:

```
an array with {int} bytes is attached as {string}
```

https://github.com/cucumber/compatibility-kit/blob/b96ee7298b719b52cc8d9d0be3aa95b7dac20109/devkit/samples/attachments/attachments.ts#L29

So I could do 


```
String json = '{ "hello": "world" }';
scenario.attach(json, 'application/json') 

byte[] jsonBytes = json.toByteArray();
scenario.attach(jsonBytes, 'application/json') 
```

The first attachment would have `IDENTITY` while the second one would have `BASE64`. And we should probably make that explicit in a test somewhere.

luke-hill: This feels like a breakout I need to do then. Because this was just a patch to ensure that someone passing in a hash can represent it as a nice setup (Which it currently doesn't do anywhere).

luke-hill: Or just thinking about it, we could invert this.

And write something like

```
  if src.respond_to?(:read)
    # base64 logic
  else
    # identity logic
  end
```

mpkorstanje: If you look at the media type you'll be patching holes forever. For example what if you attach a hash with `text/plain`?

I do think the quick fix would be to only support `string`, `src.respond_to?(:read)` and whatever `Base64.strict_encode64` accepts. Then throw an exception with an explanatory message for everything else.

But I don't know enough Ruby to help you here.

mpkorstanje: > Or just thinking about it, we could invert this.

I don't know what `:read` does. It looks like duck typing, but I don't know the what it signifies. 😄

luke-hill: I dug into this code, it's untouched in about 7+ years. I think I'll do the bit I thought of just now. I'll fix it in here so it's done once and for all.

lib/cucumber/formatter/message_builder.rb
```diff
@@ -64,13 +65,12 @@ def attach(src, media_type, filename)
           timestamp: time_to_timestamp(Time.now)
         }
 
-        if media_type&.start_with?('text/')
-          attachment_data[:content_encoding] = Cucumber::Messages::AttachmentContentEncoding::IDENTITY
-          attachment_data[:body] = src
-        else
-          body = src.respond_to?(:read) ? src.read : src
+        if streamed_file
```
mpkorstanje, line 77: Currently the behaviour of the function doesn't seem to guarantee a valid message is produced in every scenario.

And asking the user of the API for this decision doesn't seem to be quite right either.

For example it look like it is possible to attach a byte stream  with identity encoding by passing `streamed_file = true`. 

Is there a way you can detect streamed files from their type or duck-type?

Then you can write:
```
if string
  attach as string with identity encoding
else if hash:
  attach with to_json with identity encoding
else if byte stream
  attach with strict_encode64 and base64 encoding.
else 
  complain about unsupported type.
```

luke-hill: So this API "here" isn't a user one. It's a derived one (I know it's a bit manky).

The actual "user" one is the one here: https://github.com/cucumber/cucumber-ruby/pull/1876/changes#diff-bf3517ff147c7f906bd28ff054347a33276f9f3280a3a6c2f8428a137da43feeR92

lib/cucumber/glue/proto_world.rb
```diff
@@ -89,14 +89,16 @@ def log(*messages)
       # @param filename [string] the name of the file you wish to specify.
       #   This is only needed in situations where you want to rename a PDF download e.t.c. - In most situations
       #   you should not need to pass a filename
-      def attach(file, media_type = nil, filename = nil)
+      def attach(file, media_type = nil, filename = nil, streamed_file = nil)
```
luke-hill: This method is attaching a file.

The `file` input is either a string or a ref to a file (like an image).

We changed the logic a few years ago (Because of a licensing issue), to use minimime to infer the media type (But only if it's a file). Likely we should refactor this to potentially always do this. But then some people want to manually attach something as txt/json

lib/cucumber/formatter/message_builder.rb
luke-hill: I'm going to try remove the ability for the user to specify this.

luke-hill: I've been unable to :( too much spaghetti here

lib/cucumber/glue/proto_world.rb
mpkorstanje: You can also let the user provide the media type, and only if it is absent infer it based on the file contents.

But if anything, media_type shouldn't be used to decide the encoding.

PR #1881, opened 2026-05-31
Use events also when step definitions use "attach" and "log"

brasmusson:
# Description

Remove the ancient procedure call interface from formatters completely

Until now the data from "attach" and "log" calls from step definitions has been passed to the formatter using the ancient procedure call interface used in Cucumber-Ruby v1 and v2. From Cucumber-Ruby v3 all other data to formatters has been passed using events.

Introduce an event AttachCalled to pass the data from "attach" and "log" calls from step definitions to event listeners like formatters.

## Type of change

# Checklist:


lib/cucumber/runtime/user_interface.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 require 'timeout'
+require 'cucumber/events/attach_called'
 
 module Cucumber
   class Runtime
     module UserInterface
-      attr_writer :visitor
```
luke-hill, line 8: Question - Non-blocking as this PR makes sense and is easy to follow.

(2 parts)
- Why is attach redefined about 4 times?
- Where do we ever write to `visitor` previously or `configuration` (Which makes no sense), currently

luke-hill (commented):
This makes a lot of sense and removes some of the tech debt you mentioned. One QQ about some more stuff we can maybe cull

lib/cucumber/runtime/user_interface.rb
brasmusson: The Runtime class includes `UserInterface` (https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/runtime.rb#L55) and was assigning `visitor` https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/runtime.rb#L76 - where `report` was the FanOut-instance wrapping all the formatters.

brasmusson: As I understand it since the Runtime class includes `UserInterface`, then the assignment of `@configuration` in `initialize` (https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/runtime.rb#L58) is (also) setting the `configuration` attr of `UserInterface`.

brasmusson: Before this PR the call chain was:
- a stepdefinition calls `attach`
- `attach` of `ProtoWorld` is invoked which result to a call to `super` (https://github.com/cucumber/cucumber-ruby/blob/release/v11.0.0/lib/cucumber/glue/proto_world.rb#L97)
- `attach` of `UserInterface` is invoked which result in a call to `@visitor.attach`
- `missing_method` of `FanOut` is invoked which result to a call to `attach` on each formatter that has a `attach` method.

With this PR we switch to sending events in the `attach` method of `UserInterface`.

luke-hill: The call chain you've mentioned I think is slightly wrong. Because there's a second `attach` defined in the file proto_world. And when I've removed that it causes problems. I've also tried debugging and it crashes the debugger often.

I'd like to try remove some of these a->b->c->d calls (Diff PR)

brasmusson: Right. It looks more like
`ProtoWorld.attach` (https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/glue/proto_world.rb#L92) ->
`<DynamicModule>.attach` (https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/glue/proto_world.rb#L158) ->
`UserInteface.attach` (https://github.com/cucumber/cucumber-ruby/blob/v11.1.0/lib/cucumber/runtime/user_interface.rb#L44)

luke-hill: Thanks - I've made this placeholder for now: https://github.com/cucumber/cucumber-ruby/issues/1884

luke-hill: Was the previous `visitor.attach` calling this: https://github.com/cucumber/cucumber-ruby/pull/1881/changes#diff-7347b0c4c6a10379b5a657d3076808a43425fb242aedaac82b869fe0a940798cL55

And now you're calling it but in a slightly different way (Using configuration.notify)

brasmusson: Yes. `visitor` was a `FanOut` having all formatters at its recipients https://github.com/cucumber/cucumber-ruby/pull/1881/files#diff-fc62aa05d38336fb22819196433b9acb4370acadd10dbc201a48ccbd5980df6dL191, so `visitor.attach` resulted in a call to `attach` on every formatter that have an `attach` method.

StantonMatt:
I ran a focused check around the `attach`/`log` event path on `174d2a49` with Ruby 3.3.11.

```sh
bundle exec cucumber features/docs/writing_support_code/attachments.feature features/docs/defining_steps/printing_messages.feature features/docs/writing_support_code/hooks/hook_order.feature
bundle exec cucumber features/docs/formatters/json_formatter.feature features/docs/formatters/pretty_formatter.feature
bundle exec rspec spec/cucumber/formatter/json_spec.rb spec/cucumber/formatter/message_spec.rb spec/cucumber/formatter/pretty_spec.rb spec/cucumber/glue/proto_world_spec.rb spec/cucumber/glue/step_definition_spec.rb
git diff --check origin/main...HEAD
```

Results:

- docs `attach`/`log`/hook-order features: `7 scenarios`, `55 steps`, all passed
- JSON + pretty formatter features: `12 scenarios`, `119 steps`, all passed
- formatter/glue specs: `98 examples`, `0 failures`, `2 pending` existing pending examples in `ProtoWorld` attachment specs
- `git diff --check` passed
- visible GitHub checks are also all green: `17/17`, including RuboCop, CodeQL, and the full Ruby/platform test matrix

From that, I do not see a functional gap in the migration from `visitor.attach`/`FanOut` to `configuration.notify(:attach_called, ...)`: the docs-level `attach` and `log` paths still reach message/json/pretty output, and the formatter/glue specs cover the MessageBuilder, JSON/pretty formatters, `ProtoWorld`, and step-definition bridge.

luke-hill (approved):

luke-hill:
@brasmusson can you write up some technical level changelog for this please. This is a great change and this makes a **LOT** more sense to me, so thankyou a LOT for this! :tada:

PR #1882, opened 2026-05-31
Update MessageBuilder when runner generate messages

brasmusson:
# Description

* Changes when the runner generate messages at the source.

Depend on https://github.com/cucumber/cucumber-ruby-core/pull/331

## Type of change

# Checklist:


lib/cucumber/formatter/message_builder.rb
```diff
@@ -57,6 +54,12 @@ def initialize(config)
 
       def on_envelope(event)
         store_current_test_run_hook_started_id(event)
+        @current_test_step_id = event.envelope.test_step_started.test_step_id if event.envelope.test_step_started
+        return unless event.envelope.test_case_started
+
+        @current_test_case_started_id = event.envelope.test_case_started.id
+        @current_test_run_hook_started_id = nil
```
luke-hill, line 58: From my understanding doesn't this "reset" always need to happen. Then line 60 can become

`@current_test_case_started_id = event.envelope.test_case_started.id if event.envelope.test_case_started`

luke-hill:
Requires cucumber-core 18.1+

lib/cucumber/formatter/message_builder.rb
brasmusson: It is equivalent to write line 58-63 as:
```
        if event.envelope.test_case_started
          @current_test_case_started_id = event.envelope.test_case_started.id
          @current_test_run_hook_started_id = nil
          @repository.update(event.envelope)
        end
```
but rubocop will complain "Use a guard clause (`return unless event.envelope.test_case_started`) instead of wrapping the code inside a conditional expression." 

Another equivalent alternative is
```
        @current_test_case_started_id = event.envelope.test_case_started.id if event.envelope.test_case_started
        @current_test_run_hook_started_id = nil if event.envelope.test_case_started
        @repository.update(event.envelope) if event.envelope.test_case_started
```
but I'm not sure about using the same trailing `if` on three lines in a row.

luke-hill: I more meant this.

```
@current_test_case_started_id = event.envelope.test_case_started.id if event.envelope.test_case_started
@current_test_run_hook_started_id = nil
@repository.update(event.envelope)
```

brasmusson: `@current_test_run_hook_started_id = nil`
is completely wrong. What you get if you inline `store_current_test_run_hook_started_id` and snip the line that are not relevant is:
```
@current_test_run_hook_started_id = event.envelope.test_run_hook_started.id if event.envelope.test_run_hook_started

@current_test_run_hook_started_id = nil
```
`@repository.update(event.envelope)` is not completely wrong, but it is not necessary to update the repository with every type of message. Actually since the only usage of the `repository` in `message_builder` is:
`@repository.test_step_by_id[event.test_step.id].pickle_step_id`
(in `generate_snippet_envelope`)
`@repository.update(event.envelope)` is not needed at all here. It is sufficient that the `repository` is updated with the `test_case_ready` envelopes in `on_test_case_ready`.

This should also work:
```
@current_test_case_started_id = event.envelope.test_case_started.id if event.envelope.test_case_started
@current_test_run_hook_started_id = nil if event.envelope.test_case_started
```

luke-hill:
@brasmusson I think this has lots of code that has already been merged into `main`. Not sure if a merge / rebase will fix it.

lib/cucumber/formatter/message_builder.rb
luke-hill: I've re-read this and now understand. Thanks

lib/cucumber/formatter/message_builder.rb
```diff
@@ -54,6 +51,12 @@ def initialize(config)
...
+        @repository.update(event.envelope)
```
luke-hill, line 59: Writing this as a new comment. Now I understand the change on line 55.

I still think we need to fully update the repository after each envelope. So imo this should be moved above line 55.

Given every line now in this method is just assigning an iVar, we could even move the update call to the first line, to make it cleaner and leaner (As in here we always have the envelope already and don't need to generate it - which was the reason for it being at the bottom of all other `on_....` handlers)

luke-hill:
Note to self and @brasmusson 

I've merged a piece of work which naturally breaks a couple of tests. This PR should also fix the tests.

It looks like it's just one feature test - The failure looks "good" (Duplication of message output).

`features/docs/formatters/message.feature:27`

PR #1888, opened 2026-06-03
Escape apostrophes in generated snippets

### LINKED ISSUE #1710: Scenario having nested quotes produces syntactically-invalid step definition
### What did you see?

Step definition produced by running `bundle exec cucumber` on the following scenario produces a syntax error when the code snippets are implemented as a step definition.

```cucumber
Scenario: Listener is within range
  Given Lucy is located 15 metres from Sean
  When Sean shouts "free bagels at Sean's"
  Then Lucy hears Sean's message
```

Produces:

```ruby
# ⋮ 
Then('Lucy hears Sean's message') do
end
```

…which is a syntax error due to the unmatched single quote.

### What did you expect to see?

``` ruby
Then("Lucy hears Sean's message") do
end
```

…or something similar.

### Which tool/library version are you using?

Cucumber v8.0.0.

### How could we reproduce it?

1. Install Ruby (I used v2.7.8 from MacPorts) and bundler (I used v2.1.4 from MacPorts)
2. Create a directory for the example project: `mkdir shouty`
3. Run `bundle config set path 'vendor/bundle'`
4. Run `bundle install`
5. Run `bundle exec cucumber --init`
6. Create the file `features/shout.feature` with the following contents
    ```cucumber
    Scenario: Listener is within range
      Given Lucy is located 15 metres from Sean
      When Sean shouts "free bagels at Sean's"
      Then Lucy hears Sean's message
    ```
8. Run command 'bundle exec cucumber'
9. Copy-paste the snippet for the step definition into the file `features/step_definitions/steps.rb`
    ```ruby
    Given('Lucy is located {int} metres from Sean') do |int|
    # Given('Lucy is located {float} metres from Sean') do |float|
    end

    When('Sean shouts {string}') do |string|
    end

    Then('Lucy hears Sean's message') do
    end
    ```

11. Run `bundle exec cucumber` again
12. See the error message:
    ```
    /path/to/shouty/features/step_definitions/steps.rb10: syntax error, unexpected local variable or method, expecting ')'
    Then('Lucy hears Sean's message') do
                        ^
    /path/to/shouty/features/step_definitions/steps.rb12: unterminated string meets end of file (SyntaxError)
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/glue/registry_and_more.rb123:in `require'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/glue/registry_and_more.rb123:in `load_code_file'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime/support_code.rb145:in `load_file'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime/support_code.rb82:in `block in load_files!'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime/support_code.rb81:in `each'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime/support_code.rb81:in `load_files!'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime.rb274:in `load_step_definitions'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/runtime.rb74:in `run!'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/lib/cucumber/cli/main.rb29:in `execute!'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/gems/cucumber-8.0.0/bin/cucumber9:in `<top (required)>'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/bin/cucumber23:in `load'
    /path/to/shouty/vendor/bundle/ruby/2.7.0/bin/cucumber23:in `<main>'
    ```

### References

1. <https://school.cucumber.io/courses/take/bdd-with-cucumber-ruby/lessons/11303805-add-a-scenario-wire-it-up>

StantonMatt:
## Summary

- Keep single-quoted generated Cucumber Expression snippets when the expression can be represented safely.
- Use double-quoted Ruby snippets when the generated expression contains an apostrophe, with escaping for double quotes, backslashes, and interpolation markers.
- Preserve backslash escaping for single-quoted snippets and cover apostrophe, interpolation, and trailing-backslash cases.

Fixes #1710

## Verification

- `bundle exec rspec spec/cucumber/glue/snippet_spec.rb`
- `bundle exec rspec`
- `bundle exec cucumber`
- `bundle exec rubocop`
- `git diff --check`

Local repros now print `Then("Lucy hears Sean's message")` and `Then('Lucy hears trailing \\')`, and both generated snippets compile with Ruby.

### DIFF
--- a/spec/cucumber/glue/snippet_spec.rb
+++ b/spec/cucumber/glue/snippet_spec.rb
@@ -180,6 +180,17 @@ module Cucumber
 
           expect(snippet.to_s).to eq(cucumber_output)
         end
+
+        it 'escapes apostrophes in generated cucumber expression snippets' do
+          @step_text = "Lucy hears Sean's message"
+          cucumber_output = <<~CUKE.chomp
+            Given('Lucy hears Sean\\'s message') do
+              pending # Write code here that turns the phrase above into concrete actions
+            end
+          CUKE
+
+          expect(snippet.to_s).to eq(cucumber_output)
+        end
       end
     end
   end
--- a/spec/cucumber/glue/snippet_spec.rb
+++ b/spec/cucumber/glue/snippet_spec.rb
@@ -181,10 +181,10 @@ module Cucumber
           expect(snippet.to_s).to eq(cucumber_output)
         end
 
-        it 'escapes apostrophes in generated cucumber expression snippets' do
+        it 'uses double quotes when generated cucumber expression snippets contain apostrophes' do
           @step_text = "Lucy hears Sean's message"
           cucumber_output = <<~CUKE.chomp
-            Given('Lucy hears Sean\\'s message') do
+            Given("Lucy hears Sean's message") do
               pending # Write code here that turns the phrase above into concrete actions
             end
           CUKE
--- a/spec/cucumber/glue/snippet_spec.rb
+++ b/spec/cucumber/glue/snippet_spec.rb
@@ -181,7 +181,7 @@ module Cucumber
           expect(snippet.to_s).to eq(cucumber_output)
         end
 
-        it 'uses double quotes for generated cucumber expression snippets with apostrophes' do
+        it 'uses double quotes for generated snippets with apostrophes' do
           @step_text = "Lucy hears Sean's message"
           cucumber_output = <<~CUKE.chomp
             Given("Lucy hears Sean's message") do
@@ -192,7 +192,7 @@ module Cucumber
           expect(snippet.to_s).to eq(cucumber_output)
         end
 
-        it 'escapes backslashes in single quoted cucumber expression snippets' do
+        it 'escapes backslashes in single quoted snippets' do
           @step_text = 'Lucy hears trailing \\'
           cucumber_output = <<~'CUKE'.chomp
             Given('Lucy hears trailing \\') do
@@ -203,7 +203,7 @@ module Cucumber
           expect(snippet.to_s).to eq(cucumber_output)
         end
 
-        it 'escapes interpolation markers in double quoted cucumber expression snippets' do
+        it 'escapes interpolation markers in double quoted snippets' do
           @step_text = 'Lucy hears Sean\'s #@message'
           cucumber_output = <<~'CUKE'.chomp
             Given("Lucy hears Sean's \#@message") do
--- a/spec/cucumber/glue/snippet_spec.rb
+++ b/spec/cucumber/glue/snippet_spec.rb
@@ -181,10 +181,21 @@ module Cucumber
           expect(snippet.to_s).to eq(cucumber_output)
         end
 
-        it 'uses double quotes when generated cucumber expression snippets contain apostrophes' do
+        it 'uses double quotes for generated cucumber expression snippets with apostrophes' do
           @step_text = "Lucy hears Sean's message"
           cucumber_output = <<~CUKE.chomp
             Given("Lucy hears Sean's message") do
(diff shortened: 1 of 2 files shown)


luke-hill (changes requested):
Unfortunately this is what we previously had and we refactored it out many many moons ago.

The desired output here is to instead use double quoted snippets. Which is the previous behaviour. So we should use single quotes (If we can), then failing that, use double quotes. This avoids any issues.

StantonMatt:
Updated this to keep single-quoted snippets when the generated expression has no apostrophe, and switch to double-quoted snippets when it does. I reran the full RSpec suite and RuboCop on the touched files after pushing this.

spec/cucumber/glue/snippet_spec.rb
```diff
@@ -180,6 +180,39 @@ module Glue
 
           expect(snippet.to_s).to eq(cucumber_output)
         end
+
+        it 'uses double quotes for generated cucumber expression snippets with apostrophes' do
```
luke-hill: These aren't cucumber expression snippets. They're all regular snippets

StantonMatt: You are right; I renamed those examples in `06591c7` so they refer to generated snippets rather than cucumber expression snippets.

spec/cucumber/glue/snippet_spec.rb
```diff
@@ -180,6 +180,39 @@ module Glue
 
           expect(snippet.to_s).to eq(cucumber_output)
         end
+
+        it 'uses double quotes for generated snippets with apostrophes' do
```
luke-hill:
```suggestion
        it 'uses double quotes for generated snippets that contain apostrophes' do
```

spec/cucumber/glue/snippet_spec.rb
```diff
@@ -180,6 +180,39 @@ module Glue
...
+          cucumber_output = <<~CUKE.chomp
+            Given("Lucy hears Sean's message") do
+              pending # Write code here that turns the phrase above into concrete actions
+            end
+          CUKE
+
+          expect(snippet.to_s).to eq(cucumber_output)
+        end
+
+        it 'escapes backslashes in single quoted snippets' do
```
luke-hill:
```suggestion
        it 'escapes backslashes when generating snippets' do
```

spec/cucumber/glue/snippet_spec.rb
```diff
@@ -180,6 +180,39 @@ module Glue
...
+          cucumber_output = <<~'CUKE'.chomp
+            Given('Lucy hears trailing \\') do
+              pending # Write code here that turns the phrase above into concrete actions
+            end
+          CUKE
+
+          expect(snippet.to_s).to eq(cucumber_output)
+        end
+
+        it 'escapes interpolation markers in double quoted snippets' do
```
luke-hill, line 206: I think you need to review what this is saying. As it doesn't quite marry up

luke-hill:
Hi @StantonMatt are you still working on this?

PR #1896, opened 2026-07-21
Fix spec for trap_thread_dump_signal

eregon:
# Description

* The spec was timing-dependant and therefore flaky.
* Use a common pattern guaranteed to match.
* Always wait for the signal since Process.kill can be asynchronous (and in real cases always is).
* See https://github.com/truffleruby/truffleruby/issues/4383#issuecomment-5038819762

Locally I added a `1000.times { ... }` around to spec and that way I could repro the failure locally (100 was not enough).
Before the fix it failed reliably a few times, after it passes reliably.

## Type of change

- spec fix

# Checklist:

- [x] CHANGELOG.md has been updated (not user-facing so no entry)

### DIFF
--- a/spec/cucumber/cli/main_spec.rb
+++ b/spec/cucumber/cli/main_spec.rb
@@ -73,11 +73,8 @@ RSpec.describe Cucumber::Cli::Main do
       let(:runtime) { double('runtime').as_null_object }
 
       it 'dumps the thread backtrace to the error stream' do
-        kill_line = 0
-
         allow(runtime).to receive(:run!) do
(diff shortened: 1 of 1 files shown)


luke-hill:
@eregon think there's a JRuby error hre

eregon:
Ah yes because JRuby runs the handler on a non-main Thread, I'll adjust the `pattern` for that.

eregon:
@luke-hill Done, please start the CI and take another look.
I ran the spec locally on JRuby and it passed now.

PR #1891, opened 2026-06-04
bugfix: HTML Formatter and its envelopes

luke-hill:
# Description

Fix HTML formatter not running output_envelope correctly

fixes #1890

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

Please add an entry to the relevant section of CHANGELOG.md as part of this pull request.

# Checklist:

Your PR is ready for review once the following checklist is
complete. You can also add some checks if you want to.

- [ ] Tests have been added for any changes to behaviour of the code
- [ ] New and existing tests are passing locally and on CI
- [ ] `bundle exec rubocop` reports no offenses
- [ ] RDoc comments have been updated
- [ ] CHANGELOG.md has been updated

Issue #1890, opened 2026-06-04
HTML report truncated at `window.CUCUMBER_MESSAGES = [`


oellul:
### 👓 What did you see?

The generated HTML report ends with this line:
```html
<script>
window.CUCUMBER_MESSAGES = [

```

### ✅ What did you expect to see?

window.CUCUMBER_MESSAGES array should be populated:
```html
<script>
window.CUCUMBER_MESSAGES = [
{"meta":{"protocolVersion":"32.3.1","implementation":{"name":"cucumber-ruby","version":"11.1.0"},"runtime":{"name":"ruby","version":"4.0.5"},"os":{"name":"linux","version":"#1 SMP PREEMPT_DYNAMIC Sun, 17 May 2026 17:23:07 +0000"},"cpu":{"name":"x86_64"}}},
...
]
</script>
```

### 📦 Which tool/library version are you using?

ruby v4.0.5
cucumber-ruby v11.1.0
linux

### 🔬 How could we reproduce it?

1. Setup a blank project
2. Add `Gemfile`
```ruby
# frozen_string_literal: true

source 'https://rubygems.org'

gem 'cucumber'
```
3. Add basic feature file: `features/feature/test.feature`
```gherkin
Feature: Test

  Scenario: Run a step
    When I run a single step
```
4. Add step definition: `features/step_definitions/test.rb`
```ruby
# frozen_string_literal: true

When('I run a single step') do
  expect(1).to eq(1)
end
```
5. Run `bundle install`
6. Run `bundle exec cucumber --format html --out report.html`

### 📚 Any additional context?

I found that the probable cause was that output_envelope is defined but not subscribed to the :envelope event

I am working around the issue with this monkey-patch:
`features/support/html_formatter_fix.rb`
```ruby
# frozen_string_literal: true

require 'cucumber/formatter/html'

module Cucumber
  module Formatter
    class HTML
      alias _bugfix_original_initialize initialize

      def initialize(config)
        _bugfix_original_initialize(config)
        config.on_event(:envelope) do |event|
          envelope = event.envelope
          @html_formatter.write_message(envelope)
          @html_formatter.write_post_message if envelope.test_run_finished
        end
      end
    end
  end
end
```

luke-hill:
I can see the issue. Actually a slightly better fix is to just override `on_envelope` in the HTML formatter proper. I'll get working on a fix soon. Sorry for this oversight.

PR #1899, opened 2026-07-24
The Formatter::HTML should not inherit from the MessageBuilder

brasmusson:
# Description
* The  `Runtime` creates the `MessageBuilder` to always output the envelope it creates from event to the event bus. Therefore a formatter only need to listen to the `:envelope` events on the event bus.

This change should have been part of #1865. Since it was missed there #1890 occurred. #1891 fixed the symptom of #1890 but failed to remove the inheritance fo the `Formatter::HTML`

## Type of change

# Checklist:

### DIFF
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -172,13 +172,13 @@ module Cucumber
     require 'cucumber/core/report/summary'
 
     def create_formatters
-      # Define all formatters which are specified via cli options
-      formatters
-      # Define the MessageBuilder formatter - Required until all messages are generated at the source
+      # Define the MessageBuilder - Required until all messages are generated at the source
       message_builder
       # `summary_report` and `global_hooks_summary_report` are used to determine the exit code
       summary_report
       global_hooks_summary_report
+      # Define all formatters which are specified via cli options
+      formatters
       fail_fast_report if @configuration.fail_fast?
       publish_banner_printer unless @configuration.publish_quiet?
     end
--- a/lib/cucumber/runtime.rb
+++ b/lib/cucumber/runtime.rb
@@ -175,8 +175,7 @@ module Cucumber
       # Define all formatters which are specified via cli options
       formatters
       # Define the MessageBuilder formatter - Required until all messages are generated at the source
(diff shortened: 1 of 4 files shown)


luke-hill (changes requested):
I think we need to ensure we're calling repository update if we no longer call super. Especially with the missing MessageHandler method. Which ensures that attachments are good

brasmusson:
The HTML formatter does not have a repository, it just listens to the message stream and forwards the messages to the HTML-formatter from the html-formatter repo. It is the MessageBuilder instance in the Runtime (https://github.com/cucumber/cucumber-ruby/pull/1899/changes#diff-fc62aa05d38336fb22819196433b9acb4370acadd10dbc201a48ccbd5980df6dR178) that ensures that the full message stream is available on the event-bus and that the attachments are good.

luke-hill:
So will the `on_envelope` handler for the `MessageBuilder` still be called when using the HTML Formatter?

Reason I ask is because we need the `store_...` method to be called so that attachments have their source preserved.

brasmusson:
The `Runtime` creates an instance of `MessageBuilder`, it will register its `on_envelope` method on the `EventBus` for `envelope` events. When a `envelope` event is passed to the `EventBus` the `on_envelope` method of the `MessageBuilder` instance will be called. This will happen regardless of which formatters that are used during the test run.

luke-hill (approved):
Approving on the proviso that the `on_envelope` method for the `MessageBuilder` class is called **as well as** the `on_envelope` method for the `HTML formatter` (Sequentially and the MessageBuilder one needs calling first).

lib/cucumber/runtime.rb
```diff
@@ -175,8 +175,7 @@ def create_formatters
       # Define all formatters which are specified via cli options
       formatters
       # Define the MessageBuilder formatter - Required until all messages are generated at the source
-      #   Skip when a user formatter already inherits from MessageBuilder (e.g. HTML) to avoid sending every event twice.
-      message_builder unless formatters.any? { |f| f.is_a?(Formatter::MessageBuilder) }
+      message_builder
```
luke-hill, line 176: This might need re-ordering and going above `formatters`

PR #1900, opened 2026-07-26
Make fetching the OS version work on Wine

mvz:
# Description

Fetching version via Sys::Uname.uname makes sys-uname fetch a large number of OS properties. Under Wine, an unrelated property fails to be fetched, causing MetaMessageBuiler.os to raise an error. With this change, only the relevant value is fetched and the method succeeds.

The context is that I'm trying to get the Aruba test suite to run under Wine. Cucumber fails to start due to Sys::Uname.uname raising an error.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
 
 ## [Unreleased]
 
+### Fixed
+- Make fetching the OS version work on Wine ([#1900](https://github.com/cucumber/cucumber-ruby/pull/1900) [mvz](https://github.com/mvz))
+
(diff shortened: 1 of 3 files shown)


luke-hill:
@mvz in theory I assume this is fine. I've no idea on the package and what it does, but if CI is good I see no reason to block this.

Is this RtM?

luke-hill (approved):

mvz:
@luke-hill yes, this is ready. One thing: i have verified that the resulting version output is identical on Linux, but not on MacOS or windows. Since it's just a string, with rather a lot of content, I don't think the exact details of the string important, but you may have more insight in this.

luke-hill:
On linux there is no difference and given this is the main thing. And we have it tested primitively here, I think this is good to merge.

PR #1907, opened 2026-08-30
Pass the retry_total_tests configuration option to the runner

brasmusson:
# Description

Companion to https://github.com/cucumber/cucumber-ruby-core/pull/344

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- If your change may impact future contributors, explain it here, and remember to update README.md and CONTRIBUTING.md accordingly.

Please add an entry to the relevant section of CHANGELOG.md as part of this pull request.

# Checklist:

Your PR is ready for review once the following checklist is
complete. You can also add some checks if you want to.

- [ ] Tests have been added for any changes to behaviour of the code
- [X] New and existing tests are passing locally and on CI
- [X] `bundle exec rubocop` reports no offenses
- [ ] RDoc comments have been updated
- [ ] CHANGELOG.md has been updated

PR #1910, opened 2026-09-05
Report will_be_retried from the retry filter's decision

### LINKED ISSUE #1905: `retry-total` conflicts with `html-formatter` when exhaustion limit hit
### 👓 What did you see?

The HTML report does not include scenarios which are skipped from being retried after exhaustion limit is hit

### ✅ What did you expect to see?

HTML formatter would show failures

### 📦 Which tool/library version are you using?

cucumber 11.1.1
html-formatter 23.1.0

### 🔬 How could we reproduce it?

Hit the exhaustion limit on a running context. So for example have a folder with 3 features and 3 scenarios. Have all 3 fail, set --retry 1 and --retry-total 2.

The third scenario won't feature in the report.

### 📚 Any additional context?

#1669 introduced it

Enceradeira:
# Description

Fixes #1905. Companion to cucumber/cucumber-ruby-core#345, and an alternative to #1907 / cucumber/cucumber-ruby-core#344.

With `--retry 1 --retry-total 2` and three failing scenarios, the third scenario is missing from the HTML report. The `TestCaseFinished` message of that scenario claims `willBeRetried: true`, because the flag was computed from `--retry` alone, but the retry filter, honouring `--retry-total`, never retries it. The HTML formatter hides an attempt that will be retried and waits for one that never comes.

On `main` the messages are broken in two further ways, which #1907 does not address:

- The runner is given the number of retries where it expects the number of attempts. With `--retry 1` `willBeRetried` is never `true`, so a retried attempt is reported as a final result and the report shows the scenario twice.
- The first attempt's `TestCaseFinished` is emitted after the retry, with the started id of the retry (fixed in the core PR).

This PR introduces `Cucumber::RetryPolicy`, shared by the retry filter and the runner. The filter asks it whether a failed test case is retried and records what it did, the runner asks the same question when it reports the outcome. `will_be_retried?` has no side effects, so both get the same answer.

A new scenario in `features/docs/cli/retry_failing_tests.feature` asserts the message values (`attempt`, `willBeRetried`, and that every `testCaseFinished` references the started id of the same attempt), which no existing test did.

## Notes

- The `Gemfile` temporarily takes `cucumber-core` from the branch of cucumber/cucumber-ruby-core#345 so CI can run. To be removed once core 20 is released, together with raising the upper bound in `cucumber.gemspec`.
- `attempt` is emitted 1-based while the messages spec says the first attempt should be 0. Pre-existing and out of scope here.

## Type of change

- Bug fix

# Checklist:


lib/cucumber/runtime.rb
```diff
@@ -48,7 +48,7 @@ def run!
       fire_install_plugin_hook
       create_formatters
 
-      receiver = Test::Runner.new(@configuration.event_bus, @configuration.id_generator, Cucumber::Formatter::BacktraceFilter, @configuration.retry_attempts)
+      receiver = Test::Runner.new(@configuration.event_bus, @configuration.id_generator, Cucumber::Formatter::BacktraceFilter, @configuration.retry_policy)
```
luke-hill, line 51: Note to self / others. We're baking in more and more complexity here by using posargs.

Given we're releasing a breaking change next (And it's already agreed to be big). I think we need to move to kwargs so we can have some mandated and some optional not reliant solely on pos

luke-hill: My pref.

Line above.

`options = {}` then add in each opt.
Caller takes 1 arg. options

cc/ @brasmusson @Enceradeira

lib/cucumber/retry_policy.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Cucumber
+  # Decides whether a failed test case is run again, honouring both the `--retry` and the `--retry-total` options.
+  #
+  # The retry filter asks it and then records what it did, the test runner asks it when it reports the outcome of a
+  # test case. Both have to get the same answer, so #will_be_retried? never changes the state of the policy.
+  class RetryPolicy
+    def initialize(max_retries, max_permanent_failures)
```
luke-hill, line 9: Can we instead lean on our config. I believe this is available in core as well

This is not 100% "known", just a thought. I could be wrong so ignore if I am

lib/cucumber/runtime.rb
Enceradeira: Done, in both repos, since the constructor is defined in core:

- **cucumber-ruby-core** (cucumber/cucumber-ruby-core#345, commit `2d7df3f`): `id_generator`, `backtrace_filter` and `retry_policy` are now keyword arguments of `Test::Runner#initialize`. `event_bus` stays positional as it's the one mandatory dependency and the rest of core passes it positionally too (`execute`, `compile`, `Compiler.new`). `Runner.new(event_bus)` keeps working unchanged. CHANGELOG and the 20.0.0 upgrading notes are updated.
- **cucumber-ruby** (`48384fc7`): this line now builds an `options` hash above and splats it in:

  ```ruby
  options = {
    id_generator: @configuration.id_generator,
    backtrace_filter: Cucumber::Formatter::BacktraceFilter,
    retry_policy: @configuration.retry_policy
  }
  receiver = Test::Runner.new(@configuration.event_bus, **options)
  ```

If you'd rather have `event_bus` as a required keyword as well, so the caller passes only `**options`, that's a small follow-up (three call sites) — please let me know.

lib/cucumber/retry_policy.rb
Enceradeira: I might be misreading the suggestion, so please correct me. If you mean `RetryPolicy.new(configuration)` with the policy reading `retry_attempts` and `retry_total_tests` itself: I couldn't find a configuration object in core — as far as I can see `Core::Filter.new(:configuration)` is just an attribute the filters are handed, and core's `Test::Runner` only receives the policy and calls `will_be_retried?` on it.

Within cucumber-ruby it could take the `Configuration`, but as the policy is built in `Configuration#retry_policy`, that would be the configuration passing itself to the policy just to read two of its values back, and I found the two plain values easier to unit-test. That said, I'm happy to change it if you had something else in mind, or if you'd prefer the policy to take the configuration for consistency with the filters.

lib/cucumber/configuration.rb
```diff
@@ -280,6 +281,10 @@ def id_generator
       @id_generator ||= Cucumber::Messages::Helpers::IdGenerator::UUID.new
     end
 
+    def retry_policy
```
luke-hill, line 284: This is the comment I meant in core. So yes, this is good. But I would have this as an api private entity that is not configurable in any way shape or form, it's entirely closed

